PR #115, opened 2013-09-19
AFNI ebuild

TheChymera:
AFNI is the open-source environment of the NIH for processing and displaying functional MRI data—a technique for mapping human brain activity. It is an important tool in modern neuroscience research. Especially since it is one of the few FOSS fMRI analysis suites which also do not rely on non-FOSS packages, we should definitely provide this to our users.

I have encountered some minor complications with the make files they provide, but I have been able to work around that - some gentoo devs suggested I rewrite them, but I would not know how. If you would like to help with that I would very much appreciate it!
Apparently the neurodebian people have some newer (and allegedly better) build instructions of their own (described here http://afni.nimh.nih.gov/afni/community/board/read.php?1,142446,142566#msg-142566 ). Should we bother?


TheChymera:
Upstream keeps a list of files which should be able to run as scripts at https://github.com/afniHQ/AFNI/blob/master/src/prog_list.txt . This document automatically gets pulled by my ebuild (since I use the GitHub URI). How can I use it to make the ebuild selectively add run permissions to these files (currently I use `insinto` and everything gets installed, but without the required permissions).

@jlec @junghans  @kiwifb - could you guys help me out again?

TheChymera:
@jlec could you have a look at this again? I think this has come a long way, and it's finally behaving nicely (at least on my machine). Maybe you can pull it?

sci-biology/afni/afni-0_pre20130927.ebuild
```diff
@@ -0,0 +1,50 @@
...
+KEYWORDS="~amd64"
+IUSE=""
+
+RDEPEND="x11-libs/motif[-static-libs]"
+
+# x11-libs/motif[static-libs] breaks the build. 
+# See upstream discussion 
+# http://afni.nimh.nih.gov/afni/community/board/read.php?1,85348,85348#msg-85348
+
+DEPEND="x11-libs/motif[-static-libs]
```
jlec: Just put ${RDEPEND} here

sci-biology/afni/afni-0_pre20130927.ebuild
```diff
@@ -0,0 +1,50 @@
...
+	x11-libs/libXpm"
+
+S=${WORKDIR}/afni_src
+BUILD="linux_fedora_19_64"
+
+src_prepare() {
+	sed -e 's/-V 32//g' -i other_builds/Makefile.${BUILD} || die # they provide somewhat problematic makefiles :(
+	cp other_builds/Makefile.${BUILD} Makefile || die # some Makefile under ptaylor looks
+	# for the parent makefile at "Makefile".	
+	}
```
jlec: no indention here

sci-biology/afni/afni-0_pre20130927.ebuild
```diff
@@ -0,0 +1,50 @@
...
+
+src_compile() {
+	emake -j1 totality
+	}
```
jlec: no indention here

sci-biology/afni/afni-0_pre20130927.ebuild
```diff
@@ -0,0 +1,50 @@
...
+
+src_install() {
+	insinto /opt/${PN}
+	doins -r "${S}/${BUILD}"/*
+	echo "LDPATH=/opt/afni" >> "${T}"/95${PN} || die "Can't write environment variable."
+	doenvd "${T}"/95${PN}
+
+	dobin "${S}/${BUILD}/${PN}"
+	pax-mark m "${D}/usr/bin/${PN}"
+	}
```
jlec: no indention here

TheChymera:
I had actually forgotten to update the non-live ebuild. Updated it, applied the corrections you suggested to both live and non-live ebuilds, tested them again, and everything works.

TheChymera:
@jlec Bump? I believe I addressed everything you just mentioned

sci-biology/afni/afni-0_pre20130927.ebuild
```diff
@@ -0,0 +1,72 @@
...
+		-i SUMA/SUMA_Makefile || die "Could not edit SUMA/SUMA_Makefile"
+		# upstream checks if $CC is EXACTLY gcc, else sets variables for Mac
+}
+
+src_compile() {
+	emake INSTALLDIR="${D}/opt/${PN}" -j1 all plugins suma_exec
+}
+
+src_install() {
+	emake INSTALLDIR="${D}/opt/${PN}" -j1 install install_plugins
```
jlec: It looks as we are building the package from source, so it should be installed to /usr and not /opt.

TheChymera: `${D}/usr/${PN}` will do?

jlec: https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard should help.

TheChymera: Is it ok to make subflders in `/bin/`? I was thinking to do this:

```
src_install() {
	emake INSTALLDIR="${D}/bin/${PN}" -j1 install install_plugins
        emake LIBDIR="${D}/lib/${PN}
```
 Is that ok?

TheChymera: @jlec not a good idea?

jlec: /bin  Essential command binaries that need to be available in single user mode; for all users, e.g., cat, ls, cp.

Is this package "essential"?

TheChymera: sorry, I meant:

```
src_install() {
    emake INSTALLDIR="${D}/usr/bin/${PN}" -j1 install install_plugins
    emake LIBDIR="${D}/usr/lib/${PN}
```

Is that a good idea? Or is it better to not create a directory under `/usr/bin` would stuff from there get loaded into the path automatically?

If not, would this be better?

```
src_install() {
    emake INSTALLDIR="${D}/usr/bin" -j1 install install_plugins
    emake LIBDIR="${D}/usr/lib/${PN}
```

Not least of all, is it smart to use the `/usr/lib` symlink, or should I explicitly put it under `/usr/lib64`?

jlec: Try FEATURES=multilib-strict.

TheChymera: How do I try that? [This](https://devmanual.gentoo.org/archs/amd64/index.html) is the only reference abut multilib-strict that I could find, and it does not say how to actually use it. But I understand that once I am using this I can do this:

```
emake LIBDIR="${D}/usr/lib/${PN}
```

without worrying about messing stuff up?

What about `/usr/bin/${PN}` or just `/usr/bin`

jlec: take a look into man make.conf.

And of course it needs to be in /usr/bin, otherwise is isn't in $PATH.

TheChymera: Ok, thanks for clarification on the `/usr/bin` thing, but the relevant section in man make.conf just restates what I found on the wiki. Ok, I (think I) understand what this does (protects me from putting my 64libs anywhere else the symlink may point) *but* how do I set it for my ebuild? isn't this supposed to go under make.conf then?

TheChymera:
@jlec, is this better now? Please also let me know regarding the libs and how I can set `FEATURES=multilib-strict` for my ebuild.

kiwifb:
The `FEATURES=multilib-strict` will tell you your current ebuild is not multilib ready. I invite you to read the `multilib` eclass https://devmanual.gentoo.org/eclass-reference/multilib.eclass/index.html

What you need is to install to `/usr/lib$(get_libdir)` (don't forget to inherit `multilib`). To support prefix it would be nice change `${D}` to `${ED}`.

TheChymera:
Ah yes, I remember I totally messed up my system with `$(get_libdir)` once. All my lib folders got moved to my rootdir, and I had files called liblib and libliblib *everywhere*. I still did not figure out how exactly that happened, though. Is there anything I should be particularly careful about if I use `/usr/lib$(get_libdir)`?

kiwifb:
Yes because it is better for me to wait to be totally awake. `/usr/$(get_libdir)` is what you want. get_libdir return either `lib` or `lib64`.

TheChymera:
Is this ok now?

What about 

```
src_compile() {
	emake INSTALLDIR="${ED}/opt/${PN}" -j1 all plugins suma_exec
}
```
Does this still make sense given we have no business with `/opt` anymore?

TheChymera:
@kiwifb anything regarding that one thing still left under `/opt`?

kiwifb:
get_libdir should be ok now. You shouldn't need to provide INSTALLDIR during `src_compile` only in `src_install`, does this command actually install anything?

sci-biology/afni/afni-0_pre20130927.ebuild
```diff
@@ -0,0 +1,57 @@
...
+		-i SUMA/SUMA_Makefile || die "Could not edit SUMA/SUMA_Makefile"
+		# upstream checks if $CC is EXACTLY gcc, else sets variables for Mac
+}
+
+src_compile() {
+	emake INSTALLDIR="${ED}/opt/${PN}" -j1 all plugins suma_exec
+}
+
+src_install() {
+	emake INSTALLDIR="${ED}/ur/bin" -j1 install install_plugins
```
kiwifb: `/ur/bin`? Please put the missing `s` in.

kiwifb:
Bundle netcdf 3.5.0! We'll let go that one because it is positively ancient, but someone upstream should seriously update. I am sure netcdf 4.x has reader/writer for the old format. More seriously `sci-libs/gts` is bundled and will have to go.

At the moment installing is broken for me because the makefiles are positively stupid. I asked if something was installed during `src_compile`, kinda if `INSTALLDIR` is defined. Only a couple of static libraries are then installed. I am sure we won't miss them so remove `INSTALLDIR` from `src_compile`.

TheChymera:
yes, this used to work before I corrected the install phase, but now it fails with [this log](/var/tmp/portage/sci-biology/afni-9999/temp/build.log). Most likely because of this part:

```
===== AFNI install finished (fondly do we hope, fervently we do pray) =====
/bin/mv -f *.so /var/tmp/portage/sci-biology/afni-9999/image//usr/bin
mkdir -p /var/tmp/portage/sci-biology/afni-9999/image//usr/bin/funstuff
/bin/cp -f faces/face_*.jpg faces/splash_*.jpg poems/poem_*.txt /var/tmp/portage/sci-biology/afni-9999/image//usr/bin/funstuff
echo Plugins put in directory /var/tmp/portage/sci-biology/afni-9999/image//usr/bin:
Plugins put in directory /var/tmp/portage/sci-biology/afni-9999/image//usr/bin:
echo Do not forget to setenv AFNI_PLUGINPATH appropriately
Do not forget to setenv AFNI_PLUGINPATH appropriately
make -j3 LIBDIR=/var/tmp/portage/sci-biology/afni-9999/image//usr/lib64 -j1 install_lib 
make: *** No rule to make target '/var/tmp/portage/sci-biology/afni-9999/image//usr/lib64', needed by 'install_lib'.  Stop.
```

I don't get it, how could changing the directories break this? Why does it still try to do `install_lib` after AFNI clearly states it is finished?

TheChymera:
@kiwifb regarding your recommendation, does that eman I should delete the entire `src_compile()` phase? or just write 

```
src_compile() {
    emake INSTALLDIR="${ED}/opt/${PN}" -j1 all plugins suma_exec
}
``` 
in there?

kiwifb:
Yes, that's the `src_compile` I want to see. For the installation, the problem is that `Makefile.INCLUDE` has one of these
```
$(INSTALLDIR):
	if [ ! -d $(INSTALLDIR) ]; then $(MKDIR) -p $(INSTALLDIR); echo Made directory $(INSTALLDIR); fi
```
But no corresponding line for `LIBDIR`. Then install target are of the form
```
install: $(INSTALLDIR) $(INSTALL_PREREQ) ptaylor_install pkundu_install
```
And it is fine because you have a rule for `$(INSTALLDIR)` but you also have
```
install_lib: $(LIBDIR)
```
But you are missing a `$(LIBDIR)` target. The assumption is that `$(LIBDIR)` is equal to another already defined target. In the main `Makefile` you have
```
INSTALLDIR = ./$(SYSTEM_NAME)
LIBDIR = $(INSTALLDIR)
```
So as long as `LIBDIR` and `INSTALLDIR`  are the same you are in the clear, but we redefine them to be different: BOOOM.

The following in `src_prepare` should get rid of the problem
```
sed -i "s|install_lib: \$(LIBDIR)|install_lib:|" Makefile.INCLUDE
```

kiwifb:
Scrap the `sed` line it won't work because they use plain `cp` to install files instead of the `intall` command. `install` would create folders `cp` doesn't. I wanted to avoid writing a patch but fear not because there is a perverted way out. In `src_install` replace `LIBDIR` with `INSTALLDIR` and it will work. And yes I mean this
```
	emake INSTALLDIR="${ED}/usr/bin" -j1 install install_plugins
	emake INSTALLDIR="${ED}/usr/$(get_libdir)" -j1 install_lib
```

TheChymera:
Ok, that solved that issue - but apparently putting stuff into the right place uncovers all sorts of other problems with it:

```
>>> Installing (1 of 1) sci-biology/afni-9999::neurogentoo
 * This package will overwrite one or more files that may belong to other
 * packages (see list below). You can use a command such as `portageq
 * owners / <filename>` to identify the installed package that owns a
 * file. If portageq reports that only one package owns a file then do
 * NOT file a bug report. A bug report is only useful if it identifies at
 * least two or more packages that are known to install the same file(s).
 * If a collision occurs and you can not explain where the file came from
 * then you should simply ignore the collision since there is not enough
 * information to determine if a real problem exists. Please do NOT file
 * a bug report at http://bugs.gentoo.org unless you report exactly which
 * two packages install the same file(s). See
 * http://wiki.gentoo.org/wiki/Knowledge_Base:Blockers for tips on how to
 * solve the problem. And once again, please do NOT file a bug report
 * unless you have completely understood the above message.
 * 
 * Detected file collision(s):
 * 
 * 	/usr/bin/qdelaunay
 * 	/usr/bin/whirlgif
 * 	/usr/bin/djpeg
 * 	/usr/bin/cjpeg
 * 	/usr/bin/qhull
 * 	/usr/bin/rbox
 * 	/usr/bin/count
 * 	/usr/bin/mpeg_encode
 * 
 * Searching all installed packages for file collisions...
 * 
 * Press Ctrl-C to Stop
 * 
 * media-libs/qhull-2012.1-r4:0::gentoo
 * 	/usr/bin/qdelaunay
 * 	/usr/bin/qhull
 * 	/usr/bin/rbox
 * 
 * sys-devel/llvm-3.5.1:0::gentoo
 * 	/usr/bin/count
 * 
 * media-video/mpeg-tools-1.5b-r5:0::gentoo
 * 	/usr/bin/mpeg_encode
 * 
 * sci-biology/fsl-5.0.8:0::neurogentoo
 * 	/usr/bin/whirlgif
 * 
 * media-libs/libjpeg-turbo-1.4.0:0::gentoo
 * 	/usr/bin/cjpeg
 * 	/usr/bin/djpeg
 * 
 * Package 'sci-biology/afni-9999' NOT merged due to file collisions. If
 * necessary, refer to your elog messages for the whole content of the
 * above message.
```

How can I make afni not build these things? Also, I guess I shoud add the packages that actually own those files as deps.

TheChymera:
What do I do about the file collision with fsl (dependency-wise)? AFNI does not depend on FSL, that's just plain silly. how can I find out if they are the ~same file or if the two packages just happen to give the same name to an executable?

TheChymera:
Ok, judging by a diff of the sources, it looks like FSL's version is newer (additional copyright note from 1999 compared to AFNI). I'd hazard a guess that FSL happens to ship an updated version of a file with essentially the same functionality (though the code looks quite different). No idea if the 1999 update is backwards-compatible, though. I'd propose to leave it in FSL, and remove it from AFNI and deal with any issues when they arise and if they arise. But, again, how do I keep AFNI from building this?

I guess I have to do something with these lines from Makefile.INCLUDE:

```
EXTRAS   = cjpeg djpeg whirlgif mpeg_encode myget Xphace rmz aiv mycat quotize $(EXPROGS)
whirlgif:whirlgif.h whirlgif.c
	$(CC) -o whirlgif whirlgif.c $(INFLAGS) $(LFLAGS) $(LLIBS)
```

But what?

kiwifb:
OK overall no big surprises. You already have some of those ebuild colliding in dependency. A first guess (after breakfast) is that you should remove them from the "EXTRAS" line. I am not sure that will prevent them from being built in the first place. Will have to look in "EXPROGS" as well.

If you know an upstream for whirlgif the best thing is to spin it in its own ebuild and have both fsl and afni to depend on it - and remove it from what they build.

kiwifb:
Oh boys. I can locate a source tarball http://hpux.connect.org.uk/hppd/hpux/Networking/WWW/whirlgif-3.04/ for whirlgif but this is as outdated as it gets. Everyone else has moved on to gifsicle.

TheChymera:
At least on [the AFNI side of things](https://github.com/afniHQ/AFNI/issues/6#issuecomment-89369837) it seems whirlgif is pretty much useless if you have gifsicle, and even if you do not, it simply disables a feature (which at least none of my colleagues ever knew about). We can provide it as a USE flag maybe. 

Can't we solve these collisions on the portage side of things? (like "do not install x")

TheChymera:
Not least of all, upstream has offered to pull a gentoo-specific Makefile or a Makefile customized for out Gentoo needs if we give them one. I think that would be a good alternative to all those sed calls? Maybe we can also more easily solve the collision and LIBDIR issues like that?

kiwifb:
Well I was thinking that we should write our own. It may take a bit of time to do that. What's more we'll want a specific `Makefile.INCLUDE` too.

TheChymera:
Why not put it all in one? So what is the next thing to do? I already have a [pull request with AFNI](https://github.com/afniHQ/AFNI/pull/4) for a new Makefile, but it is far from what we would need, I will put all our `Makefile.linux fedora_19_64` seds in there, but other than that could you post other relevant advice in the pull request?

kiwifb:
That's nice but really they need something like autoconf. They cannot multiply the Makefiles it is not sustainable and no one knows what's in them.

TheChymera:
Regarding our stable version, we still need to somehow sed the file collisions out. Is there really no way to simply tell portage to ignore (not try to write) collisions?

TheChymera:
Why doesn't [my latest commit](https://github.com/TheChymera/neurogentoo/commit/a5085e84cb84cb5d2748995b259c84017b639897) show up here?

sci-biology/afni/afni-0_pre20130927.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	sci-libs/gsl
+	dev-libs/expat
+	x11-libs/libXpm
+	media-libs/netpbm
+	media-video/mpeg-tools
+	x11-libs/libGLw
+	media-libs/qhull
+	sys-devel/llvm
+	media-video/mpeg-tools
+	media-libs/libjpeg-turbo"
```
jlec: Why are all of these buildtime dependencies? Most packages are only providing libs, so I would guess you also need them at runtime.

kiwifb: There is a lot of work to do with these. Quite a few of them are actually bundled and linked statically if I see things correctly. And while I am at it you only need `media-video/mpeg-tools` once in that list.

TheChymera: @kiwifb so, should I have them in both `RDEPEND` and `DEPEND`. I actually doubt we need them at build time :-?

kiwifb: You currently don't need most of these at build time because afni has a bundled copy - which is usually bad(TM). A prime example is they have a copy of jpeg6b. So not only will you have to put most of these in `RDEPEND` it will have to be unbundled.

To go back to jlec's point it is fairly uncommon for a library to be needed at run time but not at build time for software built from source. Exercise to the reader: describe a scenario in which it can happen.

TheChymera:
Actually I imagine it's fairly common for a lib to be needed at run but not build time: e.g. stuff which the program needs to run the gui, functions it calls via wrappers (like nipype, biopython etc. do). The only things I would expect a program to need at build time are system tools like the compiler, `sed`, etc.

In any case, is there any dependency at all, which I should leave under DEPEND rather than RDEPEND?

kiwifb:
Wrapping an executable is different from wrapping a library. Unless you uses your library like a plugin with dlopen, you will need the library at build time. You should have stuff both under DEPEND and RDEPEND, only pure build tools are not required at runtime.

TheChymera:
@kiwifb so I move *all* of the AFNI deps to RDEPEND, and then I add RDEPEND under DEPEND?

kiwifb:
As far as I can see, yes. But I was talking about replacing Makefile.INCLUDE the other day and some major surgery will be needed there to make it use system libraries and binaries instead of its own.

TheChymera:
@kiwifb you were also talking about upstream using autotools or similar. I am thinking maybe for the time being we should deal with everything as best we can on our end, put this in the repo, and update stuff to work with our new Makefiles or autotools or when(ever) that will be done. in any case we can only do that for the live ebuild and future releases. We already have an older release. The new Makefiles won't be in that archive anyway.

TheChymera:
I am pretty sure tcsh is actually needed at build time - I moved all the rest and re-linked as you advised, @kiwifb .

TheChymera:
Bump?

kiwifb:
Sure that's better and now we should start breaking down the rest. This is not going as is inside.

TheChymera:
What do you mean by breaking down the rest?

kiwifb:
I mean we cannot let it build against the internal copies of all these in `DEPEND`.

sci-biology/afni/afni-0_pre20130927.ebuild
jlec: media-libs/libjpeg-turbo should be virtual/jpeg.

sci-biology/afni/afni-9999.ebuild
```diff
@@ -0,0 +1,66 @@
...
+EAPI=5
+
+inherit multilib pax-utils git-r3 toolchain-funcs
+
+DESCRIPTION="An open-source environment for processing and displaying functional MRI data"
+HOMEPAGE="http://afni.nimh.nih.gov/"
+SRC_URI=""
+EGIT_REPO_URI="https://github.com/afniHQ/AFNI"
+
+
```
jlec: ebuild.minorsyn               1
   sci-biology/afni/afni-9999.ebuild: Useless blank line on line: 14

jlec:
@TheChymera you need to unbundle many libs. jpeg-6b is something we cannot support bundled.

jlec:
* QA Notice: command not found:
 * 
 * 	/var/tmp/portage/sci-biology/afni-9999/temp/environment: line 2787: CC: command not found

jlec:
Don't install all static archives.

TheChymera:
What do you mean by “don't install all static archives”?

kiwifb:
All the libraries that end up in `.a`.

jlec:
https://en.wikipedia.org/wiki/Static_library do not install them unconditionally.

TheChymera:
Any ideas why the build fails after my last commit?

jlec:
It's strange but not you fault. Should be fixed since some days.

TheChymera:
Ok, so from the wikipedia page you linked me to I understand that static libs are librairies which are included into the executables at compile-time. Yes?

But you say:

> Don't install all static archives.

They don't get installed themselves, do they? they only go into the executables that I compile - right? Did you mean that AFNI ships its own static libraries, and instead of building my executables with those, I should build them with static libraries provided by the proper packages?

TheChymera:
@kiwifb is: ```rm "${ED}/usr/$(get_libdir)/*.a"``` alongside the rm conflict what we are looking for here?

TheChymera:
@kiwifb so is that a no?

kiwifb:
We would like to be more sophisticated. Yes, you could remove them manually. But what we would actually would want to do:
 * not build them in the first place - so there wouldn't be anything to install
 * use system libraries instead - preferably shared one

So the makefiles and build system have to go under surgery.

TheChymera:
You would have no interest in doing that yourself, right? It seems to me upstream would pull, but not do this themselves either.

So, the only way forward is for me to learn how to re-write their Makefiles, yes? Where on earth can I start? o.0

kiwifb:
Like a lot of people: by experience. You probably know the basic, there are basic targets that are files to produce and there are rules and definitions.... Nothing like having a project to call your own.

TheChymera:
@kiwifb Rick Reynolds [recently told me that NeuroDebian has a CMake bui ld for AFNI](http://afni.nimh.nih.gov/afni/community/board/read.php?1,148643,148681#msg-148681). I found [this rather old post](http://afni.nimh.nih.gov/afni/community/board/read.php?1,125954,125954#msg-125954) on the AFNI forums, and I was able to navigate [to this repository](http://anonscm.debian.org/cgit/pkg-exppsy/afni.git/tree/) from the AFNI Neurodebian page...

Does it look like a CMake distribution? I think not, I can't see any `CMakeLists.txt` or other cmake-relevant file in there.

TheChymera:
@kiwifb  apparently [this](http://neuro.debian.net/pkgs/afni.html#comment-2075936195) is how we get the cmake buld. Is there any way to make this work from within an ebuild?

TheChymera:
@kiwifb updated the link in my last comment, can you have a look at this again?

sci-biology/afni/ChangeLog
```diff
@@ -0,0 +1,149 @@
...
+  afni's GH has moved
+
+  04 Jun 2015; Horea Christian <h.chr@mail.ru> afni-0_pre20130927.ebuild,
+  afni-9999.ebuild:
+  added new dep x11-libs/libXft
+
+  13 Apr 2015;  <chymera@gentoo.org> afni-9999.ebuild:
+  removed blank line
+
+  08 Apr 2015;  <chymera@gentoo.org> afni-0_pre20130927.ebuild,
```
jlec: Please fix this.

sci-biology/afni/afni-0_pre20130927.ebuild
```diff
@@ -0,0 +1,66 @@
...
+	x11-libs/libXpm
+	media-libs/netpbm
+	x11-libs/libGLw
+	media-libs/qhull
+	sys-devel/llvm
+	media-video/mpeg-tools
+	media-libs/libjpeg-turbo
+	x11-libs/libXft
+	x11-libs/libXi"
+
```
jlec: For convenience you could sort this.

sci-biology/afni/afni-0_pre20130927.ebuild
```diff
@@ -0,0 +1,66 @@
...
+RDEPEND="x11-libs/motif[-static-libs]
+	sci-libs/gsl
+	dev-libs/expat
+	x11-libs/libXpm
+	media-libs/netpbm
+	x11-libs/libGLw
+	media-libs/qhull
+	sys-devel/llvm
+	media-video/mpeg-tools
+	media-libs/libjpeg-turbo
```
jlec: Why not virtual/jpeg?

sci-biology/afni/afni-0_pre20130927.ebuild
```diff
@@ -0,0 +1,66 @@
...
+}
+
+src_compile() {
+	emake -j1 all plugins suma_exec
+}
+
+src_install() {
+	emake INSTALLDIR="${ED}/usr/bin" -j1 install install_plugins
+	emake INSTALLDIR="${ED}/usr/$(get_libdir)" -j1 install_lib
+	for CONFLICT in ${BIN_CONFLICTS[@]}; do
```
jlec: declare CONFLICT local.

TheChymera: what does that mean?, how do I do that?

sci-biology/afni/afni-0_pre20130927.ebuild
TheChymera: why would that be better?

sci-biology/afni/afni-0_pre20130927.ebuild
jlec: I am sure, you are able to dig through ebuild or use google to find out what local variables are. Please try to make at least a little effort to find information yourself.

sci-biology/afni/afni-0_pre20130927.ebuild
```diff
@@ -0,0 +1,66 @@
...
+HOMEPAGE="http://afni.nimh.nih.gov/"
+SRC_URI="http://dev.gentoo.org/~jlec/distfiles/${P}.tgz"
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS=""
+IUSE=""
+
+RDEPEND="dev-libs/expat
+	media-libs/libjpeg-turbo
```
jlec: Why not virtual/jpeg ?

TheChymera: why is that any better?

jlec: because virtual let you choose which jpeg implementation you like to have. Please inform yourself about virtuals.

sci-biology/afni/afni-0_pre20130927.ebuild
TheChymera: so I just write `local CONFLICT` before the loop? [this](http://lists.gnu.org/archive/html/bug-bash/2010-07/msg00104.html) seems to suggest I need a function, bur `src_install()` already is one, right?

sci-biology/afni/afni-0_pre20130927.ebuild
TheChymera: yes, `media-libs/libjpeg-turbo` is specifically neded.

sci-biology/afni/afni-0_pre20130927.ebuild
kiwifb: It just means the variable will not exist outside of `src_install()` which is a function.

sci-biology/afni/afni-0_pre20130927.ebuild
kiwifb: Just because that's what they use doesn't mean you have too. I don't think there are implementation specific calls in `libjpeg-turbo` but if you have proof that they use stuff that is not standard then OK.

TheChymera: AFNI ships two files which we package as part of libjpeg-turbo:

* `/usr/bin/cjpeg`
* `/usr/bin/djpeg`

 I am assuming afni will expect to find the things they ship it with. if all other things which `virtual/jpeg` could point to ship these files we can use that instead.

kiwifb: Are standard jpeg program as far as I understand ([c]ompression and [d]ecompression respectively). If possible don't call them with their absolute name. Those commands should be in your `$PATH` so calling `cjpeg` and `djpeg` is quite enough.

TheChymera: which means that they would be provided by any other incarnation of `virtual/jpeg`?

kiwifb: yes.

TheChymera:
Hi again, I'm having trouble applying the `cmake*` patches from [the neurodebian repo](http://anonscm.debian.org/cgit/pkg-exppsy/afni.git/tree/debian/patches?id=debian/0.20141224_dfsg.1-1). 

I am applying them [thus](https://github.com/gentoo-science/sci/pull/115/files#diff-796703fd2564d24b60b20100cfa85098R45), and I get no error from portage, however, afterwards, I see no `CMakeLists.txt` file, which is plainly needed for the cmake build.

However, these patches should clearly be creating that file, as can be seen [in this patch](http://anonscm.debian.org/cgit/pkg-exppsy/afni.git/tree/debian/patches/cmake_sharedlibs-cmake?id=debian/0.20141224_dfsg.1-1). What's wrong, why won't this work?

kiwifb:
But do you get a message that the patch is applied? `epatch` tells you about the patches it applies.

TheChymera:
@kiwifb It says "applying" and then the names of the files, and then "ok":

```
These are the packages that would be merged, in order:

Calculating dependencies... done!
[ebuild   R    ] sci-biology/afni-0_pre20141224::neurogentoo  0 KiB

Total: 1 package (1 reinstall), Size of downloads: 0 KiB

Would you like to merge these packages? [Yes/No] y

>>> Verifying ebuild manifests

>>> Emerging (1 of 1) sci-biology/afni-0_pre20141224::neurogentoo
>>> Unpacking source...
 * Fetching https://github.com/neurodebian/afni ...
git fetch https://github.com/neurodebian/afni +refs/tags/debian/0.20141224_dfsg.1-1:refs/tags/debian/0.20141224_dfsg.1-1
git symbolic-ref refs/git-r3/sci-biology/afni/0/__main__ refs/tags/debian/0.20141224_dfsg.1-1
 * Checking out https://github.com/neurodebian/afni to /var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224 ...
git checkout --quiet refs/tags/debian/0.20141224_dfsg.1-1
GIT update -->
   repository:               https://github.com/neurodebian/afni
   at the commit:            75f52fc45d518c381d8bc7ad941d26d1772de6a2
>>> Source unpacked in /var/tmp/portage/sci-biology/afni-0_pre20141224/work
>>> Preparing source in /var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224 ...
 * Applying cmake_sharedlibs-noncmake ...                                                               [ ok ]
 * Applying cmake_sharedlibs-cmake ...                                                                  [ ok ]
 * Applying cmake_tests ...                                                                             [ ok ]
>>> Source prepared.
>>> Configuring source in /var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224 ...
>>> Working in BUILD_DIR: "/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224_build"
 * Unable to locate CMakeLists.txt under:
 * "/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/CMakeLists.txt"
 * Consider not inheriting the cmake eclass.
 * ERROR: sci-biology/afni-0_pre20141224::neurogentoo failed (configure phase):
 *   FATAL: Unable to find CMakeLists.txt
 * 
 * Call stack:
 *     ebuild.sh, line   93:  Called src_configure
 *   environment, line 3247:  Called cmake-utils_src_configure
 *   environment, line  942:  Called _execute_optionally 'src_configure'
 *   environment, line  308:  Called enable_cmake-utils_src_configure
 *   environment, line 1152:  Called die
 *           die "FATAL: Unable to find CMakeLists.txt";
 * 
 * If you need support, post the output of `emerge --info '=sci-biology/afni-0_pre20141224::neurogentoo'`,
 * the complete build log and the output of `emerge -pqv '=sci-biology/afni-0_pre20141224::neurogentoo'`.
 * The complete build log is located at '/var/tmp/portage/sci-biology/afni-0_pre20141224/temp/build.log'.
 * The ebuild environment file is located at '/var/tmp/portage/sci-biology/afni-0_pre20141224/temp/environment'.
 * Working directory: '/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224'
 * S: '/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224'

>>> Failed to emerge sci-biology/afni-0_pre20141224, Log file:

>>>  '/var/tmp/portage/sci-biology/afni-0_pre20141224/temp/build.log'

 * Messages for package sci-biology/afni-0_pre20141224:

 * Unable to locate CMakeLists.txt under:
 * "/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/CMakeLists.txt"
 * Consider not inheriting the cmake eclass.
 * ERROR: sci-biology/afni-0_pre20141224::neurogentoo failed (configure phase):
 *   FATAL: Unable to find CMakeLists.txt
 * 
 * Call stack:
 *     ebuild.sh, line   93:  Called src_configure
 *   environment, line 3247:  Called cmake-utils_src_configure
 *   environment, line  942:  Called _execute_optionally 'src_configure'
 *   environment, line  308:  Called enable_cmake-utils_src_configure
 *   environment, line 1152:  Called die
 *           die "FATAL: Unable to find CMakeLists.txt";
 * 
 * If you need support, post the output of `emerge --info '=sci-biology/afni-0_pre20141224::neurogentoo'`,
 * the complete build log and the output of `emerge -pqv '=sci-biology/afni-0_pre20141224::neurogentoo'`.
 * The complete build log is located at '/var/tmp/portage/sci-biology/afni-0_pre20141224/temp/build.log'.
 * The ebuild environment file is located at '/var/tmp/portage/sci-biology/afni-0_pre20141224/temp/environment'.
 * Working directory: '/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224'
 * S: '/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224'
```

kiwifb:
What does `find /var/tmp/portage/sci-biology/afni-0_pre20141224/work/ -name CMakeLists.txt` says?

TheChymera:
@kiwifb ah, it seems a prefix was added along the way:

```
/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/b/html/CMakeLists.txt
/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/b/ptaylor/CMakeLists.txt
/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/b/coxplot/CMakeLists.txt
/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/b/pkundu/meica.libs/CMakeLists.txt
/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/b/pkundu/CMakeLists.txt
/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/b/3DEdge/src/CMakeLists.txt
/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/b/avovk/CMakeLists.txt
/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/b/rickr/CMakeLists.txt
/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/b/faces/CMakeLists.txt
/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/b/R_scripts/CMakeLists.txt
/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/b/python_scripts/CMakeLists.txt
/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/b/tests/CMakeLists.txt
/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/b/CMakeLists.txt
/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/b/poems/CMakeLists.txt
/var/tmp/portage/sci-biology/afni-0_pre20141224/work/afni-0_pre20141224/b/SUMA/CMakeLists.txt
```

How do I best deal with this?

kiwifb:
In this case `epatch -p1` should do the trick I believe. I don't think it is a good idea in general to add files by patch but you have to deal with what you have. And that probably simplify the maintenance to just use the debian patch rather than repacking it in an extra source tarball.

TheChymera:
@kiwifb thank you :) apparently that solved that issue, but now CMake fails.

I can't really find the error, could you help me out? Check out the [build log](https://bpaste.net/show/a8587dc6e16e) and the [CMake log](https://bpaste.net/show/ec56f673486b).

kiwifb:
Hum, it appears you need the gifti sources in a subdirectory and you don't have them. I think this is the main issue. If you have the source in question they may be in a different directory than what is expected.

TheChymera:
@kiwifb how can I look for them `find /var/tmp/portage/sci-biology/afni-0_pre20141224/work/ -name gifti` gives me naught.

kiwifb:
You won't find a directory that way with the `find` command (unless I am mistaken). Possibly `find $somepath -type d gifti`. Then you'll have to search the `CMakeList.txt` to find what it expect to include. And of course you may find nothing because there is nothing to find. 

You are using a devian patch, you should make sure that the debian recipe to build this doesn't imply adding an extra source in the mix.

TheChymera:
@kiwifb I added the cmake arguments the debian build uses in these lines](https://github.com/TheChymera/neurogentoo/blob/80465999149a4103f0a5c9127bd23925a3c8b43c/sci-biology/afni/afni-0_pre20141224.ebuild#L52-L72). Are there any lines among these that clearly do not make sense on Gentoo?

TheChymera:
@kiwifb  - how can I get the full set of options that the ebuild passes to cmake?

kiwifb:
Sorry I should have piped in at your last message. If you mean what's available as in the kind of output of `./configure --help` I don't think there is an equivalent. The only way to find out is to "prepare" the ebuild (`ebuild myebuild-ver.ebuild prepare`) create a build directory and then run `ccmake` or `cmake-gui` on the source directory.

TheChymera:
Any idea how I could set `NIFTI_INCLUDE_DIR`? I think that is currently [crashing my build](https://bpaste.net/show/7b6337d991fc).

PR #290, opened 2014-11-18
Update astropy-affliated packages

sargas:
Mostly fixes to not rdepend on astropy-helpers, since it is only needed at install time. A few version bumps intermixed.

Of note: dev-python/astroquery depends on >=astropy-0.4.1, which is not in the tree yet.


TheChymera:
@sargas do you have any updates on this?

heroxbd:
Let's close this for staled development.

PR #313, opened 2015-01-10
Fix InterMine QA issues

marbre:


licenses/InterMine
```diff
@@ -0,0 +1,452 @@
+InterMine is an open source project distributed under the GNU Lesser
+General Public Licence.
```
jlec: This is LGPL-2.1. So no need to add this license, but simply reuse the on already there.

marbre:
Didn't checked that well enough. It is actually LGPL-2.1+. Fixed.

PR #318, opened 2015-01-10
sci-biology/merlin: Move to EAPI=5

marbre:


jlec:
When moving to EAPI=5, please drop all unnecessary die and fix the usage of the appropriate phases.

marbre:
Eliminated the dies, but I think the usage of the phases is fine or not?

PR #319, opened 2015-01-10
ci-biology/rnaplex: Move to EAPI=5; Update homepage and SRC_URI.

marbre:
The URL to the src was outdated, therefore the manifest had to be updated.


sci-biology/rnaplex/rnaplex-0.2.ebuild
```diff
@@ -16,9 +16,9 @@ IUSE=""
 KEYWORDS="~amd64 ~x86"
 
 DEPEND=""
-RDEPEND=""
+RDEPEND="${DEPEND}"
 S=${WORKDIR}/${MY_P}
 
 src_install() {
-	emake DESTDIR="${D}" install || die "install failed"
```
jlec: You can drop the whole phase as this is the default implementation.

sci-biology/rnaplex/rnaplex-0.2.ebuild
```diff
@@ -16,9 +16,9 @@ IUSE=""
 KEYWORDS="~amd64 ~x86"
 
 DEPEND=""
-RDEPEND=""
+RDEPEND="${DEPEND}"
```
jlec: if DEPEND and RDEPEND are empty, simply drop them.

sci-biology/rnaplex/rnaplex-0.2.ebuild
```diff
@@ -1,14 +1,14 @@
...
-EAPI=2
+EAPI=5
 
 MY_P=RNAplex-${PV}
 
 DESCRIPTION="RNA-RNA interaction search"
-HOMEPAGE="http://www.tbi.univie.ac.at/~htafer/"
-SRC_URI="http://www.tbi.univie.ac.at/~htafer/RNAplex/${MY_P}.tar.gz"
+HOMEPAGE="http://www.bioinf.uni-leipzig.de/Software/RNAplex/"
+SRC_URI="http://www.bioinf.uni-leipzig.de/Software/RNAplex/RNAplex-0.2.tar.gz"
```
jlec: you can still use MY_P here which makes future updates easier.

marbre: Copy&Paste mistake while greping the new URI. Fixed.

marbre:
The new commit should fix this. Thanks for the support :+1:

PR #322, opened 2015-01-10
sci-libs/libfac: Move to EAPI=5, clean ebuilds

marbre:


sci-libs/libfac/libfac-3.1.0.ebuild
```diff
@@ -25,7 +25,5 @@ S="${WORKDIR}/${PN}"
 RESTRICT="mirror"
```
junghans: That should go, GPL-2 never implies a fetch/mirror restriction!

marbre:
Removed the mirror restriction.

PR #323, opened 2015-01-11
Fix and clean sci-libs/givaro ebuilds; Bump version

marbre:
I've moved the old givaro ebuils to EAPI=5, fixed the SRC_URI for 3.3.0 and bumped to version 3.8.0.


marbre:
Maybe https://bugs.gentoo.org/show_bug.cgi?id=227803 should be closed. If so, I'll remove pkg_postinst() from version < 3.8.0.

licenses/CeCILL-B
```diff
@@ -0,0 +1,515 @@
+
+CeCILL-B FREE SOFTWARE LICENSE AGREEMENT
```
jlec: Isn't this license in the tree? Please check that.

kiwifb:
As for closing the bug for givaro, I am OK with that. I am keeping the version that sage specifically wants in the sage-on-gentoo overlay. I will make sure the ebuild is in line with whatever happens here.

licenses/CeCILL-B
marbre: Did so before pushing ;-) CeCILL-B is neither available in gentoo-science nor in the tree.

marbre:
Okay, therefore I would remove the pkg_postinst() phase. Should we maybe add a comment to the ebuild themselfs, that these versions are required for sage? Might avoid a removal..

kiwifb:
3.7.1 is currently used by sage and linbox 1.3.2/fflas-ffpack-1.6.0 used by sage (with explicit version check in configure). After past experiences with the combo givaro/fflas-ffpack/linbox I won't even bother with trying newer version of givaro.

marbre:
Should I try to add 3.7.1 to the science overlay?

PR #329, opened 2015-01-11
sci-libs/symmetrica: Move to EAPI=5 and clean the ebuild

marbre:


kiwifb:
Hum... I have more patches in the sage-on-gentoo overlay. I didn't remember I put this in the science overlay. I think if you want to clean it up for inclusion, you should start from the sage-on-gentoo overlay version.

sci-libs/symmetrica/symmetrica-2.0.ebuild
```diff
@@ -31,11 +23,11 @@ src_prepare() {
...
-	cp "${FILESDIR}/makefile" "${S}/makefile"
+	cp "${FILESDIR}/makefile" "${S}/makefile" || die
 }
 
 src_install() {
-	emake DESTDIR="${D}" install || die "make install failed"
+	emake DESTDIR="${D}" install
 
 #	symmetrica's documentation are just text files with a ".doc" extension.
 	dodoc README *.doc
```
junghans: I would do:
```
default
use doc && dodoc *.doc
```

sci-libs/symmetrica/symmetrica-2.0.ebuild
```diff
@@ -11,19 +11,11 @@ HOMEPAGE="http://www.neu.uni-bayreuth.de/de/Uni_Bayreuth/Fakultaeten/1_Mathemati
 MY_P=SYM$(replace_all_version_separators '_')
 SRC_URI="http://www.neu.uni-bayreuth.de/de/Uni_Bayreuth/Fakultaeten/1_Mathematik_Physik_und_Informatik/Fachgruppe_Informatik/prof_diskrete_algorithmen/en/research/SYMMETRICA/${MY_P}_tar.gz -> ${MY_P}.tar.gz"
 
-MY_P=SYM$(replace_all_version_separators ' _')
```
junghans: I would use `MY_P=SYM${PV//./_}` instead, but that is matter of taste.

marbre: Sounds good.

marbre:
All done. Thanks for the review.

sci-libs/symmetrica/ChangeLog
```diff
@@ -2,6 +2,10 @@
 # Copyright 1999-2010 Gentoo Foundation; Distributed under the GPL v2
 # $Header: $
 
+  11 Jan 2015; Marius Brehler <marbre@linux.sungazer.de>
+  symmetrica-2.0.ebuild:
+  Move to EAPUI=5 and clean the ebuild
```
jlec: Typo

sci-libs/symmetrica/symmetrica-2.0.ebuild
```diff
@@ -1,29 +1,21 @@
...
 
-EAPI=2
+EAPI=5
 
 inherit versionator
 
 DESCRIPTION="A collection of routine to handle a variety of topics"
 HOMEPAGE="http://www.neu.uni-bayreuth.de/de/Uni_Bayreuth/Fakultaeten/1_Mathematik_Physik_und_Informatik/Fachgruppe_Informatik/prof_diskrete_algorithmen/en/research/SYMMETRICA/index.html"
-MY_P=SYM$(replace_all_version_separators '_')
+MY_P=SYM${PV//./_}
```
jlec: Why don't you use the eclass anymore?

sci-libs/symmetrica/ChangeLog
marbre: Fixed.

sci-libs/symmetrica/symmetrica-2.0.ebuild
marbre: This was a hint by @junghans (second comment on the outdated diff). Should I revert this or let it as is and additionally remove the inherit versionator (which I forgot to clean up).

junghans: Yeah, for simple stuff, like this, using `versionator` is overkill, but it is a matter of taste. If you keep it this way, drop `inherit versionator`.

marbre: Done.

sci-libs/symmetrica/symmetrica-2.0.ebuild
```diff
@@ -1,29 +1,19 @@
...
 SRC_URI="http://www.neu.uni-bayreuth.de/de/Uni_Bayreuth/Fakultaeten/1_Mathematik_Physik_und_Informatik/Fachgruppe_Informatik/prof_diskrete_algorithmen/en/research/SYMMETRICA/${MY_P}_tar.gz -> ${MY_P}.tar.gz"
 
-MY_P=SYM$(replace_all_version_separators ' _')
-
 LICENSE="MIT"
 SLOT="0"
 KEYWORDS="~amd64 ~ppc ~x86 ~ppc64"
 IUSE=""
```
junghans: `doc` missing here.

sci-libs/symmetrica/symmetrica-2.0.ebuild
```diff
@@ -31,12 +21,10 @@ src_prepare() {
...
+	cp "${FILESDIR}/makefile" "${S}/makefile" || die
 }
 
 src_install() {
-	emake DESTDIR="${D}" install || die "make install failed"
-
-#	symmetrica's documentation are just text files with a ".doc" extension.
-	dodoc README *.doc
+	default
+	use doc && dodoc *.doc
```
junghans: If this is the case, you could install them as `*.txt` instead, but up to you.

sci-libs/symmetrica/symmetrica-2.0.ebuild
junghans: As the saved files is arbitrary and  `MY_P` is only used once, you could use it in place.
```
SRC_URI="http://www.neu.uni-bayreuth.de/de/Uni_Bayreuth/Fakultaeten/1_Mathematik_Physik_und_Informatik/Fachgruppe_Informatik/prof_diskrete_algorithmen/en/research/SYMMETRICA/SYM${PV//./_}_tar.gz -> ${P}.tar.gz"
```
But again, that is up to you!

sci-libs/symmetrica/symmetrica-2.0.ebuild
```diff
@@ -31,12 +21,10 @@ src_prepare() {
 #	plus documentation that you can use as you wish in your programs.
 #	For sage and ease of use we make it into a library with the following
 #	makefile (developped by F. Bissey and T. Abbott (sage on debian).
-	cp "${FILESDIR}/makefile" "${S}/makefile"
+	cp "${FILESDIR}/makefile" "${S}/makefile" || die
```
kiwifb: I have more patches in the sage-on-gentoo overlay, it would be best if the ebuilds were in sync or only present in the sage-on-gentoo overlay.

junghans: I think, we should merge this request first and merge the changes from the sage-on-gentoo overlay in a 2nd one.

sci-libs/symmetrica/symmetrica-2.0.ebuild
marbre: It's done same in the [sage overlay](https://github.com/cschwan/sage-on-gentoo/blob/master/sci-libs/symmetrica/symmetrica-2.0-r4.ebuild), so I think we should leave it as is.

marbre:
I fixed the missing use flag. I agree that it might be easier to merge the changes from the sage-on-gentoo overlay in a second step.

sci-libs/symmetrica/symmetrica-2.0.ebuild
kiwifb: I am not fussy about that one but I prefer to stay close to upstream. Someone familiar with may expect .doc but I don't think they would be lost if it became .txt.

sci-libs/symmetrica/symmetrica-2.0.ebuild
kiwifb: Fair.

PR #330, opened 2015-01-11
Bring back (revised) matio ebuild

marbre:
I didn't noticed that sci-mathematics/scilab depoends on '<sci-libs/matio-1.5'.. Sorry for that.

Instead of using autotools.eclass, I replaced this by autotools-utils and bumped to EAPI=5. We are now closer to the ebuilds in the tree.


sci-libs/matio/matio-1.3.4.ebuild
```diff
@@ -0,0 +1,48 @@
...
+IUSE="examples fortran static-libs"
+SRC_URI="mirror://sourceforge/${PN}/${P}.tar.gz"
+#DEPEND="doc? ( app-doc/doxygen virtual/latex-base )"
+RDEPEND=""
+
+PATCHES=( "${FILESDIR}/${P}"-autotools.patch )
+
+src_configure() {
+	local myeconfargs=(
+		--docdir="${EPREFIX}/usr/share/doc/${PF}" \
```
jlec: no \ needed in an bash aray

sci-libs/matio/matio-1.3.4.ebuild
```diff
@@ -0,0 +1,48 @@
...
+		--enable-shared \
+		--disable-test \
+		$(use_enable fortran) \
+		#$(use_enable doc docs) \
+		$(use_enable static-libs static)
```
jlec: can be drop as it is automatically handled by the eclass

sci-libs/matio/matio-1.3.4.ebuild
```diff
@@ -0,0 +1,48 @@
...
+	)
+	autotools-utils_src_configure
+}
+
+src_install() {
+	autotools-utils_src_install
+	dodoc README ChangeLog
```
jlec: Standard docs which are handled by default

sci-libs/matio/matio-1.3.4.ebuild
```diff
@@ -0,0 +1,48 @@
...
+	#if use doc; then
+	#	insinto /usr/share/doc/${PF}
+	#	doins -r doxygen/html || die
+	#fi
+	if use examples; then
+		insinto /usr/share/doc/${PF}/examples
+		doins test/test* || die
```
jlec: drop die

junghans:
Just drop `sci-mathematics/scilab-5.[34].*` ,too.

marbre:
Should I do that and send a pull request or is this up to someone with more privileges?

PR #332, opened 2015-01-11
sci-misc/kaldi: New package

akreal:

PR #333, opened 2015-01-12
sci-libs/arrayfire: Pull master branch; Add git url; Remove keywords

marbre:
The [default branch](https://github.com/arrayfire/arrayfire/branches) of the arrayfire repository is devel. I doubt it is a good idea to pull this, so setting EGIT_BRANCH to pull the master.

Further, the git url was missing.

### DIFF
--- a/sci-libs/arrayfire/ChangeLog
+++ b/sci-libs/arrayfire/ChangeLog
@@ -1,6 +1,12 @@
-# ChangeLog for sys-cluster/mpe2
 # Copyright 1999-2015 Gentoo Foundation; Distributed under the GPL v2
 # $Header: $
+
+*arrayfire-0.9999 (12 Jan 2015)
+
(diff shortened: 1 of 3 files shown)


marbre:
As mentioned in #332 live ebuilds don't have keywords. Therefore removed the ~amd keyword.

junghans:
Looks good, maybe add a comment to the ebuild why we need to specify `master` explicitly. If the master branch is more like the release branch, you might want to change the version to `0.9999` and `9999` for the devel branch.

marbre:
Quite good idea :) I added a symlink (0.9999 -> 9999), as no further changes are needed.
Comment why to set EGIT_REPO was added.

Tagged versions/stable build will follow in early February (https://github.com/arrayfire/arrayfire/issues/321).

PR #336, opened 2015-01-13
Version bump of mendeleydesktop

marbre:
Mendeleydesktop 1.13 was released, this is a version bumped ebuild.

Further, I updated the license file. As these [terms of use](http://www.mendeley.com/terms/) don't contain the word EULA anymore, I rename the license file to Mendeley-terms.
Actually I am not sure if this is the correct license. The tarball itself contains a *LICENSE* file in *work/mendeleydesktop-1.13/* and an identical *License.txt* file  in *work/mendeleydesktop-1.13/share/doc/mendeleydesktop/*, which does not get installed. These are quite similar to the terms, but **not identical**.


sci-misc/mendeleydesktop/mendeleydesktop-1.13.ebuild
```diff
@@ -0,0 +1,114 @@
...
+
+QA_PREBUILT="/opt/mendeleydesktop/.*"
+
+pkg_nofetch() {
+	elog "Please download ${A} from:"
+	elog "http://www.mendeley.com/download-mendeley-desktop/"
+	elog "and move it to ${DISTDIR}"
+}
+
+src_unpack() {
```
junghans: I guess `src_unpack` from `inherit vcs-snapshot` would do the same.

marbre: *inherit vcs-snapshot* and afterwards calling *src_unpack* **or** calling *vcs-snapshot_src_unpack* explicitly in *src_unpack* results both in *work/mendeleydesktop-1.13-linux-x86_64*. That's not what I want.

junghans: I was thinking to remove `src_unpack` completely, which results in `vcs-snapshot_src_unpack` being called. But what I forgot was, that `vcs-snapshot.eclass` will unpack to the basename of `${A}`, so you would need to use the `->` operator, however that won't work as the tarball for x86 and amd64 need to be stored under different names.

So leave it like it is!

sci-misc/mendeleydesktop/mendeleydesktop-1.13.ebuild
```diff
@@ -0,0 +1,114 @@
...
+	if use amd64 || use amd64-linux ; then
+		mv -f "${MY_P_AMD64}" "${P}" || die
+	else
+		mv -f "${MY_P_X86}" "${P}" || die
+	fi
+}
+
+src_prepare() {
+	# remove bundled Qt libraries
+	rm -rf lib/mendeleydesktop/plugins \
```
junghans: `rm -f` will never fail, so either drop the `die` or remove the `-f`.

sci-misc/mendeleydesktop/mendeleydesktop-1.13.ebuild
```diff
@@ -0,0 +1,114 @@
...
+LICENSE="Mendeley-terms"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
+
+RESTRICT="fetch"
+
+DEPEND=""
+RDEPEND="
+	>=dev-qt/qtcore-4.6:4
```
junghans: `dev-qt/qtcore-4.6` is not even in gx86 anymore, just replace it by `dev-qt/qtcore:4`.

sci-misc/mendeleydesktop/mendeleydesktop-1.13.ebuild
marbre: You're right. Removed the force option.

sci-misc/mendeleydesktop/mendeleydesktop-1.13.ebuild
marbre: Fixed that, also for the other qt packages.

PR #338, opened 2015-01-14
Add nvidia-cuda-toolkit release candidate (7.0.18)

marbre:
This brings the CUDA 7.0 release candidate to gentoo-science. It is quite similar to the 6.5.* ebuilds, expect x86 support is fully dropped with this CUDA release.
I further updated the NVIDIA-CUDA license file and added a note in pkg_postinst that nvidia intends a newer driver version (only available in the CUDA rc bundle) for use with CUDA 7.0.


junghans:
I leave this one to @jlec!

PR #340, opened 2015-01-14
update sci-libs/libspatialindex to 1.8.5

hasufell:


sci-libs/libspatialindex/libspatialindex-1.8.5.ebuild
```diff
@@ -0,0 +1,37 @@
...
+	eautoreconf
+}
+
+src_configure() {
+	econf \
+		$(use_enable static-libs static) \
+		$(use_enable debug)
+}
+
+src_install() {
```
junghans: With `autotools-utils.eclass` this function would be automatic.

hasufell: I don't want to use autotools-utils.eclass. I prefer to have control over my ebuilds.

junghans: Sure, but if it is *your* ebuild, you better add yourself to `metadata.xml` ;-)

Btw, is there any reason to not add this ebuild to gx86 directly?

hasufell: > Btw, is there any reason to not add this ebuild to gx86 directly?

I could ask that question for this whole overlay, no? gx86 workflow sucks and is broken and there is no review process. I'm only adding things there if users complain or if it is overly useful in some way.

junghans: Well, I see this overlays as a place for user-maintained science-related ebuilds and ebuilds, which have QA issues. Anyhow, I will merge this request as soon as travis has verified it.

PR #344, opened 2015-01-16
New ebuild for minixml 2.9

chatcannon:
Small library, required for NeXus http://www.nexusformat.org/


dev-libs/minixml/minixml-2.9.ebuild
```diff
@@ -0,0 +1,23 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Header: $
+
+EAPI=5
+
+inherit eutils
```
jlec: not used.

dev-libs/minixml/metadata.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+<herd>sci</herd>
+<maintainer>
+  <email>cjk34@cam.ac.uk</email>
+  <name>Chris Kerr</name>
+  <description>I don't develop this software, I'm just a Gentoo user who wanted to install it</description>
+</maintainer>
+<longdescription>Mini-XML is a lightweight, ultra-portable XML library that requires only an ANSI C compiler and a 'make' program.</longdescription>
```
jlec: longline.

junghans:
Why don;t you call the package `mxml` ?

chatcannon:
mxml is an abbreviation, the name of the software is "Mini-XML". I suppose I could have added a hyphen to the package name.

junghans:
The reason why I was asking, is that the tarball name, and directory in there, is `mxml` and packages are usually named after that.

Also could you rebase/merge this change with latest master.

junghans:
Debian calls it `mxml`, too. So I would prefer that package name.

chatcannon:
Just submitted a new PR (#350) with the ebuild renamed to mxml

PR #347, opened 2015-01-17
sci-libs/factory: Drop old, move to EAPI=5 and add ebuild with support for static-libs use flag

marbre:
I think there is no need for factory-3.1.1 anymore, so it can be dropped.

The new factory-3.1.1-r2.ebuild is quite similar to factory-3.1.1-r1.ebuild, but the TODO was solved (hopefully in a acceptable way). As this installs less files by default, I think it is a good idea to add a new revision of the ebuild.
The ./configure doesn't allow to just disable building the static libs, so I decided to delete them from "${D}".
Doing so before autotools-utils_src_install, shows no effect, as autotools-utils_src_install just creates the static libs again.

Open minded for further improvements, so let me know what you think.


sci-libs/factory/factory-3.1.1-r2.ebuild
```diff
@@ -0,0 +1,47 @@
...
+src_configure() {
+	myeconfargs=( $(use_with singular Singular) )
+
+	autotools-utils_src_configure
+}
+
+src_install() {
+	autotools-utils_src_install
+
+	# We cannot disable the build of static libs, so delete after installed to ${D}
```
junghans: `autotools-utils_src_configure` add `--disable-static` automatically for `USE=-static-libs`, is this not enough?

kiwifb: No because it is not an option in that configure script. Those options are automatically added when you use libtool if I am not mistaken - and factory is not using libtool.
Is it still required to build macaulay2? If not it should just disappear. tomka was the main pusher for that stuff.

junghans: I see, if so, the comment should says that something like `--disable-static` won't disable the build of static libs as libtool isn't used.

marbre:
Thanks @kiwifb for the detailed explanation. I justed checked if the provided *configure* supports --disable-static, but I am not that deep into *configure* to know that this depends on the usage of libtool. I adjusted the comment.
Should I/Am i allowed to merge the pull request by myself or does this breaks the workflow?

kiwifb:
Depends, do you have submission right to git://git.overlays.gentoo.org/proj/sci.git ? If so you technically can, if you just have rights on github, abstain.

marbre:
Well I have the rights since a few days, that's the reason for the question, but I don't want to break the common workflow.
As I was not 100% certain with the changes, I created the pull request to get some feedback, instead of pushing my changes directly without review.

kiwifb:
Well I guess you'll have to do it at least once. Getting some feedback is good. Anyway, I have one last feedback:
If the only difference between r1 and r2 is the static-libs use flag I would remove r1 altogether, there is no point to it.

marbre:
1. macaulay2 downloads factory 4.0.0 itself, builds it, but don't installs it.
Anyway, sci-libs/libfac, depends and sci-mathematics/singular depend on factory.
2. I think you're right an we can drop the r1 ebuild too.

tom111:
Regarding Macaulay2: 
- It links the relevant parts of factory statically and installs nothing.
- Last time we checked, Singular and Macaulay2 needed different flavors of factory that were not compatible, so I let M2 build it internally.
- Macaulay2 ebuild is in a bad situation at the moment anyway. I can't bump at the moment because upstream say they need their own patched versions of mpfr/mpir built internally. I don't want to add that kind of stuff to the tree.
In total, I have not been following factory much, because Macaulay2 handles it pretty well. If you want to look at Macaulay2 anytime, let me know.

sci-libs/factory/factory-3.1.1-r2.ebuild
```diff
@@ -39,4 +37,12 @@ src_configure() {
...
 }
 
-# TODO: get rid of factories static libs ?
+src_install() {
+	autotools-utils_src_install
+
+	# Passing --disable-static to configure won't disable the build of static libs,
+	# as libtool isn't used. Therefore the static libs are deleted after installed to ${D}.
+	if ! use static-libs; then
+		find "${D}" -type f -name "*.a" -delete || die
```
junghans: Please use `${ED}`. And it could be shortened to `use !static-libs || find ... || die` if you like.

kiwifb:
Thanks for confirming tomka. I have been building singular with the internal factory as well, which gives me headaches for the factory bits of libsingular. But if you are using internal factory as well this ebuild may have no consumer left. Curse upstream for splitting a library that cannot be used as is for two different consumers.

sci-libs/factory/factory-3.1.1-r2.ebuild
marbre: I wasn't aware of ${ED}, as it is not documented in [variables section](http://devmanual.gentoo.org/ebuild-writing/variables/). I also think that using ${ED}=${D%/}${EPREFIX}/ or using ${D} directly doesn't matter in this case. Anyway, I adjusted the ebuild. Thanks for the advice.

marbre:
@kiwifb so do you think we should remove factory (and thus also libfac) from the overlay?

kiwifb:
Let's keep it around for a bit especially if a proper 4.x that is usable appears.

PR #351, opened 2015-01-20
dev-python/matlab_wrapper: Add to overlay

marbre:
This adds a matlab wrapper for python to the overlay.

We hopefully can add python3_3 support within the next days, by adding python3_3 to PYTHON_COMPAT. I will test this on work where I have MATLAB available. Possibly some patches are required to so. If this necessary, I will add them to the overlay till they find there way back into [matlab_wrapper branch](https://github.com/mrkrd/matlab_wrapper) on github.

I think I did everything the correct way, but as I am not very experienced with pyhon ebuilds, I would like to hear your feedback.


dev-python/matlab_wrapper/matlab_wrapper-9999.ebuild
```diff
@@ -0,0 +1,25 @@
...
+PYTHON_COMPAT=( python2_7 )
+
+inherit distutils-r1
+
+DESCRIPTION="MATLAB wrapper for Python"
+HOMEPAGE="https://github.com/mrkrd/matlab_wrapper"
+
+if [ ${PV} == "9999" ] ; then
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/mrkrd/matlab_wrapper.git git://github.com/mrkrd/matlab_wrapper.git"
```
junghans: Please use `${PN}` instead of `matlab_wrapper`.

marbre: You're right, just copied the URI from github without adjusting this... Fixed.

PR #352, opened 2015-01-20
dev-python/bokeh: version bump

marbre:
The commits add a version bumped ebuild for bokeh to the overlay.

I applied the diff of [#529986](https://bugs.gentoo.org/show_bug.cgi?id=529986), but adjusted the required use flag of *net-libs/nodejs* and added additional dies. I further ported the setup.patch from version 0.5.0 to 0.7.1.

This ebuild definitely needs some review. Every assistance is welcome.


dev-python/bokeh/files/bokeh-0.7.1-setup.patch
```diff
@@ -0,0 +1,10 @@
+--- setup.py	2015-01-12 19:37:38.000000000 +0100
```
junghans: Please write a header, why this is needed, also see <http://dev.gentoo.org/~vapier/clean-patches>!

dev-python/bokeh/bokeh-0.7.1.ebuild
```diff
@@ -0,0 +1,106 @@
...
+
+	if ! use chaco; then
+		sed -i -e "/'bokeh.chaco_gg'/d" setup.py || die
+	fi
+
+	sed -i -e "s/jsbuild = get_user_jsargs()/jsbuild = False/g" setup.py || die
+	cd bokehjs || die
+	npm install ||die
+	cd .. || die
+	python ./setup.py --build_js sdist || die
```
junghans: Why not `esetup.py` instead of `python ./setup.py`?

dev-python/bokeh/files/bokeh-0.7.1-setup.patch
marbre: Oh, I wasn't aware of that and actually nearly no of the patches if looked into so far come with a header. I will do this the other patches (other ebuilds) I pushed so far too.

dev-python/bokeh/bokeh-0.7.1.ebuild
marbre: Changed this on my local branch. Anyway this is not sufficient, as esetup.py pulls phantomjs. AFAIR this is prohibited in the all phases except fetch. I'll take a deeper look inside...

Just thought of a quick version bump for a co scientist, who is currently migrating from Debian to Gentoo ;)

marbre:
Adding phantomjs to RDEPEND was sufficient.

dev-python/bokeh/files/bokeh-0.7.1-setup.patch
junghans: There is always something new to learn ;-)

dev-python/bokeh/bokeh-0.7.1.ebuild
```diff
@@ -0,0 +1,107 @@
...
+DOCS=( CHANGELOG QUICKSTART.md README.md )
+
+PATCHES=( "${FILESDIR}"/${P}-setup.patch )
+
+DISTUTILS_NO_PARALLEL_BUILD=1
+
+python_prepare_all() {
+	# https://github.com/bokeh/bokeh/issues/1352
+	sed -e "s:        'bokeh.mplexporter',:&\n        'bokeh.mplexporter.renderers',:" \
+		-i setup.py || die
```
junghans: Looking at <https://github.com/bokeh/bokeh/blob/23fb4188090309945d5dba11dd6d4da34c744332/setup.py>, this fix seems to be upstream, now!

dev-python/bokeh/bokeh-0.7.1.ebuild
```diff
@@ -0,0 +1,107 @@
...
+
+	if ! use chaco; then
+		sed -i -e "/'bokeh.chaco_gg'/d" setup.py || die
+	fi
```
junghans: `bokeh.chaco_gg` does not appear in <https://github.com/bokeh/bokeh/blob/23fb4188090309945d5dba11dd6d4da34c744332/setup.py>, is chaco still supported?

dev-python/bokeh/bokeh-0.7.1.ebuild
```diff
@@ -0,0 +1,107 @@
...
+
+	sed -i -e "s/jsbuild = get_user_jsargs()/jsbuild = False/g" setup.py || die
+	cd bokehjs || die
+	npm install ||die
+	cd .. || die
+	esetup.py --build_js sdist
```
junghans: That raises a question, too. 4 lines above, `jsbuild` was hard-coded to `False`, which makes me wonder, why it is build here?

dev-python/bokeh/bokeh-0.7.1.ebuild
marbre: Thanks for checking, didn't find the time so far. Fixed.

dev-python/bokeh/bokeh-0.7.1.ebuild
marbre: I think they've ported some of the chaco features to bokeh, see the [FAQ](http://bokeh.pydata.org/en/latest/docs/faq.html#what-is-the-relationship-between-bokeh-and-chaco). I removed the dependencies and the use flag.

dev-python/bokeh/bokeh-0.7.1.ebuild
marbre: This was adopted from https://bugs.gentoo.org/show_bug.cgi?id=529986 ... I've no idea what exactly is happening ...

marbre: Okay, a look into [setup.py](https://github.com/bokeh/bokeh/blob/23fb4188090309945d5dba11dd6d4da34c744332/setup.py#L261) tells what is done:

```python
def get_user_jsargs():
print("""
Bokeh includes a JavaScript library (BokehJS) that has its own
build process. How would you like to handle BokehJS:
1) build and install fresh BokehJS
2) install last built BokehJS from bokeh/bokehjs/build
""")
mapping = {"1": True, "2": False}
```

junghans: I see, it is basically to overwrite the interactivity during installing.

dev-python/bokeh/bokeh-0.7.1.ebuild
```diff
@@ -0,0 +1,96 @@
...
+	# exclude server tests for now
+	nosetests -v \
+		-e multiuser_auth_test \
+		-e usermodel_test \
+		|| die
+}
+
+pkg_postinst() {
+	elog "websockets work only with python2_7"
+	local chaco_msg="chaco useflag doesn't affect python targets other than python2_7"
```
junghans: I guess, also this chaco stuff here can be removed, too.

dev-python/bokeh/bokeh-0.7.1.ebuild
```diff
@@ -0,0 +1,96 @@
...
+	cd "${BUILD_DIR}"/lib || die
+	# exclude server tests for now
+	nosetests -v \
+		-e multiuser_auth_test \
+		-e usermodel_test \
+		|| die
+}
+
+pkg_postinst() {
+	elog "websockets work only with python2_7"
```
junghans: Please inherit `readme.gentoo` and move this line to `python_prepare_all` as `DOC_CONTENTS="websockets work only with python2_7", add `readme.gentoo_create_doc` to  `python_install_all` and `readme.gentoo` will provide an `pkg_postinst` automatically.

dev-python/bokeh/bokeh-0.7.1.ebuild
marbre: Yes, didn't checked this well enough, sorry. Fixed.

dev-python/bokeh/bokeh-0.7.1.ebuild
marbre: Very good advice. Thanks a lot.

PR #353, opened 2015-01-27
sci-biology/discrover: bump package version, add USE flag for link-time ...

maaskola:
Changes: bump package version, add USE flag for link-time optimization, install doc to /usr/share/doc/${PF}

There's one further change that wasn't noted in the ChangeLog: the live ebuild has been updated to a newer git eclass.


sci-biology/discrover/discrover-1.5.2.ebuild
```diff
@@ -37,16 +37,19 @@ DEPEND="${RDEPEND}
...
 		media-gfx/imagemagick
 	)
+	lto? ( >=sys-devel/gcc-4.8 )
 "
 src_configure() {
 	local mycmakeargs=(
 		$(cmake-utils_use_with cairo CAIRO)
 		$(cmake-utils_use_with dreme DREME)
 		$(cmake-utils_use_with doc DOC)
+		$(cmake-utils_use_with lto LTO)
```
marbre: I still use gcc 4.7 by default on my machine, also version 4.8 is installed. If I try to compile discrover with *ebuild discrover-1.5.2.ebuild compile* on my maschine it fails with:

> CMake Error at CMakeLists.txt:85 (MESSAGE):
>  Disabled: link-time optimization - LTO support in GCC version < 4.8 is
>  unreliable
>
>
>-- Configuring incomplete, errors occurred!

I think it's better to let the ebuild fail during pkg_pretend() like realized for [arrayfire](https://github.com/gentoo-science/sci/blob/master/sci-libs/arrayfire/arrayfire-9999.ebuild#L51). So if gcc < 4.8 is used **and** *lto* is set, the ebuild should directly quit the installation. But thats only my humble opinion ;)

sci-biology/discrover/discrover-9999.ebuild
```diff
@@ -37,16 +37,19 @@ DEPEND="${RDEPEND}
...
 		media-gfx/imagemagick
 	)
+	lto? ( >=sys-devel/gcc-4.8 )
 "
 src_configure() {
 	local mycmakeargs=(
 		$(cmake-utils_use_with cairo CAIRO)
 		$(cmake-utils_use_with dreme DREME)
 		$(cmake-utils_use_with doc DOC)
+		$(cmake-utils_use_with lto LTO)
```
marbre: Same here.

sci-biology/discrover/discrover-1.5.2.ebuild
maaskola: That seems to be a good idea! I think I might do that.

My initial thought was that DEPENDing on >=sys-devel/gcc-4.8.0 conditionally on the USE flag lto (as is done in the ebuild now) might ensure usage of gcc >= 4.8. But I realized now that having emerged gcc >= 4.8 doesn't necessarily mean that it's configured to be used. Hence I'm currently emerging gcc < 4.8 to test that assumption. Anyway, I believe your suggestion is the solution.

sci-biology/discrover/discrover-9999.ebuild
maaskola: Indeed. I'll implement the same solution here.

maaskola:
Alright, I've used the suggested approach, and tested it. I switched between GCC 4.7.3-r1 and 4.8.3 and between USE lto enabled and disabled.
This should work fine now. Thanks for the suggestion!

maaskola:
Oops, please wait a second. I noticed I shouldn't be using upper case letters for the USE flag in the argument of `die`! I'll fix this and push again.

maaskola:
OK! Should be good to go!

sci-biology/discrover/discrover-1.5.2.ebuild
```diff
@@ -37,16 +37,28 @@ DEPEND="${RDEPEND}
...
 		$(cmake-utils_use_with tikzlogo TIKZLOGO)
 		$(cmake-utils_use_with misc_scripts MISC_SCRIPTS)
 		$(cmake-utils_use_with rmathlib RMATHLIB)
 		$(cmake-utils_use_with tcmalloc TCMALLOC)
+		-DDOC_DIR=/usr/share/doc/${PF}
```
junghans: I guess, you want "`-DDOC_DIR=${EPREFIX}/usr/share/doc/${PF}"`

maaskola: Yeah, after consulting the documentation, that seems like a good idea. I might do it, or something similar.
I'm still checking however, if it's not possible to use something like ${CMAKE_INSTALL_PREFIX}/share/doc/${PF} this would be the most appropriate, I think. I have to figure out if I can use that variable (it's certainly passed to cmake so I should be able to figure it out.) Pointers welcome! :)

maaskola: OK,
the [cmake eclass](http://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo-x86/eclass/cmake-utils.eclass?view=markup#l557) sets `CMAKE_INSTALL_PREFIX` equal to `${EPREFIX}${PREFIX}`. So I think I'll use
`-DDOC_DIR=${EPREFIX}${PREFIX}/share/doc/${PF}`

maaskola:
Alright, here we go again. Latest and greatest!

PR #354, opened 2015-01-27
New ebuild for Mantid

chatcannon:
Mantid is a GUI workspace for visualisation and data processing of neutron scattering and muon spectroscopy data.


sci-visualization/mantid/mantid-3.3.0.ebuild
```diff
@@ -0,0 +1,84 @@
...
+	sci-libs/opencascade
+"
+
+DEPEND="${RDEPEND}
+	dev-python/sphinx
+	doc?	( app-doc/doxygen
+		  dev-python/sphinx
+		  dev-python/sphinx-bootstrap-theme )
+	test?	( dev-util/cppcheck )
+"
```
marbre: I don't have much experience with the python-single-r1.eclass, but I think the use of [PYTHON_USEDEP](http://wiki.gentoo.org/wiki/Project:Python/python-single-r1#PYTHON_USEDEP) mandatory.

sci-visualization/mantid/files/mantid-3.3.0-minigzip-OF.patch
```diff
@@ -0,0 +1,15 @@
...
+ #  include <unix.h> /* for fileno */
+ #endif
+ 
++#ifndef OF
++#  define OF(args) args
++#endif
++
+ #ifndef WIN32 /* unlink already in stdio.h for WIN32 */
+   extern int unlink OF((const char *));
+ #else
```
marbre: @junghans suggested to clean patches as explained [here](http://dev.gentoo.org/~vapier/clean-patches). I think this would be a good idea.

sci-visualization/mantid/mantid-3.3.0.ebuild
```diff
@@ -0,0 +1,84 @@
...
+src_prepare() {
+	epatch	"${FILESDIR}/${P}-HAVE_IOSTREAM.patch" \
+		"${FILESDIR}/${P}-FindOpenCascade.patch" \
+		"${FILESDIR}/${P}-minigzip-OF.patch"
+}
+
+# pkg_setup() {
+# 	python-single-r1_pkg_setup
+# 	fortran-2_pkg_setup
+# }
```
marbre: If commented out, this can be dropped.

sci-visualization/mantid/files/mantid-3.3.0-minigzip-OF.patch
```diff
@@ -0,0 +1,15 @@
+diff --git a/MantidPlot/src/zlib123/minigzip.c b/MantidPlot/src/zlib123/minigzip.c
+index 97fbae1..5f4612f 100644
+--- a/MantidPlot/src/zlib123/minigzip.c
++++ b/MantidPlot/src/zlib123/minigzip.c
```
marbre: After a closer look to this patches, is it possible to DEPEND on sys-libs/zlib[minizip] and build against the system lib instead of shipping an outdated zlib version and build it internally?

chatcannon: No, it's not possible to depend on minizip because this is miniGzip -- it works on .gz files not .zip files. As far as I can tell, this uses system zlib as there is no bundled zlib. (I didn't include sys-libs/zlib in DEPEND because it is part of @system -- is that the right thing to do?)

sci-visualization/mantid/mantid-3.3.0.ebuild
chatcannon: I'm running Funtoo and I was worried about this conflicting with Funtoo's different Python dependency system. However, on testing it seems to work fine.

sci-visualization/mantid/files/mantid-3.3.0-minigzip-OF.patch
chatcannon: That link doesn't work but I found a version in Google's cache. As far as I can tell all it wants me to do is write a few lines at the top of the file describing what the patch is doing and why.

sci-visualization/mantid/files/mantid-3.3.0-FindOpenCascade.patch
```diff
@@ -0,0 +1,84 @@
...
+                 $ENV{CASROOT}/inc
+ )
+ 
+@@ -18,50 +11,56 @@ if ( WIN32 )
+   add_definitions ( -DWNT )
+ endif ( WIN32 )
+ 
+-set ( OC_REDHAT_RPM /opt/OpenCASCADE/lib64 )
++find_path ( OPENCASCADE_LIBRARY_DIR libTKernel.so PATHS
++               $ENV{CASROOT}/lib64
```
junghans: Could we just do the same by adding `-DOPENCASCADE_LIBRARY_DIR=/usr/$(getlib_dir)` to `mycmakeargs`?

sci-visualization/mantid/files/mantid-3.3.0-FindOpenCascade.patch
```diff
@@ -0,0 +1,84 @@
...
++)
++
+ find_library ( OPENCASCADE_LIB_TKERNEL
+                  NAMES TKernel
+-                 PATHS ${OC_REDHAT_RPM}         
++                 PATHS ${OPENCASCADE_LIBRARY_DIR}
```
junghans: Same here, just set `OPENCASCADE_LIB_TKERNEL` in `mycmakeargs`.

sci-visualization/mantid/files/mantid-3.3.0-HAVE_IOSTREAM.patch
```diff
@@ -0,0 +1,17 @@
...
+@@ -351,10 +351,8 @@ else ()
+   add_definitions ( -DNO_OPENCASCADE )
+ endif ()
+ 
+-# A few defines needed for OpenCascade on the Mac
+-if ( APPLE )
+-  add_definitions ( -DHAVE_IOSTREAM -DHAVE_LIMITS -DHAVE_IOMANIP )
+-endif ()
++# A few defines needed for OpenCascade
++add_definitions ( -DHAVE_IOSTREAM -DHAVE_LIMITS -DHAVE_IOMANIP )
```
junghans: Same here, if these defines are needed just add them to `CPPFLAGS`.

sci-visualization/mantid/files/mantid-3.3.0-minigzip-OF.patch
```diff
@@ -0,0 +1,15 @@
...
+@@ -50,6 +50,10 @@
+ #  include <unix.h> /* for fileno */
+ #endif
```
junghans: Why is the patch needed at all. Should `mantid` link against the system's `zlib` instead?

sci-visualization/mantid/mantid-3.3.0.ebuild
```diff
@@ -0,0 +1,84 @@
...
+
+S="${WORKDIR}/${P}-Source"
+
+src_prepare() {
+	epatch	"${FILESDIR}/${P}-HAVE_IOSTREAM.patch" \
+		"${FILESDIR}/${P}-FindOpenCascade.patch" \
+		"${FILESDIR}/${P}-minigzip-OF.patch"
+}
+
+# pkg_setup() {
```
junghans: Why is this commented out?

sci-visualization/mantid/files/mantid-3.3.0-FindOpenCascade.patch
chatcannon: What's `getlib_dir`? It doesn't exist as a command on my system and a duckduckgo search gave no relevant results.

sci-visualization/mantid/files/mantid-3.3.0-FindOpenCascade.patch
chatcannon: The path to the library depends on the installed version of opencascade (and on which one is selected by eselect-opencascade if there are more than one)

sci-visualization/mantid/files/mantid-3.3.0-HAVE_IOSTREAM.patch
chatcannon: That would define them for the entire project rather than for just these files

sci-visualization/mantid/files/mantid-3.3.0-minigzip-OF.patch
chatcannon: Read my reply to the comment from @marbre

sci-visualization/mantid/files/mantid-3.3.0-FindOpenCascade.patch
junghans: I meant `get_libdir`, which is part of `multilib.eclass`.

sci-visualization/mantid/files/mantid-3.3.0-minigzip-OF.patch
junghans: Sorry, I overlooked that.

sci-visualization/mantid/files/mantid-3.3.0-minigzip-OF.patch
junghans: In addition, this patch should be send upstream.

sci-visualization/mantid/files/mantid-3.3.0-FindOpenCascade.patch
chatcannon: OK, found it. There still needs to be a patch to replace the `set (OC_REDHAT_RPM ...` with a `find_path` or a `set( ... CACHE)` or something else that can be overridden with a `-D...` command-line argument.

sci-visualization/mantid/files/mantid-3.3.0-HAVE_IOSTREAM.patch
junghans: Good point, but whenever there is a chance to avoid patches, which needs porting with every release, by setting a couple of cmake variables we should go this way. And again it would be good to get these changes upstream.

chatcannon:
I have checked out the mantid github repository and will be sending some pull requests with these patches.

sci-visualization/mantid/files/mantid-3.3.0-FindOpenCascade.patch
junghans: I see, the selected version lives in `${CASROOT}`, right? That could be used in the ebuild as well. This patch can be dropped by using cmake's `find_library` build-in options to add search locations using `-DCMAKE_LIBRARY_PATH=${CASROOT}` and `-DFIND_LIBRARY_USE_LIB64_PATHS=<ON/OFF>`. See <http://www.cmake.org/cmake/help/v3.0/command/find_library.html> for details.

sci-visualization/mantid/mantid-3.3.0.ebuild
```diff
@@ -0,0 +1,79 @@
...
+		  dev-python/sphinx-bootstrap-theme[${PYTHON_USEDEP}] )
+	test?	( dev-util/cppcheck )
+"
+
+S="${WORKDIR}/${P}-Source"
+
+src_prepare() {
+	epatch	"${FILESDIR}/${P}-HAVE_IOSTREAM.patch" \
+		"${FILESDIR}/${P}-FindOpenCascade.patch" \
+		"${FILESDIR}/${P}-minigzip-OF.patch"
```
junghans: `cmake-utils.eclass` has support for `PATCHES`.

sci-visualization/mantid/mantid-3.3.0.ebuild
```diff
@@ -0,0 +1,79 @@
...
+	sci-libs/scipy[${PYTHON_USEDEP}]
+	dev-cpp/muParser
+	dev-libs/jsoncpp
+	dev-libs/openssl
+	sci-libs/opencascade[qt4]
+"
+
+DEPEND="${RDEPEND}
+	dev-python/sphinx
+	doc?	( app-doc/doxygen
```
junghans: Seeing below that `doc` get mapped on `DOT`, I guess, you want to depend on `app-doc/doxygen[dot]`.

sci-visualization/mantid/files/mantid-3.3.0-minigzip-OF.patch
junghans: Nice!

sci-visualization/mantid/mantid-3.3.0.ebuild
```diff
@@ -0,0 +1,86 @@
...
+		  dev-texlive/texlive-latexextra )
+	test?	( dev-util/cppcheck )
+"
+
+S="${WORKDIR}/${P}-Source"
+
+PATCHES=( "${FILESDIR}/${P}-minigzip-OF.patch" )
+
+src_configure() {
+	export CPPFLAGS="-DHAVE_IOSTREAM -DHAVE_LIMITS -DHAVE_IOMANIP ${CPPFLAGS}"
```
junghans: I would use `append-cppflags "-DHAVE_IOSTREAM -DHAVE_LIMITS -DHAVE_IOMANIP"` from `flag-o-matic.eclass` instead.

sci-visualization/mantid/mantid-3.3.0.ebuild
```diff
@@ -0,0 +1,86 @@
...
+	mycmakeargs=(	$(cmake-utils_use_enable doc QTASSISTANT)
+			$(cmake-utils_use_use doc DOT)
+			$(cmake-utils_use_no opencascade)
+			$(cmake-utils_use opencl OPENCL_BUILD)
+			$(cmake-utils_use_build shared-libs SHARED_LIBS)
```
junghans: In Gentoo build the shared libs is the default and there is the standard `static-libs` use flag to enable static libs.
So I would rewrite this to `$(cmake-utils_use_build !static-libs SHARED_LIBS)`.

sci-visualization/mantid/mantid-3.3.0.ebuild
```diff
@@ -0,0 +1,86 @@
...
+			$(cmake-utils_use_use tcmalloc TCMALLOC)
+			$(cmake-utils_use paraview MAKE_VATES)
+			$(cmake-utils_use_build test TESTING)
+			-DCMAKE_PREFIX_PATH="${CASROOT}"
```
junghans: Maybe adding a default like `${CASROOT:=/usr/lib}` would be useful.

junghans: Actually I would just add a line like
`[[ -z ${CASROOT} ]] && die "CASROOT environment variable not defined, that usually means you need to use 'eselect opencascade'."`

PR #356, opened 2015-01-31
Enable testing for sci-visualization/mantid

chatcannon:
The ebuild I recently submitted had `RESTRICT=test` -- this PR enables running the tests that aren't broken.

### DIFF
--- a/sci-visualization/mantid/mantid-3.3.0.ebuild
+++ b/sci-visualization/mantid/mantid-3.3.0.ebuild
@@ -87,30 +87,30 @@ src_test() {
 	# Tests are not built by default
 	emake AllTests
 	# Run only the tests that work without data files or GUI access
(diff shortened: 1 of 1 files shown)


sci-visualization/mantid/mantid-3.3.0.ebuild
```diff
@@ -85,7 +84,45 @@ src_configure() {
 }
 
 src_test() {
+	pushd "${BUILD_DIR}" # For some reason this is not done automatically...
```
junghans: If you are not using `cmake-utils_src_test`, you will have to do that manually.
You could try to put the tests in the `myctestargs` array.

TheChymera:
mantid is no longer distributed by Gentoo Science. If the package is still relevant we could bring it back pending a PR.

PR #364, opened 2015-02-06
Fix ebuild for librecad 2.0.7

j-g-:
- Fix the ${S} variable, build fails with wrong non-existent source directory
- Some improvements, and add tools USE flag present in 2.0.6 from the main portage tree.

### DIFF
--- a/media-gfx/librecad/librecad-2.0.7.ebuild
+++ b/media-gfx/librecad/librecad-2.0.7.ebuild
@@ -24,6 +24,7 @@ DEPEND="
 	media-libs/freetype"
 
 RDEPEND="${DEPEND}"
+S="${WORKDIR}/LibreCAD-${PV}"
 
 src_prepare() {
 	# currently RS_VECTOR3D causes an internal compiler error on GCC-4.8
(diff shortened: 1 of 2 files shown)


media-gfx/librecad/librecad-2.0.7.ebuild
```diff
@@ -12,8 +12,8 @@ SRC_URI="https://github.com/LibreCAD/LibreCAD/archive/${PV/_/}.zip -> ${P}.zip"
 
 LICENSE="GPL-2"
 SLOT="0"
-KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
```
jlec: no need to drop the prefix keywords or is it broken there?

j-g-: You are right, I didn't knew they were for prefix, I couldn't find info after a google search, and tought they were redundant, but I was just reading the arch.list file (I'm still learning about writing ebuilds).
Do I send it again without that?

jlec: Just add them back and update your pull request.

PR #360, opened 2015-02-04
Bump sci-biology/ugene version to 1.15.1

stharward:
The source site (unipro.ru) no longer has a downloadable tarball for
1.13.1, the most recent version in this repo.

Cursory testing (compiles, merge, run) on amd64.

PR #365, opened 2015-02-10
UGENE: removed defunct ebuild, updated source url

TheChymera:
The old sources are no longer available, the new sources are under the updated URL.


junghans:
* `ChangeLog` entry missing
* Already in review as #360

TheChymera:
:+1:

PR #367, opened 2015-02-13
dev-util/smartgit version bump to 6.5.6

LMolr:
In the update, they removed some of the `hg` suffixes.

The launcher requires `SWT_GTK3` env var to be set:
```ShellSession
SWT_GTK3=0 smartgit
```


jlec:
Always user repoman or repo-commit to commit your changes. Plain git will not work.

LMolr:
Done in: http://git.io/Ngce

PR #368, opened 2015-02-13
Pythontex bump

TheChymera:
Bumped pythontex, made some minor changes to the live ebuild.


dev-tex/pythontex/pythontex-0.14.ebuild
```diff
@@ -0,0 +1,67 @@
...
+
+DEPEND="${PYTHON_DEPS}
+	app-text/texlive"
+RDEPEND="${DEPEND}
+	dev-texlive/texlive-xetex
+	>=dev-python/matplotlib-1.2.0[${PYTHON_USEDEP}]
+	highlighting? ( dev-python/pygments[${PYTHON_USEDEP}] )"
+
+src_prepare() {
+	S="${WORKDIR}/${P}/${PN}"
```
jlec: put this in global scope

dev-tex/pythontex/pythontex-0.14.ebuild
```diff
@@ -0,0 +1,67 @@
...
+	cd "${S}" || die
```
jlec: unneeded.

dev-tex/pythontex/pythontex-0.14.ebuild
```diff
@@ -0,0 +1,67 @@
...
+}
+
+src_compile() {
+	ebegin "Compiling ${PN}"
+	latex ${PN}.ins extra > "${T}"/build-latex.log || die "Building style from ${PN}.ins failed"
+	eend
+}
+
+src_install() {
+	python_optimize .
```
jlec: this needs to run after installation on the installed files.

dev-tex/pythontex/pythontex-0.14.ebuild
```diff
@@ -0,0 +1,67 @@
...
+	else
+		python_newscript pythontex2.py pythontex.py
+		python_doscript pythontex_2to3.py
+		python_newscript depythontex2.py depythontex.py
+	fi
+
+	python_export PYTHON_SCRIPTDIR
+
+	python_moduleinto ${PYTHON_SCRIPTDIR}
+	python_domodule "${S}"/pythontex_engines.py "${S}"/pythontex_utils.py
```
jlec: What are you trying to do with this whole block?

dev-tex/pythontex/pythontex-0.14.ebuild
```diff
@@ -0,0 +1,67 @@
...
+
+	insinto /usr/share/texmf-site/tex/latex/pythontex/
+	doins "${S}"/pythontex.sty
+
+	insinto /usr/share/texmf-site/source/latex/pythontex/
+	doins "${S}"/pythontex.dtx "${S}"/pythontex.ins
```
jlec: you should use latex-package_src_doinstall() for those installations.

jlec:
And ChangeLog entry is missing.

dev-tex/pythontex/pythontex-0.14.ebuild
```diff
@@ -0,0 +1,67 @@
...
+
+EAPI=5
+
+PYTHON_COMPAT=( python{2_7,3_2,3_3} )
+
+inherit latex-package python-single-r1
+
+DESCRIPTION="Fast Access to Python from within LaTeX"
+HOMEPAGE="https://github.com/gpoore/pythontex"
+SRC_URI="https://github.com/gpoore/${PN}/archive/v${PV}.zip"
```
jlec: Please rename the downloaded package to some sane naming:

-> ${P}.zip

dev-tex/pythontex/pythontex-0.14.ebuild
```diff
@@ -0,0 +1,67 @@
...
+
+SLOT="0"
+LICENSE="LPPL-1.3 BSD"
+KEYWORDS="~amd64"
+IUSE="highlighting"
+
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
+
+DEPEND="${PYTHON_DEPS}
```
jlec: unzip is missing to unzip the source zip file.

dev-tex/pythontex/pythontex-0.14.ebuild
TheChymera: It's not my repo/project. We deal with what upstream offers - right?

dev-tex/pythontex/pythontex-0.14.ebuild
TheChymera: Installing the python scripts/modules from this package? Is this not the right way to do it? it's pretty much exactly like v0.12, and you (or maybe someone else) pulled that.

dev-tex/pythontex/pythontex-0.14.ebuild
TheChymera: how?

jlec: At the end fo src_install either

python_optimize

or

python_optimize "${ED}"/path/to/python/files

dev-tex/pythontex/pythontex-0.14.ebuild
jlec: if those files should end in site-dir, just python_domodule. In case those are scripts, just do python_doscripts. the whole python_export/python_moduleinto doesn't make sense.

dev-tex/pythontex/pythontex-0.14.ebuild
jlec: So deal with it.

Look for "Customization of Output File Names in SRC_URI" in
http://devmanual.gentoo.org/ebuild-writing/eapi/index.html

dev-tex/pythontex/pythontex-0.14.ebuild
```diff
@@ -0,0 +1,68 @@
...
+	app-text/texlive
+	app-arch/unzip"
+RDEPEND="${DEPEND}
```
jlec: unzip is not a runtime dependency

dev-tex/pythontex/pythontex-0.14.ebuild
TheChymera: what if I want the modules/scripts to not have the python version denominator at the end? that's why I'm doing the entire if statement thing...

dev-tex/pythontex/pythontex-0.14.ebuild
TheChymera: Where can I find some examples/documentation on how to use that - qgrep could only show me stuff like `latex-package_src_doinstall styles`

dev-tex/pythontex/pythontex-0.14.ebuild
jlec: What do you mean with that? Please give an example.

dev-tex/pythontex/pythontex-0.14.ebuild
jlec: man latex-package

dev-tex/pythontex/pythontex-0.14.ebuild
TheChymera: iirc the if statement should ensure that the pythontex.py entry in my path will point to an installed instance of either the python3.py or python2.py file from upstream. What do you mean by an example? Any example of what?

dev-tex/pythontex/pythontex-0.14.ebuild
TheChymera: So I just replace this line with `latex-package_src_doinstall`? It automatically figures out what to do in the dir?

jlec: lines 57-62 can probably be replaced by

latex-package_src_doinstall ${PN}.{dtx,ins,sty}

dev-tex/pythontex/pythontex-0.14.ebuild
jlec: + python_export PYTHON_SCRIPTDIR
+
+ python_moduleinto ${PYTHON_SCRIPTDIR}
+ python_domodule "${S}"/pythontex_engines.py "${S}"/pythontex_utils.py

this makes no sense. What should it do? where should those files and in?

jlec:
And start using repoman.

dev-tex/pythontex/pythontex-0.14.ebuild
TheChymera: @jlec they should end up in `${PYTHON_SCRIPTDIR}` ? Why does that make no sense?

TheChymera:
@jlec `repoman commit` gives me this:

```
zenbookhost pythontex # repoman commit
!!! Section 'neurogentoo' in repos.conf has name different from repository name 'science' set inside repository

*** the local copy of metadata.dtd needs to be refetched, doing that now

--2015-02-16 12:52:19--  http://www.gentoo.org/dtd/metadata.dtd
Resolving www.gentoo.org... 89.16.167.134, 2001:41c8:0:936::136, 2001:41c8:0:936::139
Connecting to www.gentoo.org|89.16.167.134|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 4720 (4.6K) [application/xml-dtd]
Saving to: ‘/usr/portage/distfiles/metadata.dtd.Lala7h’

/usr/portage/distfiles/metad 100%[=============================================>]   4.61K  --.-KB/s   in 0s     

2015-02-16 12:52:19 (524 MB/s) - ‘/usr/portage/distfiles/metadata.dtd.Lala7h’ saved [4720/4720]


>>> Creating Manifest for /home/chymera/src/neurogentoo/dev-tex/pythontex

Note: use --include-dev (-d) to check dependencies for 'dev' profiles

RepoMan sez: "Doing nothing is not always good for QA."

(Didn't find any changed files...)
```

Is this ok? How should I use repoman in my workflow? Just run `repoman commit` before I start a pull request here? Use it after every commit? 

Also, where do I find the "travis output"? if I click on details and then on the two red links in the middle I see a wall of code, but with no obvious errors in it.

dev-tex/pythontex/pythontex-0.14.ebuild
```diff
@@ -0,0 +1,64 @@
...
+	if python_is_python3; then
+		python_newscript pythontex3.py pythontex.py
+		python_newscript depythontex3.py depythontex.py
+	else
+		python_newscript pythontex2.py pythontex.py
+		python_doscript pythontex_2to3.py
+		python_newscript depythontex2.py depythontex.py
+	fi
+
+	python_export PYTHON_SCRIPTDIR
```
jlec: Do you mean /usr/lib/python-exec/${impl} ? If so, then just use python_doscript.

jlec:
Always use repoman commit instead of git commit in your workflow. 

And here you can find travis for your PR http://dev.gentoo.org/~jlec/paste/travis.png .

TheChymera:
@jlec - I think this is done now? I fixed those `domodule` calls. And ran repoman commit. Cn you pull it now?

TheChymera:
sooooo? Did I address everything, @jlec?

jlec:
Drop all but the latest version and check for missing die.

TheChymera:
@jlec Looking good now? Where should I check for missing die? What lines need die? I heard most of the ebuild functions have die built-in; and it looks like I'm using ebuild-specific commands on every line.

dev-tex/pythontex/pythontex-0.14.ebuild
```diff
@@ -0,0 +1,64 @@
...
+		python_domodule pythontex3.py
+		python_domodule depythontex3.py
+	else	
+		python_domodule pythontex2.py
+		python_domodule depythontex2.py
+		python_domodule pythontex_2to3.py
+	fi
+
+	python_domodule pythontex_engines.py 
+	python_domodule pythontex_utils.py
```
jlec: All those python_domodule should be python_doscripts. How did you use these scripts?

dev-tex/pythontex/pythontex-0.14.ebuild
```diff
@@ -0,0 +1,64 @@
...
+}
+
+src_compile() {
+	ebegin "Compiling ${PN}"
+	latex ${PN}.ins extra > "${T}"/build-latex.log || die "Building style from ${PN}.ins failed"
+	eend
+}
+
+src_install() {
+	python_setup
```
jlec: no need for this.

dev-tex/pythontex/pythontex-0.14.ebuild
```diff
@@ -0,0 +1,64 @@
...
+	if python_is_python3; then
```
jlec: This logic isn't working.

dev-tex/pythontex/pythontex-0.14.ebuild
```diff
@@ -0,0 +1,64 @@
...
+	python_domodule pythontex_engines.py 
+	python_domodule pythontex_utils.py
+
+	latex-package_src_doinstall ${PN}.{dtx,ins,sty}
+
+	latex-package_src_install
+
+	dodoc README
+	mktexlsr
+	python_optimize
```
jlec: No need for this, no modules installed.

dev-tex/pythontex/pythontex-0.14.ebuild
jlec: You are missing to install pythontex.py

TheChymera: where should I put them? currently I do not seem to be needing them - `pythontex` can be run from the command line just like that :-/ no idea how come, though....

dev-tex/pythontex/pythontex-0.14.ebuild
TheChymera: how come? what can I replace it with? depending on what python I use I need to install different files.

dev-tex/pythontex/pythontex-0.14.ebuild
jlec: With this ebuild you cannot run `pythontex` from the command line. There needs to be something different on your system. Please read the documentation and the code from upstream, these are the executable scripts which import from the modules you are installing above.

dev-tex/pythontex/pythontex-0.14.ebuild
```diff
@@ -0,0 +1,63 @@
...
+}
+
+src_compile() {
+	ebegin "Compiling ${PN}"
+	latex ${PN}.ins extra > "${T}"/build-latex.log || die "Building style from ${PN}.ins failed"
+	eend
+}
+
+src_install() {
+	if python_is_python3; then
```
jlec: This only works on a single python ABI installation not on a multi ABI as you are intending to do.

TheChymera: how can you tell that I want to make it a multi-ABI thing? how can I switch it to single-ABI? what would I need to chnge for it to be multi-ABI compatible? just install all the files?

dev-tex/pythontex/pythontex-0.14.ebuild
TheChymera: True, apparently texlive also provides the file? 0.o

```
zenbookhost chymera # equery belongs /usr/bin/pythontex 
 * Searching for /usr/bin/pythontex ... 
app-text/texlive-core-2014-r2 (/usr/share/texmf-dist/scripts/pythontex/pythontex.py)
app-text/texlive-core-2014-r2 (/usr/bin/pythontex -> ../share/texmf-dist/scripts/pythontex/pythontex.py)
```

What do I do in this situation? is this ebuild obsolete then? I would say not, because just with what texlive gives me I get:

```
chymera@zenbookhost ~ $ pythontex
  File "/usr/bin/pythontex", line 42, in <module>
    import pythontex2 as pythontex
ImportError: No module named pythontex2
```

jlec: You need to investigate further. Here my strategy:

1. check "/usr/bin/pythontex" -> it is the same as in the python tex package, so we can simply depend on texlive-core, as this is needed anyways.
2. Find out whether the main tex distribution installs all files -> yes in .../scripts/
2. check whether some package installs pythontex2.py & pythontex3.py -> so far I cannot find any so we need to pack this package. Or file a bug against  texlive-core for installing a broken package.

Strategy clear? Then go ahead with this package.

jlec: I personally would fix the texlive-core package as it needs have fixed python support anyways.

jlec: pythontex is dropped from texlive-core now. We need to install it here too.

TheChymera: you mean in the official portage tree? (it seems we do not have texlive-core here).

If so, could you also close [my bug report](https://bugs.gentoo.org/show_bug.cgi?id=542690)?

TheChymera: I think I solved the missing files issue? I have no idea if I pu them in the correct places, though - Upstream reccomends to put them in `texmf-dist/scripts/pythontex` - what do you think? https://github.com/gpoore/pythontex/issues/57#issuecomment-77781129

jlec: You are not using repoman!

jlec:
fixed in 6de5dbd94259c5955fdfdcd2510f034ea444097d

PR #369, opened 2015-02-13
projects/sci: dev-util/smartgit: Version Bump (6.0.6 -> 6.5.6)

LMolr:


dev-util/smartgit/smartgit-6.5.6.ebuild
```diff
@@ -23,7 +23,7 @@ RESTRICT="fetch"
 RDEPEND=">=virtual/jre-1.7:1.7"
```
junghans: Is `git` or `hg` needed a runtime? Or add a use flag to pull them in like `app-portage/layman` does.

dev-util/smartgit/smartgit-6.5.6.ebuild
```diff
@@ -7,11 +7,11 @@ EAPI=5
 inherit eutils java-utils-2 versionator
 
 MY_PV="$(replace_all_version_separators _)"
```
junghans: For these simple things, I prefer `MY_PV=${PV//./_}`, but up to you.

jlec: But it is perfectly fine to use the eclass functions here.

dev-util/smartgit/smartgit-6.5.6.ebuild
LMolr: The application starts without any `hg` or `git` executable installed.
However, it relies on them to perform vcs operations.
Therefore i think that `git` and `hg` USE flags (maybe defaulted to enabled?) are the right choice.

Since I am not familiar with ebuild writing, I have just updated the most straightforward things. 
I'll look into adding USE flags if they are needed.

junghans: No worries, that is why we have travis and doing a review!

Just add `hg` and `git` into `IUSE` and something like `hg? ( dev-vcs/mercurial )` to `RDEPEND`.

dev-util/smartgit/smartgit-6.5.6.ebuild
```diff
@@ -7,23 +7,27 @@ EAPI=5
...
 DESCRIPTION="Git client with support for GitHub Pull Requests+Comments, SVN and Mercurial"
 HOMEPAGE="http://www.syntevo.com/smartgit"
-SRC_URI="http://www.syntevo.com/download/${PN}hg/${MY_P}.tar.gz"
+SRC_URI="http://www.syntevo.com/download/${PN}/${MY_P}.tar.gz"
 
 SLOT="0"
 LICENSE="smartgit"
 KEYWORDS="~amd64 ~x86"
-IUSE=""
+IUSE="git mercurial"
```
jlec: This is not the correct way. Pure optional runtime packages should not be added as USE based dependency but you should either hard depend on it or use optfeature() from eutils.eclass

jlec:
Thanks for your work!

PR #391, opened 2015-03-21
Updated ebuild for sci-biology/igv

ercpe:
Fixed upstream bug https://github.com/broadinstitute/IGV/issues/114 as requested by Martin Mokrejŝ (note: i've excluded the test code by passing -Dnotests=true to ant so it may break at some point).

Also improved the ebuild by unbundling stuff we already have in portage. There are quite a number of libraries which aren't used for compilation which may be needed at runtime. I haven't tested the program in depth: it started at didn't throw errors when i clicked around - everything else is way beyond my knowledge.


jlec:
Thanks for your help!!

PR #393, opened 2015-03-28
Updated emboss-r1.eclass and all embassy/emboss packages to confirm to QA

SoapZaNet:
emboss-r1.eclass has been modified to address criticisms made in
http://article.gmane.org/gmane.linux.gentoo.devel/94991

### DIFF
--- a/eclass/emboss-r1.eclass
+++ b/eclass/emboss-r1.eclass
@@ -38,16 +38,13 @@
 # @ECLASS-VARIABLE: EBO_EXTRA_ECONF
 # @DEFAULT_UNSET
 # @DESCRIPTION:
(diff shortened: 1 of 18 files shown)


eclass/emboss-r1.eclass
```diff
@@ -38,17 +38,14 @@
 # @ECLASS-VARIABLE: EBO_EXTRA_ECONF
 # @DEFAULT_UNSET
 # @DESCRIPTION:
-# Extra config options passed to econf, similar to EXTRA_ECONF.
+# Extra config options passed to econf, similar to EXTRA_ECONF
+# but can be set by the ebuild.
```
jlec: Drop the reference to EXTRA_ECONF here. That causes confusion.

PR #396, opened 2015-03-29
Grass:7 major version bump. Added sci-geosciences/laszip and sci-geosciences/liblas

SoapZaNet:
sci-geosciences/grass-7.0.0 has 3 new optional USE flags: openmp, las and netcdf
USE=las includes building with the liblas library, which itself requires laszip.


sci-geosciences/grass/grass-7.0.0.ebuild
```diff
@@ -0,0 +1,303 @@
...
+KEYWORDS="~amd64 ~x86"
+IUSE="X cairo cxx ffmpeg fftw gmath jpeg las motif mysql netcdf nls odbc opengl openmp png postgres python readline sqlite tiff truetype wxwidgets"
+
+TCL_DEPS="
+	>=dev-lang/tcl-8.5:0
+	>=dev-lang/tk-8.5:0
+	"
+
+RDEPEND="
+	${PYTHON_DEPS}
```
jlec: this should be 

python?( ${PYTHON_DEPS} )

jlec:
@wraeth @idella Could you please check if grass-7.0.1 can also support liblas/laszip?

wraeth:
@jlec grass-7.0.1 does support liblas (though disabled in grass::gentoo as liblas is not in the main tree). Grass doesn't link laszip directly but supports .laz files if liblas is built with it.

It should also be noted that this ebuild uses configure options that aren't available in 7.0.1 - for example Tcl/Tk is no longer supported in favour of wxPython.

idella:
well this version of this ebuild is now surpassed. The question I believe is;
Do you want to have liblas/laszip ebuilds written by wraeth and added

jlec:
Take the additional ebuild and add support for it in grass. Give credit to @SoapZaNet though.
@wraeth for future reference, if something is not in the tree, you shouldn't treat the feature as non-existing, but try to add the missing deps.

jlec:
Fixed by and others
gentoo/gentoo@2fa849db86f415ee6eca0a7fb965c88606ace3e6
gentoo/gentoo@291b79be78175ee4087496bafbb015817d6294b6
gentoo/gentoo@08e80b8aebc0f36a2fd0c07e32bad61fbc6d29eb

PR #413, opened 2015-05-11
Fix the pythontex ebuild to actually install the latex files.

filmor:


filmor:
Is there any particular reason for this? Why are you not using the src_install and src_compile functions directly?

jlec:
Could you please rebase on latest master?

PR #425, opened 2015-05-21
sci-biology/salmon: add salmon

mschubert:
> Salmon is a wicked-fast program to produce a highly-accurate, transcript-level quantification estimates from RNA-seq data. Salmon achieves is accuracy and speed via a number of different innovations, including the use of lightweight alignments.

https://github.com/COMBINE-lab/salmon


sci-biology/salmon/salmon-0.3.2.ebuild
```diff
@@ -0,0 +1,23 @@
...
+HOMEPAGE="https://github.com/COMBINE-lab/salmon"
+SRC_URI="https://github.com/COMBINE-lab/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
+IUSE=""
+
+DEPEND=""
+RDEPEND="${DEPEND}"
```
jlec: If empty, both can be dropped.

sci-biology/salmon/salmon-0.3.2.ebuild
```diff
@@ -0,0 +1,23 @@
...
+
+src_prepare() {
+	epatch "${FILESDIR}/${P}-no-boost-static.patch"
```
jlec: Please use the PATCHES=().

marbre:
Please add a short description to the patch and clean it up like described in http://dev.gentoo.org/~vapier/clean-patches

mschubert:
Thank your for the hints, I'm learning something new with each PR :-)

Should be fixed now.

PR #429, opened 2015-05-28
sci-libs/minc

TheChymera:
Dependency for the possibly-upcoming [Freesurfer ebuild](https://github.com/gentoo-science/sci/issues/401).


sci-libs/minc/minc-2.2.00.ebuild
```diff
@@ -0,0 +1,27 @@
...
+DESCRIPTION="Medical Imaging file format and Toolbox"
+HOMEPAGE="http://packages.bic.mni.mcgill.ca/tgz/${P}.tar.gz"
+IUSE=""
+KEYWORDS="~amd64 ~x86"
+SLOT="0"
+LICENSE="BSD"
+
+DEPEND="sys-devel/flex
+	sys-devel/autoconf
+	sys-devel/gcc[cxx]
```
jlec: If you don't check in the ebuild that g++ used, you can drop it from depend. I can have gcc[cxx] installed but having CXX=icpc defined.

sci-libs/minc/minc-2.2.00.ebuild
```diff
@@ -0,0 +1,27 @@
...
+	sys-libs/zlib
+	dev-perl/Text-Format"
+
+RDEPEND="dev-util/byacc
```
jlec: Is this runtime needed?

sci-libs/minc/ChangeLog
```diff
@@ -0,0 +1,8 @@
+# ChangeLog for sci-libs/minc
+# Copyright 1999-2015 Gentoo Foundation; Distributed under the GPL v2
+# $Header: $
+
+*minc-2.2.00 (28 May 2015)
+
+  28 May 2015;  <chymera@gentoo.org> +metadata.xml, +minc-2.2.00.ebuild:
```
marbre: Seriously, you should fix your local setup first before messing up more and more ChangeLog entries.

TheChymera: I would gladly fix it if I (or anybody else) knew how.

sci-libs/minc/minc-2.2.00.ebuild
TheChymera: can I add a comment for this, so that if at any point it ends up not working I (or whoever else may want to look into it) knows that cxx might be an issue?

jlec: Just assume that there is a working eclass. For fortran we have even build a whole eclass for that.

sci-libs/minc/minc-2.2.00.ebuild
TheChymera: I do not know, I just know that upstream lists it as a dep. Should I add it to DEPEND instead?

TheChymera:
@jlec bump?

sci-libs/minc/minc-2.2.00.ebuild
```diff
@@ -0,0 +1,26 @@
...
+KEYWORDS="~amd64 ~x86"
+SLOT="0"
+LICENSE="BSD"
+
+DEPEND="sys-devel/flex
+	sys-devel/autoconf
+	sys-libs/zlib
+	dev-perl/Text-Format"
+
+RDEPEND="dev-util/byacc
```
jlec: Still I doubt that this is a runtime dependency. https://en.wikipedia.org/wiki/Yacc

TheChymera:
@jlec, yes, probably not needed at runtime. Good to go now?

sci-libs/minc/minc-2.2.00.ebuild
```diff
@@ -0,0 +1,26 @@
...
+KEYWORDS="~amd64 ~x86"
+SLOT="0"
+LICENSE="BSD"
+
+DEPEND="sys-devel/flex
+	sys-devel/autoconf
+	sys-libs/zlib
+	dev-perl/Text-Format
+	dev-util/byacc"
+
```
jlec: Are you sure the RDEPENDs aren't needed at buildtime too?

TheChymera: Not sure, but I would assume no - what should I do in this case?

TheChymera:
@jlec I tried to do a few tests without some of the packages - I think this is right now. Anything else?

sci-libs/minc/minc-2.2.00.ebuild
```diff
@@ -0,0 +1,26 @@
...
+LICENSE="BSD"
+
+DEPEND="sys-devel/flex
+	sys-devel/autoconf
+	sci-libs/netcdf
+	sys-libs/zlib
+	dev-perl/Text-Format
+	dev-util/byacc
+	x11-libs/libXmu
+	x11-libs/libXi"
```
jlec: Those libs are most probably also runtime deps.

TheChymera:
@jlec better now?

sci-libs/minc/minc-2.2.00.ebuild
```diff
@@ -0,0 +1,29 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Header: $
+
+EAPI=5
+
+inherit cmake-utils
+
+DESCRIPTION="Medical Imaging file format and Toolbox"
+HOMEPAGE="http://packages.bic.mni.mcgill.ca/tgz/${P}.tar.gz"
```
jlec: ?? That looks like to SRC_URI.

sci-libs/minc/minc-2.2.00.ebuild
```diff
@@ -0,0 +1,29 @@
...
+SLOT="0"
+LICENSE="BSD"
+
+COMMON_DEP="x11-libs/libXmu
+	x11-libs/libXi"
+
+DEPEND="sys-devel/flex
+	sys-devel/autoconf
+	sci-libs/netcdf
+	sys-libs/zlib
```
jlec: These are most probably also runtime deps.

sci-libs/minc/minc-2.2.00.ebuild
```diff
@@ -0,0 +1,29 @@
...
+IUSE=""
+KEYWORDS="~amd64 ~x86"
+SLOT="0"
+LICENSE="BSD"
+
+COMMON_DEP="x11-libs/libXmu
+	x11-libs/libXi"
+
+DEPEND="sys-devel/flex
+	sys-devel/autoconf
```
jlec: looks like a cmake build system. So I doubt you need autoconf.

TheChymera: it is listed as a dep [here](https://en.wikibooks.org/wiki/MINC/Tutorials/SourceInstall).

jlec: You should either of the two buildsystems. You cannot use both.

PR #430, opened 2015-05-30
dev-python/pyfftw: Implement python_test()

marbre:
I tried to implement the python_test() phase for pyfftw, unfortunately all tests fail with

`ImportError: No module named 'pyfftw.pyfftw'`

I tested
```bash
cd "${TEST_DIR}" || die
```
and
```bash
cd "${TEST_DIR}"/lib || die
```
as described in the [wiki](https://wiki.gentoo.org/wiki/Project:Python/distutils-r1#distutils_install_for_testing), but had no luck. Maybe somebody sees my mistake and can help.

In this ebuild, the package is named `pyfftw` while upstream names it `pyFFTW`. Anyway, renaming the ebuilds didn't helped. Thus, this shouldn't be a naming problem.

### DIFF
--- a/dev-python/pyfftw/ChangeLog
+++ b/dev-python/pyfftw/ChangeLog
@@ -1,6 +1,9 @@
 # Copyright 1999-2015 Gentoo Foundation; Distributed under the GPL v2
 # $Header: $
 
+  30 May 2015; Marius Brehler <marbre@linux.sungazer.de> pyfftw-9999.ebuild:
+  Implement python_test()
+
 *pyfftw-9999 (04 Feb 2015)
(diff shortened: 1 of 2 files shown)


marbre:
Copying the `setup.py` and the `test` directory works in general. On my machine the test suite fails
```
Ran 903 tests in 314.593s
FAILED (failures=2, errors=3, skipped=6)
```
[upstream](https://travis-ci.org/hgomersall/pyFFTW), it doesn't. Should I merge anyway @jlec? Furthermore, should the copied data be deleted in the end of the `python_test()` implementation after `esetup.py test` finished?

jlec:
merge it. Upstream succeeds because they are using build_ext --inplace

PR #432, opened 2015-06-06
Nibabel - using tarball instead of zip download

TheChymera:


sci-libs/nibabel/nibabel-1.3.0.ebuild
```diff
@@ -10,7 +10,7 @@ inherit distutils-r1
 
 DESCRIPTION="Access a cacophony of neuro-imaging file formats"
 HOMEPAGE="http://nipy.org/nibabel/"
-SRC_URI="https://github.com/nipy/nibabel/archive/${PV}.zip"
+SRC_URI="https://github.com/nipy/nibabel/archive/${PV}.tar.gz"
```
jlec: better use 

SRC_URI="https://github.com/nipy/nibabel/archive/${PV}.tar.gz -> ${P}.tar.gz"

TheChymera: ok, tried to do that, and (perhaps in an unrelated matter) repoman stopped working:
```
chymera@desktophost ~/src/neurogentoo/sci-libs/nibabel $ repoman commit

*** the local copy of metadata.dtd needs to be refetched, doing that now

  File "/usr/lib/python-exec/python2.7/repoman", line 1175, in <module>
    if not fetch_metadata_dtd():
  File "/usr/lib/python-exec/python2.7/repoman", line 1141, in fetch_metadata_dtd
    prefix='metadata.dtd.', dir=destdir)
  File "/usr/lib64/python2.7/tempfile.py", line 308, in mkstemp
    return _mkstemp_inner(dir, prefix, suffix, flags)
  File "/usr/lib64/python2.7/tempfile.py", line 239, in _mkstemp_inner
    fd = _os.open(file, flags, 0600)
OSError: [Errno 2] No such file or directory: '/usr/portage/distfiles/metadata.dtd.0Nyx6P'
```

any ideas what could have happened? I guess I could change the permissions for /usr/portage/distfiles - but that does not seem right.

jlec: Either add your user to portage group or set DISTDIR to something you can write to.

TheChymera: my user is already in the portage group :-/

PR #433, opened 2015-06-09
media-libs/avbin-bin build fixed -- package S (source directory) depends on architecture

flagist0:


media-libs/avbin-bin/avbin-bin-7.ebuild
```diff
@@ -15,13 +15,13 @@ SLOT="0"
...
-	if use amd64; then
-		cd "${WORKDIR}"/avbin-linux-x86-64-${PV} || die
-	elif use x86; then
-		cd "${WORKDIR}"/avbin-linux-x86-32-${PV} || die
-	fi
+if use amd64; then
+	S="${WORKDIR}"/avbin-linux-x86-64-${PV}
+elif use x86; then
+	S="${WORKDIR}"/avbin-linux-x86-32-${PV}
+fi
```
jlec: That won't work as S is cached and the cache could also be generated on an hppa machine, which would result in S=. It needs to wrapped in some phases e.g. pkg_prepare()

jauhien:
@flagist0: please, merge these commits into one and update this PR (git push -f will work for this).

PR #442, opened 2015-06-18
sci-libs/minc-toolkit new alternative MINC distribution

TheChymera:
Package-Manager: portage-2.2.20

Perhaps better than https://github.com/gentoo-science/sci/pull/429

### DIFF
--- a/sci-libs/minc-toolkit/metadata.xml
+++ b/sci-libs/minc-toolkit/metadata.xml
@@ -7,4 +7,7 @@
   MINC-based packages that historically have been developed somewhat 
   independently.
 </longdescription>
+<use>
+  <flag name="itk">Emerge sci-libs/itk to complement the package</flag>
+</use>
 </pkgmetadata>
--- a/sci-libs/minc-toolkit/metadata.xml
+++ b/sci-libs/minc-toolkit/metadata.xml
@@ -8,6 +8,6 @@
   independently.
 </longdescription>
 <use>
-  <flag name="itk">Emerge sci-libs/itk to complement the package</flag>
+  <flag name="itk">Build ITK-based tools</flag>
 </use>
 </pkgmetadata>
--- a/sci-libs/minc-toolkit/metadata.xml
+++ b/sci-libs/minc-toolkit/metadata.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+  <herd>sci</herd>
+  <longdescription lang="en">
+  This metaproject bundles multiple MINC (Medical Imaging NetCDF) and 
+  MINC-based packages that historically have been developed somewhat 
+  independently.
+</longdescription>
+</pkgmetadata>
--- a/sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
+++ b/sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
@@ -1,6 +1,6 @@
-#DUS Copyright 1999-2015 Gentoo Foundation
+# Copyright 1999-2015 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
-# $Header: $
+# $Id$
 
 EAPI="5"
 
--- a/sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
+++ b/sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
@@ -28,3 +28,8 @@ DEPEND="dev-lang/perl
 RDEPEND="sci-libs/hdf5
 	sci-libs/netcdf
 	${COMMON_DEP}"
+
+src_configure() {
+	local mycmakeargs=( -DUSE_SYSTEM_HDF5=1 )
+	cmake-utils_src_configure
+}
--- a/sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
+++ b/sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
@@ -17,7 +17,7 @@ IUSE="itk"
 KEYWORDS=""
 
 COMMON_DEP="dev-libs/libpcre
-	sci-libs/fftw
+	sci-libs/fftw:3.0
 	sci-libs/gsl
 	sci-libs/hdf5
 	itk? ( sci-libs/itk )
--- a/sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
+++ b/sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
@@ -1,0 +1,17 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Header: $
+
+EAPI="5"
+
+inherit cmake-utils git-r3
+
+DESCRIPTION="Metaproject uniting all the available MINC tools"
+HOMEPAGE="https://github.com/BIC-MNI/minc-toolkit"
+SRC_URI=""
+EGIT_REPO_URI="git://github.com/BIC-MNI/minc-toolkit.git"
+
+LICENSE="BSD"
+SLOT="0"
+IUSE=""
+KEYWORDS=""
--- a/sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
+++ b/sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
@@ -16,7 +16,10 @@ SLOT="0"
 IUSE=""
 KEYWORDS=""
 
-COMMON_DEP="sci-libs/hdf5
+COMMON_DEP="dev-libs/libpcre
+	sci-libs/gsl
+	sci-libs/hdf5
+	sci-libs/itk
 	media-libs/freeglut
 	x11-libs/libXmu
 	x11-libs/libXi"
@@ -32,8 +35,11 @@ RDEPEND="sci-libs/netcdf
 
 src_configure() {
 	local mycmakeargs=(
+		-DUSE_SYSTEM_GSL=1
 		-DUSE_SYSTEM_HDF5=1
+		-DUSE_SYSTEM_ITK=1
 		-DUSE_SYSTEM_NETCDF=1
+		-DUSE_SYSTEM_PCRE=1
 		-DUSE_SYSTEM_ZLIB=1
 		)
 	cmake-utils_src_configure
--- a/sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
+++ b/sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
@@ -15,3 +15,16 @@ LICENSE="BSD"
 SLOT="0"
 IUSE=""
 KEYWORDS=""
+
+COMMON_DEP="x11-libs/libXmu
+	x11-libs/libXi"
+
+DEPEND="dev-lang/perl
+	media-libs/freeglut
+	sys-devel/bison
+	sys-devel/flex
+	${COMMON_DEP}"
+
+RDEPEND="sci-libs/hdf5
+	media-libs/netcdf
+	${COMMON_DEP}"
--- a/sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
+++ b/sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
@@ -43,6 +43,7 @@ src_configure() {
 		-DUSE_SYSTEM_NETCDF=1
 		-DUSE_SYSTEM_PCRE=1
 		-DUSE_SYSTEM_ZLIB=1
+		-DMT_BUILD_ITK_TOOL=$(usex itk 1 0)
 		)
 	cmake-utils_src_configure
 }
--- a/sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
+++ b/sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
@@ -16,11 +16,11 @@ SLOT="0"
 IUSE=""
 KEYWORDS=""
 
-COMMON_DEP="x11-libs/libXmu
+COMMON_DEP="media-libs/freeglut
+	x11-libs/libXmu
 	x11-libs/libXi"
 
 DEPEND="dev-lang/perl
-	media-libs/freeglut
 	sys-devel/bison
 	sys-devel/flex
 	${COMMON_DEP}"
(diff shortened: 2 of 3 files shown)


sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
```diff
@@ -0,0 +1,30 @@
...
+LICENSE="BSD"
+SLOT="0"
+IUSE=""
+KEYWORDS=""
+
+COMMON_DEP="x11-libs/libXmu
+	x11-libs/libXi"
+
+DEPEND="dev-lang/perl
+	media-libs/freeglut
```
jlec: This is RDEPEND too.

sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
```diff
@@ -0,0 +1,30 @@
...
+	sys-devel/bison
+	sys-devel/flex
+	${COMMON_DEP}"
+
+RDEPEND="sci-libs/hdf5
+	sci-libs/netcdf
```
jlec: these two are most probably also build time deps.

TheChymera: Upstream [omits](https://github.com/BIC-MNI/minc-toolkit#dependencies) them from the build deps.

marbre: No, that is incorrect. The cmake project is able to fetch netcdf and hdt5 and build it as part of minc. That is however not desired and system libs should be used instead.

See https://github.com/BIC-MNI/minc-toolkit/blob/master/CMakeLists.txt#L70 and follow `USE_SYSTEM_NETCDF`.

sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
```diff
@@ -0,0 +1,35 @@
...
+	${COMMON_DEP}"
+
+src_configure() {
+	local mycmakeargs=( -DUSE_SYSTEM_HDF5=1 )
+	cmake-utils_src_configure
```
marbre: `USE_SYSTEM_ZLIB`, `USE_SYSTEM_NETCDF`, etc. missing. Some of the options might be optional and should be activated via use flags. If so, use conditional dependencies are required.

sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
TheChymera: @marbre 
Ok, I am proceding to set as many as I can to 1 (ON). Is what @jlec said true, namely that the corresponding system atoms for these programs would then be required at build time? If so, is that all of them, or which exactly?

marbre: I just took a quick look. Some definitely seem to be required build time dependencies. Some might be optional. It is your pull request, so please check this, I can't tell you without digging deeper into the CMakeLists.

sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
TheChymera: working on that now. I can't tell which the packages for PCRE, GSL, FFTW3F, and ITK are. There are a few packages with those names, or in the case of FFTW3F, none. Can you help me out?

marbre: Seriously, you shouldn't use software for scientific purposes you don't know. fftw3f is the floating point library, while fftw3 is the double precision lib.
Further, just check the cmake modules in https://github.com/BIC-MNI/minc-toolkit/tree/master/cmake-modules to see from which destination the sources are fetched. This will point you to the right packages if compared with the output of `eix $packagename`.

TheChymera: @marbre really? it seems fftw is a fourier transform thing: http://www.fftw.org/ - so why do they cal it fftw3f and not fftw then?

marbre: I use it for several years now, so I am sure. It is version 3 of the FFTW, therefore the 3. And the **f** is the suffix for float. Without the suffix it's the double implementation as already mentioned.

sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
```diff
@@ -0,0 +1,46 @@
...
+
+LICENSE="BSD"
+SLOT="0"
+IUSE=""
+KEYWORDS=""
+
+COMMON_DEP="dev-libs/libpcre
+	sci-libs/gsl
+	sci-libs/hdf5
+	sci-libs/itk
```
marbre: This is optional, as itk is only required for `MT_BUILD_ITK_TOOLS`, see https://github.com/BIC-MNI/minc-toolkit/blob/master/CMakeLists.txt#L308
Thus, this is an use conditional dependency. Check https://devmanual.gentoo.org/general-concepts/dependencies/.

sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
```diff
@@ -0,0 +1,46 @@
...
+	media-libs/freeglut
```
marbre: Only required for `MT_BUILD_VISUAL_TOOLS`, thus use conditional dep. See https://github.com/BIC-MNI/minc-toolkit/blob/master/CMakeLists.txt#L376 and please be more accurate in case of dependencies.

TheChymera:
Ok, I think the ebuild looks really nice now - buuuuuut, it won't build - see log [here](https://bpaste.net/show/ca25171777b7).

Any ideas?

kiwifb:
It builds. it just won't install. There must be something wrong in the rule for installation.

TheChymera:
How can I find out what?

kiwifb:
```
CMake Error at cmake_install.cmake:36 (file):
  file INSTALL cannot find
  "/var/tmp/portage/sci-libs/minc-toolkit-9999/work/minc-toolkit-9999_build/external//usr/bin".
```
You have to find what exactly `cmake` is trying to do there. It looks like it either looking at installing files relative to the building directory or finding them in that location for install. It is a bit difficult to infere without looking at `CMakeList.txt` like I am doing right now.

sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
```diff
@@ -0,0 +1,47 @@
...
+src_configure() {
+	local mycmakeargs=(
+		-DUSE_SYSTEM_FFTW3F=1
+		-DUSE_SYSTEM_GSL=1
+		-DUSE_SYSTEM_HDF5=1
+		-DUSE_SYSTEM_NETCDF=1
+		-DUSE_SYSTEM_PCRE=1
+		-DUSE_SYSTEM_ZLIB=1
+		)
+	cmake-utils_src_configure
```
marbre: No cmake arguments set if `itk` use flag is set.

sci-libs/minc-toolkit/metadata.xml
```diff
@@ -0,0 +1,13 @@
...
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+  <herd>sci</herd>
+  <longdescription lang="en">
+  This metaproject bundles multiple MINC (Medical Imaging NetCDF) and 
+  MINC-based packages that historically have been developed somewhat 
+  independently.
+</longdescription>
+<use>
+  <flag name="itk">Emerge sci-libs/itk to complement the package</flag>
```
marbre: Description is not very helpful. Please improve it based on the description in the CMakeLists.txt.

sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
```diff
@@ -0,0 +1,47 @@
...
+LICENSE="BSD"
+SLOT="0"
+IUSE="itk"
+KEYWORDS=""
+
+COMMON_DEP="dev-libs/libpcre
+	sci-libs/fftw
+	sci-libs/gsl
+	sci-libs/hdf5
+	itk? ( sci-libs/itk )
```
jlec: No USE dependent dependencies if there are no buildtime changes. Use optfeatures from eutils.eclass for this,

marbre: Yes, `MT_BUILD_ITK_TOOL=On` needs to be set.

sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
TheChymera: none is needed, we install it via portage.

sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
TheChymera: @jlec - so practically, I should just remove the use flag and add a message like 

`optfeature "foo support" app-misc/foo`

In what part of the ebuild do I put that?

marbre: **cmake arguments, as I have already mentioned!** See your own comment, where you said you don't need itk.

sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
marbre: Put `MT_BUILD_ITK_TOOL` depending on itk use flag...

TheChymera: @marbre and I should also add -DUSE_SYSTEM_ITK=1, right?

marbre: Yes and no, add a use dependent cmake argument to enable USE_SYSTEM_ITK and MT_BUILD_ITK_TOOL if the itk use flag is set.

TheChymera: @marbre how do I conditionally turn this `MT_BUILD_ITK_TOOLS` on?

I tried `itk? ( DMT_BUILD_ITK_TOOLS=1 )` but it doesn't work.

TheChymera: @marbre ?

jlec: Please check the manpage of cmake-utils.eclass and the cmake documentation.

TheChymera: @jlec you mean this? It contains no info of that sort - all the `cmake-utils_use_*` functions have fixed variable names, and `MT*` is not pat of any of them.

marbre: https://devmanual.gentoo.org/tools-reference/bash/

jlec: Yes I meant that. But the easiest way to implement something is usex().

TheChymera: @jlec thank you. does this look good now?

sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
```diff
@@ -0,0 +1,49 @@
...
+src_configure() {
+	local mycmakeargs=(
+		cmake-utils_use_use itk SYSTEM_ITK
+		-DUSE_SYSTEM_FFTW3F=1
+		-DUSE_SYSTEM_GSL=1
+		-DUSE_SYSTEM_HDF5=1
+		-DUSE_SYSTEM_NETCDF=1
+		-DUSE_SYSTEM_PCRE=1
+		-DUSE_SYSTEM_ZLIB=1
+		-DMT_BUILD_ITK_TOOL=$(usex itk 1 0)
```
jlec: That look good, although I am unsure weather ON & OFF is the better cmake style syntax.

TheChymera: any idea why travis is failing, I don't get the version thing - do I have to update my repoman?

TheChymera:
@jlec @marbre  hi again, do you guys know how come the travis build fails? this should be good to go, right?

jauhien:
@TheChymera: travis fails not on your changes.

TheChymera:
soooo, can I haz merge nao?

sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
marbre: What about jlec's comment regarding ON & OFF, before merging your changes?

sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
```diff
@@ -0,0 +1,49 @@
...
+	sys-devel/bison
+	sys-devel/flex
+	${COMMON_DEP}"
+
+RDEPEND="sci-libs/netcdf
+	${COMMON_DEP}"
+
+src_configure() {
+	local mycmakeargs=(
+		cmake-utils_use_use itk SYSTEM_ITK
```
jlec: This doesn't work. You need to wrap it into $()

TheChymera:
@jlec - seems good, right?

sci-libs/minc-toolkit/minc-toolkit-9999.ebuild
```diff
@@ -0,0 +1,49 @@
...
+SRC_URI=""
+EGIT_REPO_URI="git://github.com/BIC-MNI/minc-toolkit.git"
+
+LICENSE="BSD"
+SLOT="0"
+IUSE="itk"
+KEYWORDS=""
+
+COMMON_DEP="dev-libs/libpcre
+	sci-libs/fftw
```
jlec: repoman should have told you, that you need to set a slot here.

TheChymera:
@jlec bump?

TheChymera:
this apparently still fails on my machine with:


```
Install the project...
-- Install configuration: "Gentoo"
CMake Error at cmake_install.cmake:36 (file):
  file INSTALL cannot find
  "/var/tmp/portage/sci-libs/minc-toolkit-9999/work/minc-toolkit-9999_build/external//usr/bin".
```

Any idea why? why would it be looking for that path?

TheChymera:
@kiwifb hey, sorry if you're busy, but can you maybe help me out with this? I checked and `/var/tmp/portage/sci-libs/minc-toolkit-9999/work/minc-toolkit-9999_build/external/` does indeed not exist, but I have no idea whether it should actually exist, or how I can make portage not check for it if it does indeed not...

PR #449, opened 2015-07-10
sci-misc/kaldi: version bump

akreal:


jauhien:
What is the reason for this ebuild to be not the usual live ebuild that tracks the trunk, but point to some concrete revision instead? In this case why not to create a snapshot? I can host it in my dev space e.g.

akreal:
Upstream doesn't provide any releases, I want to produce some sort of it. Snapshots would do, too. I can host them on personal VPS, are there any particular requirements for a location referred in ebuild?

jauhien:
For tree I would ask you to host snapshot on the gentoo infrastructure, but for a package in overlay I guess your VPS would be ok, or you can deside to host snapshot in my devspace (http://dev.gentoo.org/~jauhien/distfiles/FILENAME, see sys-cluster/ampi for an example). If you'll choose second option, please, tell me what svn tag (or how do they call it) you've used and update this PR with corrected ebuild, I'll do a snapshot. To update PR you can ammend your commit and do `git push -f` to your branch.

akreal:
Done.

jauhien:
Thanks, I will test and then merge it in the nearest time.

akreal:
Thank you for review and comments!

PR #451, opened 2015-07-21
Updated pagmo URIs.

mirandaconrado:
The PaGMO library moved from sourceforge to GitHub, as described in http://sourceforge.net/projects/pagmo/.

### DIFF
--- a/sci-libs/pagmo/ChangeLog
+++ b/sci-libs/pagmo/ChangeLog
@@ -1,6 +1,9 @@
 # ChangeLog for sci-libs/pagmo
-# Copyright 1999-2014 Gentoo Foundation; Distributed under the GPL v2
+# Copyright 1999-2015 Gentoo Foundation; Distributed under the GPL v2
 # $Header: $
+
+  22 Jul 2015; Conrado Miranda <miranda.conrado@gmail.org> pagmo-9999.ebuild:
+  sci-libs/pagmo: Updated URIs.
 
   06 Jan 2014; Justin Lecher <jlec@gentoo.org> pagmo-9999.ebuild:
   Add missing python deps
--- a/sci-libs/pagmo/pagmo-9999.ebuild
+++ b/sci-libs/pagmo/pagmo-9999.ebuild
@@ -9,9 +9,9 @@ PYTHON_COMPAT=( python2_7 )
 inherit cmake-utils git-r3 python-single-r1
 
 DESCRIPTION="Parallelization engine for optimization problems"
-HOMEPAGE="http://pagmo.sourceforge.net/"
+HOMEPAGE="https://github.com/esa/pagmo"
 SRC_URI=""
-EGIT_REPO_URI="git://pagmo.git.sourceforge.net/gitroot/pagmo/pagmo"
+EGIT_REPO_URI="https://github.com/esa/pagmo.git"
 
 LICENSE="GPL-3"
 SLOT="0"
--- a/sci-libs/pagmo/pagmo-9999.ebuild
+++ b/sci-libs/pagmo/pagmo-9999.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2014 Gentoo Foundation
+# Copyright 1999-2015 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 # $Header: $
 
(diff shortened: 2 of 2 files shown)


sci-libs/pagmo/pagmo-9999.ebuild
```diff
@@ -9,9 +9,9 @@ PYTHON_COMPAT=( python2_7 )
 inherit cmake-utils git-r3 python-single-r1
 
 DESCRIPTION="Parallelization engine for optimization problems"
-HOMEPAGE="http://pagmo.sourceforge.net/"
+HOMEPAGE="https://github.com/esa/pagmo"
 SRC_URI=""
-EGIT_REPO_URI="git://pagmo.git.sourceforge.net/gitroot/pagmo/pagmo"
+EGIT_REPO_URI="https://github.com/esa/pagmo.git"
```
marbre: Please also add the `git://` URL, like in [pylibconfig](https://github.com/gentoo-science/sci/blob/master/dev-python/pylibconfig2/pylibconfig2-9999.ebuild#L12).

marbre:
Please run `repoman full` to check for any issues with the ebuild (also your modifications are minor). This should point you to the issue, that the copyright years has to be adjusted.
Further, use `repo-commit` for future commits. This also adds a ChangeLog entry, which is missing and should be added.

Nevertheless, thanks for your your pull request, we're happy to merge :)

mirandaconrado:
Sorry for not following the protocol before! I used repoman and repo-commit and followed the pylibconfig2 example now, so everything should be in order. Don't hesitate to call me on another mistake =)

sci-libs/pagmo/ChangeLog
```diff
@@ -1,7 +1,10 @@
 # ChangeLog for sci-libs/pagmo
-# Copyright 1999-2014 Gentoo Foundation; Distributed under the GPL v2
+# Copyright 1999-2015 Gentoo Foundation; Distributed under the GPL v2
 # $Header: $
 
+  22 Jul 2015;  <conrado@gentoo.org> pagmo-9999.ebuild:
```
marbre: One minor issue. I think you forgot to set `ECHANGELOG_USER`. See [here](https://github.com/gentoo-science/sci/blob/master/CONTRIBUTING.md#recommended). Anyway, It's fine to correct the ChangeLog entry manually and to push without re-commit if it's only a ChangeLog fix.

mirandaconrado:
I'm so ashamed :@ Thank you for the patience!

marbre:
You're welcome and we're happy for support. Thanks for your changes, merged it.

sci-libs/pagmo/pagmo-9999.ebuild
```diff
@@ -9,9 +9,9 @@ PYTHON_COMPAT=( python2_7 )
 inherit cmake-utils git-r3 python-single-r1
 
 DESCRIPTION="Parallelization engine for optimization problems"
-HOMEPAGE="http://pagmo.sourceforge.net/"
+HOMEPAGE="https://github.com/esa/pagmo"
 SRC_URI=""
-EGIT_REPO_URI="git://pagmo.git.sourceforge.net/gitroot/pagmo/pagmo"
+EGIT_REPO_URI="https://github.com/esa/${PN}.git git://git@github.com:esa/${PN}.git"
```
jlec: The second URI can be dropped.

marbre: @jlec This was added, because I suggested so. AFAIR @junghans or somebody else gave the advice during one of my early pull request.

jlec: the git@github.com urls are for ssh access. you portage user shouldn't be able to use your ssh key. So it can be dropped. https is reliable enough.

mirandaconrado: I agree with @jlec, since ssh is only required for remote editing, but @marbre provided a [precedent](https://github.com/gentoo-science/sci/blob/master/dev-python/pylibconfig2/pylibconfig2-9999.ebuild#L12).

marbre: Just checked https://github.com/gentoo-science/sci/pull/327. In the 2nd comment @junghans suggested to add git:// too. Can't remember why.

Further, you don't need an ssh key to pull from github via ssh, depending on the repository settings. For example `git clone git://github.com/DamnWidget/sublime-text.git` works without problems (layman -a added this to my */etc/portage/repos.conf/layman.conf*).

But sure, we can drop it. I would than grep over the whole overlay and fix that.

marbre: Okay, fixed that with commit https://github.com/gentoo-science/sci/commit/0e4c2d82cb4b6bfdc17daa4ec56246444e19c291. Actually `git://github.com/` is okay, but `git://git@github.com:` isn't. It was wrong in three of my ebuilds and unfortunately I referenced one of the broken ons. Sorry for the that, I probably should have noticed that during the pull request.

mirandaconrado: Interesting that omitting the git user works. Anyway, no harm done.

Issue #453, opened 2015-07-21
eselect alternative is totally broken on prefix


kiwifb:
In the past I have fixed eselect to work on prefix. The fix doesn't apply anymore and trying to port it as is doesn't work either because of changes in alternatives-2.eclass.

Typical problem with the current set up on prefix. links and files normally installed under `/etc/env.d/alternatives` are not installed in the right place. They are installed in so called double prefix. So you find them under `$EPREFIX/$EPREFIX/etc/env.d/alternatives` that's one `$EPREFIX` too many.

Root cause are found in `usr/share/eselect/libs/alternatives-common.bash`, namely
```
inherit config output path-manipulation

: "${ALTERNATIVESDIR_ROOTLESS:=/etc/env.d/alternatives}"
ALTERNATIVESDIR="${EROOT%/}${ALTERNATIVESDIR_ROOTLESS}"
```
In a prefix because `ALTERNATIVESDIR_ROOTLESS` is populated with ` @sysconfdir@/env.d/alternatives` the end result will include the prefix as well. So in a prefix you will have
```
inherit config output path-manipulation

: "${ALTERNATIVESDIR_ROOTLESS:=${EPREFIX}/etc/env.d/alternatives}"
ALTERNATIVESDIR="${EROOT%/}${ALTERNATIVESDIR_ROOTLESS}"
```
where `$EPREFIX` is expanded to its value. `$EROOT` is equivalent to `$ROOT/$EPREFIX` so `ALTERNATIVESDIR` ends up with a double prefix. Bad.

The old solution that I created before `ALTERNATIVESDIR*` where put in `alternatives-common.bash` and was still in `alternatives.bash` was to remove the extra prefix from `ALTERNATIVESDIR_ROOTLESS` with `sed` during `src_install`.

This is unfortunately rendered moot by the following in `alternatives-2.eclass
```
alternatives_for() {
	debug-print-function ${FUNCNAME} "${@}"

	dodir /etc/env.d/alternatives

	ALTERNATIVESDIR_ROOTLESS="${ED}/etc/env.d/alternatives" \
		eselect alternatives add ${@} || die

	ALTERNATIVES_CREATED+=( ${1} )
}
```
Again we have `$ED` which is `$D/$EPREFIX` again `$D` is already including the prefix so affixing `$EROOT` will double it. And you end creating stuff in strange place during `src_install`.

The most easy fix is to define
```
[...]

PR #454, opened 2015-07-22
Update eslect support for prefix. The current set up was badly broken…

kiwifb:
… with multiple double prefix and other oddities.

Package-Manager: portage-2.2.20

This is to solve issue #453, the usage of `EROOT` instead of `ROOT` everywhere place. Introduced `ERROT` in the only place it was making a difference. I am open to the suggestion it should `EPREFIX` instead.

### DIFF
--- a/app-admin/eselect/ChangeLog
+++ b/app-admin/eselect/ChangeLog
@@ -2,6 +2,10 @@
 # Copyright 1999-2015 Gentoo Foundation; Distributed under the GPL v2
 # $Header: /var/cvsroot/gentoo-x86/app-admin/eselect/ChangeLog,v 1.179 2011/01/22 21:38:52 ulm Exp $
 
+  22 Jul 2015; François Bissey <francois.bissey@canterbury.ac.nz>
+  files/eselect-1.4.4-alternatives.patch:
+  Restore a couple of needed EROOT
+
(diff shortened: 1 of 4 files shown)


kiwifb:
There is a couple of `$EROOT` that needs to be restored on further examination. Do not merge this now.

jlec:
I messed up a couple of things with multilib build and int64 support in the last update. Will fix this hopefully this week.

kiwifb:
I noticed, but I believe the stuff is independent of the multilib build breaking, it is double prefix all over the place. But I guess only once the design of the eclass/eselect is fixed I will be able to say for sure that all the `EROOT` I replaced needed so. I know those to set up the initial alternatives needed to but I am not sure about those involved in link maintenance.

jlec:
Could you please rebase onto current master and also send a PR for https://github.com/gentoo-science/eselect ?

kiwifb:
I will, I knew I would have to rebase. I am giving you a time frame because I am in the middle of paper work to buy a house and it is anything but straightforward.

kiwifb:
Regenerating the branch.

heroxbd:
Hi François, could you help me test https://github.com/gentoo-science/sci/pull/541 ?  Thanks.

kiwifb:
I'll be quite busy the next 48h but I'll have a look after that.

PR #456, opened 2015-07-25
Fix Scilab and jlatexmath-fop to build with new versions of dependencies

khumba:
Scilab 5.5.2's Java code doesn't compile unmodified against batik-1.8, fop-2.0, and xmlgraphics-common-2.0, so I've written some patches to get it building again.  I'm not familiar with the Scilab internals or the libraries above, but the fixes seemed simple enough.  I'll take care of sending them upstream.

Also dev-java/jlatexmath-fop-1.0.3 specified an incorrect jar dep, preventing it from building.

### DIFF
--- a/sci-mathematics/scilab/files/scilab-5.5.2-batik-1.8.patch
+++ b/sci-mathematics/scilab/files/scilab-5.5.2-batik-1.8.patch
@@ -1,0 +1,98 @@
+From: Bryan Gardiner <bog@khumba.net>
+Project-Bug-URL: https://bugzilla.scilab.org/show_bug.cgi?id=13813
+Gentoo-Bug-URL: https://github.com/gentoo-science/sci/pull/456
+
+Updates Scilab to build against (and require) Batik >=1.8, fixing an API change
+with SAXSVGDocumentFactory being moved.  This is not tested beyond "yes Scilab
+still links to Batik to export SVGs," as I'm not sure of a good test case, and
+it's awaiting upstream review.
+
+--- a/scilab/configure
++++ b/scilab/configure
+@@ -16600,8 +16600,8 @@ $as_echo "$as_me: WARNING: Could not find or use the Java package/jar batik used
+                 fi
+ 
+ 
+-    { $as_echo "$as_me:${as_lineno-$LINENO}: checking minimal version (1.7) of batik" >&5
+-$as_echo_n "checking minimal version (1.7) of batik... " >&6; }
++    { $as_echo "$as_me:${as_lineno-$LINENO}: checking minimal version (1.8) of batik" >&5
++$as_echo_n "checking minimal version (1.8) of batik... " >&6; }
+    saved_ac_java_classpath=$ac_java_classpath
+    export ac_java_classpath="$BATIK:$ac_java_classpath"
+    if test "x" == "x"; then
+@@ -16614,7 +16614,7 @@ import org.apache.batik.Version;
+ 
+ public class conftest {
+     public static void main(String[] argv) {
+-        String minVersion="1.7";
++        String minVersion="1.8";
+ 
+                             System.out.println(Version.getVersion());
+                                   if (compare(minVersion, Version.getVersion()) > 0) {
+@@ -16663,7 +16663,7 @@ EOF
+                    STDOUT=`cat conftest.java.output`
+                 fi
+                 echo "configure: CLASSPATH was $CLASSPATH" >&5
+-                  as_fn_error $? "Wrong version of batik. Expected at least 1.7. Found $STDOUT" "$LINENO" 5
++                  as_fn_error $? "Wrong version of batik. Expected at least 1.8. Found $STDOUT" "$LINENO" 5
+                             fi
+         fi
+         if test -f conftest.java.output; then
+@@ -16673,7 +16673,7 @@ EOF
+         echo "configure: failed program was:" >&5
+         cat conftest.java >&5
+         echo "configure: CLASSPATH was $CLASSPATH" >&5
+-          as_fn_error $? "Wrong version of batik. Expected at least 1.7. Found $STDOUT" "$LINENO" 5
++          as_fn_error $? "Wrong version of batik. Expected at least 1.8. Found $STDOUT" "$LINENO" 5
+             fi
+ 
+    else
+@@ -16686,7 +16686,7 @@ import org.apache.batik.Version;
+ 
+ public class conftest {
+     public static void main(String[] argv) {
+-        String minVersion="1.7";
++        String minVersion="1.8";
+ 
+                             System.out.println(Version.getVersion());
+                                   if (compare(minVersion, Version.getVersion()) != 0) {
+@@ -16735,7 +16735,7 @@ EOF
+                    STDOUT=`cat conftest.java.output`
+                 fi
+                 echo "configure: CLASSPATH was $CLASSPATH" >&5
+-                  as_fn_error $? "Wrong version of batik. Expected exact version 1.7. Found $STDOUT" "$LINENO" 5
++                  as_fn_error $? "Wrong version of batik. Expected exact version 1.8. Found $STDOUT" "$LINENO" 5
+                             fi
+         fi
+         if test -f conftest.java.output; then
+@@ -16745,7 +16745,7 @@ EOF
+         echo "configure: failed program was:" >&5
+         cat conftest.java >&5
+         echo "configure: CLASSPATH was $CLASSPATH" >&5
+-          as_fn_error $? "Wrong version of batik. Expected exact version 1.7. Found $STDOUT" "$LINENO" 5
++          as_fn_error $? "Wrong version of batik. Expected exact version 1.8. Found $STDOUT" "$LINENO" 5
+             fi
+ 
+     fi
+--- a/scilab/configure.ac
++++ b/scilab/configure.ac
+@@ -1038,7 +1038,7 @@ interface for JOGL2 - or libGL (OpenGL library) are installed and if the version
+                    BATIK=$PACKAGE_JAR_FILE
+                 fi
+                 AC_SUBST(BATIK)
+-                AC_JAVA_CHECK_VERSION_PACKAGE([batik],[import org.apache.batik.Version;],$BATIK,[1.7],[Version.getVersion()])
++                AC_JAVA_CHECK_VERSION_PACKAGE([batik],[import org.apache.batik.Version;],$BATIK,[1.8],[Version.getVersion()])
+ 
+                 # Commons I/O library
+                 AC_JAVA_CHECK_PACKAGE([commons-io],[org.apache.commons.io.output.CountingOutputStream],[Commons I/O library])
+--- a/scilab/modules/graph/src/java/org/scilab/modules/graph/utils/ScilabGraphUtils.java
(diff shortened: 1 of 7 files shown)


khumba:
Upstream bugs:
https://bugzilla.scilab.org/show_bug.cgi?id=13724 (xmlgraphics-commons)
https://bugzilla.scilab.org/show_bug.cgi?id=13813 (batik)
https://bugzilla.scilab.org/show_bug.cgi?id=14009 (fop)

marbre:
Hello Bryan, thanks a lot for your work. I recently fixed the jlatexmath-fop and scilab ebuilds blindly to get travis back to work.

For the next time, please try to split into multiple merge requests, one per package. This makes the reviewing easier for us.

dev-java/batik/batik-1.8-r2.ebuild
```diff
@@ -0,0 +1,110 @@
...
+	local imports=()
+	imports+=(sources/org/apache/batik/script/jpython/JPythonInterpreterFactory.java)
+	imports+=(sources/org/apache/batik/script/jacl/JaclInterpreterFactory.java)
+	for import in ${imports[@]}; do
+		einfo "Fixing missing import in ${import}"
+		sed -i '23i import org.apache.batik.script.ImportInfo;' ${import}
+		eend $?
+	done
+
+	cd lib
```
marbre: Missing die.

sci-mathematics/scilab/files/scilab-5.5.2-batik-1.8.patch
```diff
@@ -0,0 +1,134 @@
+diff --git a/scilab/configure b/scilab/configure
+index fd0e339..a4451ce 100755
+--- a/scilab/configure
++++ b/scilab/configure
```
marbre: Please add a little header like described [here](http://dev.gentoo.org/~vapier/clean-patches).

dev-java/batik/batik-1.8-r2.ebuild
```diff
@@ -0,0 +1,110 @@
...
+
+	# Add some missing imports to avoid a compiling issue.
+	# https://bugs.gentoo.org/show_bug.cgi?id=551952
+	# https://issues.apache.org/jira/browse/BATIK-1123
+	local imports=()
+	imports+=(sources/org/apache/batik/script/jpython/JPythonInterpreterFactory.java)
+	imports+=(sources/org/apache/batik/script/jacl/JaclInterpreterFactory.java)
+	for import in ${imports[@]}; do
+		einfo "Fixing missing import in ${import}"
+		sed -i '23i import org.apache.batik.script.ImportInfo;' ${import}
```
marbre: Missing die.

dev-java/batik/batik-1.8-r2.ebuild
khumba: Thanks for the careful review of the ebuild.  I left a note for -r1::gentoo here:
https://bugs.gentoo.org/show_bug.cgi?id=555870

sci-mathematics/scilab/files/scilab-5.5.2-batik-1.8.patch
khumba: Done.  Thanks for the fast review and for getting the ball rolling previously, Marius.

dev-java/batik/batik-1.8-r2.ebuild
khumba: Done.

khumba:
Done, thanks, and yes I'll split it up next time :).

sci-mathematics/scilab/files/scilab-5.5.2-xmlgraphics-common-2.0.patch
```diff
@@ -0,0 +1,125 @@
...
+awaiting upstream review.
+
+--- a/scilab/etc/classpath.xml.vc
++++ b/scilab/etc/classpath.xml.vc
+@@ -120,7 +120,7 @@ The option is "disableUnderMode" and can be:
+     </path>
+ 
+     <!-- PDF doc generation (dep of jeuclid) -->
+-    <path value="$SCILAB/thirdparty/xmlgraphics-commons-1.4.jar" load="onUse">
++    <path value="$SCILAB/thirdparty/xmlgraphics-commons-2.0.1.jar" load="onUse">
```
marbre: Since the patch seems only to work for xmlgraphics-commons-2.0.1, please replace the dependency `dev-java/xmlgraphics-commons:2` with `dev-java/xmlgraphics-commons-2.0.1:2` in the scilab ebuild. Otherwise slot 2 version of xmlgraphics-commons would be acceptable.

marbre:
So, I just took a closer look into your patches. One minor issue to fix.

One additional note for future contribution. Please check the [CONTRIBUTING.md](https://github.com/gentoo-science/sci/blob/master/CONTRIBUTING.md), especially the section about `repo-commit`. Since no line like `Package-Manager: portage-2.2.20` appears in your commit messages, I think you directly used `git`. You're commits are fine, but `repo-commit` is preferable. Anyway, `repoman full` doesn't complain anywhere regarding the modified ebuilds.

Last but not least, thank you for taking the effort to push the batik changes to the tree (and not "only" taking care of your pull request). I really appreciate this.

sci-mathematics/scilab/files/scilab-5.5.2-xmlgraphics-common-2.0.patch
khumba: I think, that the 'thirdparty' references are only for building a bundled Scilab package, and at no point are any of them used on Gentoo.  The thirdparty directory doesn't exist in the built source tree nor in the installed image, and Scilab runs fine without these changes.  So I feel like it would be better to remove these changes rather than to lock the ebuild to a specific version of xmlgraphics-common and batik.  Thoughts?

khumba: I checked that Scilab would load the batik-1.8 jars even without the thirdparty changes, and it does, so I've removed all thirdparty changes from the patches.

khumba:
Yes, I did use git directly; I didn't realize that repo-commit added something important to the commit message, and I wanted to have a multi-line commit message.  I've redone the commits using repo-commit.  I assume it's okay that I didn't use 'hub' :).

Marius, do you know anything about the decision not to ship batik-all.jar as of batik-1.8?  I'd like not to have to add this ebuild, but without looking into it too deeply, *without* any additional changes, the ebuild's `java-pkg_jar-from --build-only batik-1.8 batik-all.jar` line fails:

    >>> Preparing source in /var/tmp/portage/sci-mathematics/scilab-5.5.2/work/scilab-5.5.2 ...
     * Applying scilab-5.5.2-followlinks.patch ... [ ok ]
     * Applying scilab-5.5.2-gluegen.patch ... [ ok ]
     * Applying scilab-5.5.2-fix-random-runtime-failure.patch ... [ ok ]
     * Applying scilab-5.5.2-accessviolation.patch ... [ ok ]
     * Applying scilab-5.5.2-missinglib.patch ... [ ok ]
     * Applying scilab-5.5.2-batik-1.8.patch ... [ ok ]
     * Applying scilab-5.5.2-fop-2.0.patch ... [ ok ]
     * Applying scilab-5.5.2-xmlgraphics-common-2.0.patch ... [ ok ]
     * Applying scilab-5.5.2-java-heap.patch ... [ ok ]
     * ERROR: sci-mathematics/scilab-5.5.2::science failed (prepare phase):
     *   Failed to find batik-all.jar in batik-1.8
     * 
     * Call stack:
     *     ebuild.sh, line   93:  Called src_prepare
     *   environment, line 5273:  Called java-pkg_jar-from '--build-only' 'batik-1.8' 'batik-all.jar'
     *   environment, line 3982:  Called die
     *           die "Failed to find ${target_jar:-jar} in ${target_pkg}";
     * 
     * If you need support, post the output of `emerge --info '=sci-mathematics/scilab-5.5.2::science'`,
     * the complete build log and the output of `emerge -pqv '=sci-mathematics/scilab-5.5.2::science'`.
    !!! When you file a bug report, please include the following information:
    GENTOO_VM=icedtea-bin-7  CLASSPATH="" JAVA_HOME="/opt/icedtea-bin-7.2.5.5"
    JAVACFLAGS="-source 1.5 -target 1.5" COMPILER=""
    and of course, the output of emerge --info =scilab-5.5.2

and changing that to just say `batik.jar`, the configure then fails:

    checking batik-all... no
    configure: WARNING: Could not find or use the Java package/jar batik-all used by Apache SVG Library (looking for package org.apache.batik.parser.Parser)
    checking batik... /usr/share/batik-1.8/lib/batik-parser.jar
    checking minimal version (1.8) of batik... configure: error: Wrong version of batik. Expected at least 1.8. Found 

and I don't know how to fix this.  I'm tempted to file a bug on the Gentoo tracker.

marbre:
`repo-commit` mainly executes `repoman`, updates the Manifest if necessary and (depending on the options) adds a ChangeLog entry. So in this case there were no reason to redo the commits, since repoman was fine and also ChangeLog entry was commited.

Regarding your question, no, I don't know anything about the batik ebuild. My changes were done with the goal to get travis back to work. The repoman call executed by travis failed due to unfulfilled dependencies. But I agree that it is a good idea to discuss this with the gentoo maintainer of batik. So feel free to fill a bug report. I suggest, well maybe wait two or three days more, since it might than be clarified if we need an additional batik ebuild.

khumba:
Bug filed about Batik here: https://bugs.gentoo.org/show_bug.cgi?id=556062

I've also noticed that the ebuild's dependency on batik is wrong.  It should be `gui? ( dev-java/batik:1.8 )`, not `gui? ( doc? ( dev-java/batik:1.8 ) )` (per configure.ac, [1], and personal experience testing exporting SVG graphics).  That's easy to fix, but then the `java-pkg_jar-from --build-only batik-1.8 batik-all.jar` line doesn't actually seem to have any difference to the success of the build.  Rather than shifting that line as well, should I just remove it?

[1] http://wiki.scilab.org/Dependencies%20of%20Scilab%205.X

marbre:
Well that seems to be only the half truth. Below the *GUI/Console* table, they state that `fop, batik, xmlgraphics-commons and jeuclid-core are also needed for the documentation build process.` Actually batik is than required for both useflags. It might be okay to set `gui? ( dev-java/batik:1.8 )` and `doc? ( dev-java/batik:1.8 )`. This would also apply to some other dependencies. @jlec Should we *duplicate* the deps for `gui?` and `doc?`?

jlec:
You could define some *DEPEND variable and use that.

chewi:
@marbre I believe this line takes care of that.

```sh
REQUIRED_USE="xcos? ( gui ) doc? ( gui )"
```

khumba:
The REQUIRED_USE is appropriate, since configure.ac only looks at `--with-gui` before requiring batik.  It should also be in CDEPEND, not DEPEND.  I've fixed that but I'm waiting on the batik-all issue before I push anything new.

khumba:
Okay, now that Gentoo bug 556062 is resolved (batik-1.8-r2 bumped not to delete batik-all.jar), we don't need to add it here, so I've removed it.  I've also moved batik to the common dependency list.  Please take a look; thanks.

marbre:
Thanks you@khumba for your effort. Merged.

PR #457, opened 2015-07-26
Fixes orphaned Java HDF5 source link in ebuild.

cjhanks:


cjhanks:
After looking at this, maybe the previous version was okay (?).  The one currently available via layman is definitely wrong though.

PR #462, opened 2015-08-16
sci-biology/salmon: fix deps & version bump to 0.4.2

mschubert:


sci-biology/salmon/salmon-0.4.2.ebuild
```diff
@@ -0,0 +1,22 @@
...
+
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
+IUSE=""
+
+PATCHES=( "${FILESDIR}"/${PN}-0.3.2-no-boost-static.patch )
+
+DEPEND="dev-libs/boost
+        dev-libs/jemalloc"
```
jlec: It looks like the indention are spaces. Please use ```repoman full```

marbre: https://travis-ci.org/gentoo-science/sci/builds/75818212:
sci-biology/salmon/salmon-0.4.2.ebuild: Malformed Id header on line: 3
sci-biology/salmon/salmon-0.4.2.ebuild: Ebuild contains leading spaces on line: 21

@jlec Regarding the malformed header, I belive this due to the change of the header format, with regard to gentoos change from cvs to git. How should we proceed in the overlay?

jlec: @marbre I will convert it to be the same as in the tree. Nevertheless, repoman also complains about the $id thing. We need to wait.

mschubert:
Oh no, this was my mistake - I used `repoman full` but then converted to spaces after because I thought I remembered it from somewhere. Sorry, fixed now.

PR #466, opened 2015-08-24
sci-libs/symengine: add sympy-symengine

mschubert:

@jlec:
I didn't find anything about the `boost` USE flag in the docs, so I made it a required dep.


sci-libs/symengine/symengine-0.1.0.ebuild
```diff
@@ -0,0 +1,46 @@
...
+RDEPEND="${DEPEND}
+	python? ( dev-python/numpy[${PYTHON_USEDEP}] )"
+
+CMAKE_BUILD_TYPE=Release
+
+src_configure() {
+	local mycmakeargs=(
+		-DCMAKE_INSTALL_PREFIX:PATH="${EPREFIX}"/usr \
+		$(cmake-utils_use_with python) \
+		$(cmake-utils_use_with openmp)
```
marbre: I could imagine it should be checked of the used gcc is openmp compatible, see https://github.com/gentoo-science/sci/blob/master/sci-libs/fftw/fftw-3.3.4-r1.ebuild#L35 for example.

sci-libs/symengine/symengine-9999.ebuild
```diff
@@ -0,0 +1,47 @@
...
+RDEPEND="${DEPEND}
+	python? ( dev-python/numpy[${PYTHON_USEDEP}] )"
+
+CMAKE_BUILD_TYPE=Release
+
+src_configure() {
+	local mycmakeargs=(
+		-DCMAKE_INSTALL_PREFIX:PATH="${EPREFIX}"/usr \
+		$(cmake-utils_use_with python) \
+		$(cmake-utils_use_with openmp)
```
marbre: Dito.

sci-libs/symengine/symengine-0.1.0.ebuild
```diff
@@ -0,0 +1,46 @@
...
+SRC_URI="https://github.com/sympy/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
+IUSE="python threads openmp"
+
+DEPEND="
+	dev-libs/jemalloc
+	dev-libs/boost
```
marbre: Is the `python` use flag required here?

sci-libs/symengine/symengine-9999.ebuild
```diff
@@ -0,0 +1,47 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Header: $
```
marbre: Replace by `# $Id$`. This changed with the gentoo tree migrating to git.

sci-libs/symengine/symengine-0.1.0.ebuild
mschubert: I'm not sure I understand the question - symengine can be built with or without python support.

kiwifb: Note that I had a discussion with upstream and python bindings will be split. Technically the python binding, which are definitely optional are not `python2` specific but work also with `python3`. But the current packaging makes it hard to build against several `python`.

marbre: Well the question is if symengine requires Boost.Python. If yes, the `python` use flag has to be conditionally added to `dev-libs/boost`. As I just realized symengine is written in C++, thus other parts of boost might be required.

mschubert: @marbre It does not require boost-python, just `date_time`, `system` and `filesystem`.
@kiwifb I'll give the alternate python3 version a try, give me a couple of minutes (I suspect as long as I have `python-single-r1` it will only support one version of python at a time anyway?)

mschubert: Ok, python didn't work as I expected, but the rest is updated.

kiwifb: That's strange but I don't know what you have done.

mschubert: I thought that `python-single-r1` will just specify which versions the ebuild is compatible with and then the user change choose using the `USE=python_single_target_python*` flag - but I was wrong, with `python-single-r1` I can only have one python version in `${PYTHON_COMPAT}`. I don't see how I can easily set "python2_7 OR python3_3", because then I could use the cmake flag `DPYTHON_BIN=python*`.

kiwifb:
For the `boost` useflag, you can just put it in `IUSE` and add it to `metadata.xml` with a description.

mschubert:
Thanks for the hint, I re-added the `boost` flag.

sci-libs/symengine/symengine-0.1.0.ebuild
kiwifb: On 08/25/15 09:26, Michael Schubert wrote:
> In sci-libs/symengine/symengine-0.1.0.ebuild
> <https://github.com/gentoo-science/sci/pull/466#discussion_r37806930>:
> 
>> +PYTHON_REQ_USE=""
>> +
>> +inherit cmake-utils python-single-r1
>> +
>> +DESCRIPTION="Fast symbolic manipulation library, written in C++"
>> +HOMEPAGE="https://github.com/sympy/symengine"
>> +SRC_URI="https://github.com/sympy/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
>> +
>> +LICENSE="MIT"
>> +SLOT="0"
>> +KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
>> +IUSE="python threads openmp"
>> +
>> +DEPEND="
>> +	dev-libs/jemalloc
>> +	dev-libs/boost
> 
> I thought that |python-single-r1| will just specify which versions the
> ebuild is compatible with and then the user change choose using the
> |USE=python_single_target_python*| flag - but I was wrong, with
> |python-single-r1| I can only have one python version in
> |${PYTHON_COMPAT}|. I don't see how I can easily set "python2_7 OR
> python3_3", because then I could use the cmake flag |DPYTHON_BIN=python*|.
> 

As I said I requested upstream to separate the python bindings and they
agreed that it was a good idea that they will act upon. So this is just
temporary.

For the rest read carefully
https://devmanual.gentoo.org/eclass-reference/python-single-r1.eclass/index.html
and https://wiki.gentoo.org/wiki/Project:Python/python-single-r1 you
probably forgot to use
python-single-r1_pkg_setup.

I was kind of planning to make the ebuild but you are doing a good job
given the state of upstream.

Francois

sci-libs/symengine/symengine-0.1.0.ebuild
```diff
@@ -0,0 +1,53 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
+
+EAPI=5
+
+PYTHON_COMPAT=( python2_7 )
+PYTHON_REQ_USE=""
```
jlec: You can drop this if empty.

sci-libs/symengine/symengine-0.1.0.ebuild
```diff
@@ -0,0 +1,53 @@
...
+inherit cmake-utils python-single-r1
+
+DESCRIPTION="Fast symbolic manipulation library, written in C++"
+HOMEPAGE="https://github.com/sympy/symengine"
+SRC_URI="https://github.com/sympy/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
+IUSE="python threads openmp boost"
```
jlec: Please sort them alphabetically.

jlec: You need to add a line with ```REQUIRED_USE="python? ( ${PYTHON_REQUIRED_USE}" )```. That way the necessary USE flags for python are set.

sci-libs/symengine/symengine-0.1.0.ebuild
```diff
@@ -0,0 +1,53 @@
...
+
+DEPEND="
+	dev-libs/jemalloc
+	dev-libs/boost
```
jlec: Does boost need python support with USE=python? Or are the python bindings pure python code?

sci-libs/symengine/symengine-0.1.0.ebuild
```diff
@@ -0,0 +1,53 @@
...
+	python? ( dev-python/setuptools[${PYTHON_USEDEP}] )"
+RDEPEND="${DEPEND}
+	python? ( dev-python/numpy[${PYTHON_USEDEP}] )"
+
+CMAKE_BUILD_TYPE=Release
+
+src_configure() {
+	local mycmakeargs=(
+		-DCMAKE_INSTALL_PREFIX:PATH="${EPREFIX}"/usr \
```
jlec: You don't need line breaks with ```\```in bash arrays (```a=()```).

sci-libs/symengine/symengine-0.1.0.ebuild
```diff
@@ -0,0 +1,53 @@
...
+		$(cmake-utils_use_with boost) \
+		$(cmake-utils_use_with python)
+	)
+
+	if use openmp; then
+		if [[ $(tc-getCC) == *gcc ]] && ! tc-has-openmp; then
```
jlec: You need to wrap this check in a check of BUILD_TYPE. Please see the devmanual for that variable.

mschubert: Google does not return any results for inurl:devmanual.gentoo.org "BUILD_TYPE". If I look at other gentoo-sci ebuilds, all I find is a comment saying "this should be checked for BUILD_TYPE!=binary".

sci-libs/symengine/symengine-0.1.0.ebuild
mschubert: Cython only. But I did forget to add that, fixed now.

sci-libs/symengine/symengine-0.1.0.ebuild
kiwifb: Not sure what jlec is on about either. I guess you don't need to check the compiler for binary packages. However testing the compiler capabilities is best done in `pkg_setup`. So his should be moved. Finally, you test the compiler capabilities but do not set any options for the build. If `openmp` is used automagically there is no point in putting an option or checking for it.

sci-libs/symengine/symengine-0.1.0.ebuild
```diff
@@ -0,0 +1,55 @@
...
+DEPEND="
+	dev-libs/jemalloc
+	boost? ( dev-libs/boost )
+	python? (
+		dev-python/cython[${PYTHON_USEDEP}]
+		dev-python/setuptools[${PYTHON_USEDEP}] )"
+RDEPEND="${DEPEND}
+	python? ( dev-python/numpy[${PYTHON_USEDEP}] )"
+
+CMAKE_BUILD_TYPE=Release
```
kiwifb: Why do you need to set this instead of leaving it alone? https://devmanual.gentoo.org/eclass-reference/cmake-utils.eclass/index.html

sci-libs/symengine/symengine-0.1.0.ebuild
kiwifb: Further if `openmp` is a requirement (i.e. it cannot be turned off) when `CC=gcc` it is really properly check in `pkg_setup` and there is no need for it in `IUSE`.

jlec: @mschubert 
I meant [[ ${MERGE_TYPE != binary ]], see https://devmanual.gentoo.org/ebuild-writing/eapi/index.html. There are occasions like for binary only distros derived from gentoo which do not install any compiler but still are using ebuilds to install packages. On their system that check will always fail in a false positive way if not wrapped by MERGE_TYPE.
Now the question is, if the openmp support is automagic, then I would recommend to hard require it. See the gentoo docs about automagic dependencies and why they are bad.

Secondly, the check should be moved, but into pkg_pretend(). That check is even run before anything starts, see https://devmanual.gentoo.org/ebuild-writing/functions/.

sci-libs/symengine/symengine-0.1.0.ebuild
jlec: Indeed, if things are not blowing up because the TYPE is set to Gentoo, I would leave it as is.

sci-libs/symengine/symengine-0.1.0.ebuild
```diff
@@ -0,0 +1,55 @@
...
+IUSE="boost openmp python threads"
+REQUIRED_USE="python? ( ${PYTHON_REQUIRED_USE} )"
+
+DEPEND="
+	dev-libs/jemalloc
+	boost? ( dev-libs/boost )
+	python? (
+		dev-python/cython[${PYTHON_USEDEP}]
+		dev-python/setuptools[${PYTHON_USEDEP}] )"
+RDEPEND="${DEPEND}
```
jlec: Good that you added the missing dep, but now you are depending during runtime on the buildtime deps. I would recommend to switch it around and have ```DEPEND="${RDEPEND}```

sci-libs/symengine/symengine-0.1.0.ebuild
marbre: I assume it will blow up: https://github.com/sympy/symengine/blob/master/CMakeLists.txt#L21

kiwifb: I will have a word with upstream, that's a stupid constraint.

marbre:
@mschubert If you think all issues are solved (expect the CMAKE_BUILD_TYPE) give us a hint, that this is maybe ready to merge ;)

mschubert:
@marbre @jlec Should be all done now.

I'm not changing `CMAKE_BUILD_TYPE` for now, until upstream allows for the "Gentoo" build.

sci-libs/symengine/symengine-0.1.0.ebuild
```diff
@@ -0,0 +1,59 @@
...
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
+IUSE="boost openmp python threads"
+REQUIRED_USE="python? ( ${PYTHON_REQUIRED_USE} )"
+
+RDEPEND="
+	dev-libs/jemalloc
+	boost? ( dev-libs/boost )
+	python? ( dev-python/numpy[${PYTHON_USEDEP}] )"
+DEPEND="${RDEPEND}
+	openmp? ( sys-cluster/openmpi )
```
marbre: Why does OpenMP require OpenMP***I*** ?

sci-libs/symengine/symengine-9999.ebuild
```diff
@@ -0,0 +1,60 @@
...
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
+IUSE="boost openmp python threads"
+REQUIRED_USE="python? ( ${PYTHON_REQUIRED_USE} )"
+
+RDEPEND="
+	dev-libs/jemalloc
+	boost? ( dev-libs/boost )
+	python? ( dev-python/numpy[${PYTHON_USEDEP}] )"
+DEPEND="${RDEPEND}
+	openmp? ( sys-cluster/openmpi )
```
marbre: Dito.

marbre:
@mschubert One more remark. Can you please check if it is a good idea to enable `openmp` and `threads`? Hopefully this results in two libraries, but not both enabled in a single library.

For example `fftw` builds `libfftw3*_threads.so` and `libfftw3*_omp.so`. It's normally not a good idea to combine both.

sci-libs/symengine/symengine-0.1.0.ebuild
mschubert: That's because I misunderstood what OpenMP does. Fixed now.

mschubert:
Enabling both `openmp` and `threads` does not change the name of the library built.

I'm not sure I understand, why is this an issue?

kiwifb:
`openmp` is an easy way to use `threads` so having both at the same time is a bit redundant unless your are making a binary package and you are not sure if the `openmp` runtime will be available. You usually use one or the other but not both at the same time.

marbre:
Thanks @kiwifb for helping out :)
After a very short look into the [CMakeLists.txt](https://github.com/sympy/symengine/blob/master/CMakeLists.txt), it seems that both can be enabled and are not blocking each other. This should maybe get clarified by upstream.
I assume a `REQUIRED_USE` should be added, to avoid both flags enabled. Please take a look into the [Conflicting USE Flags](https://devmanual.gentoo.org/general-concepts/use-flags/]) section.

marbre:
Btw, is there something wrong with the committing author? You're committing as layman.

marbre:
@mschubert I would prefer adding `openmp? ( !threads ) threads? ( !openmp )` to `REQUIRED_USE`. @jlec May I asked for your recommendation on this?
After clarifying and possible modifying this, we're happy to merge your request.

jlec:
@mschubert
```ebuild
REQUIRED_USE=" ?? ( openmp threads)"
```
https://devmanual.gentoo.org/ebuild-writing/eapi/index.html

mschubert:
Ok, updated.

PR #475, opened 2015-09-01
dev-python/jupyterhub + depedencies

marbre:
This pull request adds jupyterhub to the science overlay. jupyterhub is a multi-user server, that manages and provided notebooks for several users.

@jlec Here we finally go with the npm stuff in the configurable-http-proxy ebuild.


net-libs/configurable-http-proxy/configurable-http-proxy-9999.ebuild
```diff
@@ -0,0 +1,42 @@
...
+	"
+DEPEND="${RDEPEND}"
+
+src_install() {
+	# npm_src_install() from https://github.com/neurogeek/gentoo-overlay/blob/master/eclass/npm.ecla
+
+	local npm_files="${NPM_FILES} ${NPM_EXTRA_FILES}"
+	local node_modules="${D}/usr/$(get_libdir)/node_modules/${NPM_MODULE}"
+
+	mkdir -p ${node_modules} || die "Could not create DEST folder"
```
jlec: use dodir here

net-libs/configurable-http-proxy/configurable-http-proxy-9999.ebuild
```diff
@@ -0,0 +1,42 @@
...
+
+	for f in ${npm_files}
+	do
```
jlec: move this to line 34

net-libs/configurable-http-proxy/configurable-http-proxy-9999.ebuild
```diff
@@ -0,0 +1,42 @@
...
+		if [[ -e "${S}/$f" ]]; then
+			cp -r "${S}/$f" ${node_modules}
```
jlec: missing die.

net-libs/configurable-http-proxy/configurable-http-proxy-9999.ebuild
```diff
@@ -0,0 +1,42 @@
...
+else
+	SRC_URI="https://github.com/jupyter/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="~amd64"
+fi
+
+LICENSE="BSD"
+SLOT="0"
+
+RDEPEND="
+	>=net-libs/nodejs-0.8.28[npm]
```
jlec: is npm really need at runtime? We are not installing anything with it.

net-libs/configurable-http-proxy/configurable-http-proxy-9999.ebuild
```diff
@@ -0,0 +1,42 @@
...
+	"
+DEPEND="${RDEPEND}"
+
+src_install() {
+	# npm_src_install() from https://github.com/neurogeek/gentoo-overlay/blob/master/eclass/npm.ecla
```
jlec: Any chance to get this eclass in shape and move it into the overlay?

marbre: @jlec, I discussed this by email with @neurogeek some time ago. One of the main reasons why the npm ebuilds and the eclass didn't made it to the tree is that npm often uses local installations instead of global installations and therefore installs package duplicates. Further, versioning of npm packages isn't really compatible with ebuild versioning. Thus as long as there are no a bundle of packages requiring a npm.eclass, I would try to avoid the effort.

marbre:
@jlec Finally, much more additional work is needed to get configurable-http-proxy working. First of all, my ebuild is scrap. I missed to set several variables, which are set in the eclass. This results in an "empty" install. I've already redone it, but configurable-http-proxy has additional dependencies normally installed via npm and pulled in with `npm install -g jupyter/configurable-http-proxy`:
~~~
/usr/bin/configurable-http-proxy -> /usr/lib/node_modules/configurable-http-proxy/bin/configurable-http-proxy
configurable-http-proxy@0.4.1-dev /usr/lib/node_modules/configurable-http-proxy
├── strftime@0.8.4
├── commander@2.8.1 (graceful-readlink@1.0.1)
├── http-proxy@1.11.2 (eventemitter3@1.1.1, requires-port@0.0.1)
└── winston@0.8.3 (cycle@1.0.3, stack-trace@0.0.9, eyes@0.1.8, isstream@0.1.2, colors@0.6.2, async@0.2.10, pkgin
~~~
Thus, this pull request in incomplete. Contrary to my previous comment, an npm.eclass might be helpful here.

neurogeek:
I wouldn't mind working on getting an npm.eclass ready, but there are so many low quality packages in npm that it would be hard to streamline the ebuild creation process (and minimize breakage all over the npm ebuild tree), there are versioning problems, non-global install problems, naming problems (There is a package called 'grunt-', sigh!). 

There is also another thing.. a dev-npm or dev-nodejs category would hold too many packages in it, since it seems that it is encouraged (implicitly?) to create many small npm packages: i.e consider Grunt.. It has 20 direct dependencies... so just to package Grunt, you need to packages at least 50 npm packages!!.

PR #476, opened 2015-09-01
dev-python/notebook: Use system mathjax

flacjacket:
Also, install all python 2 dependencies for tests

### DIFF
--- a/dev-python/notebook/notebook-9999.ebuild
+++ b/dev-python/notebook/notebook-9999.ebuild
@@ -24,6 +24,7 @@ IUSE="doc test"
 CDEPEND="
 	dev-python/setuptools[${PYTHON_USEDEP}]"
 RDEPEND="${CDEPEND}
+	dev-libs/mathjax
 	dev-python/jinja[${PYTHON_USEDEP}]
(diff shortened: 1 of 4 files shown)


dev-python/notebook/notebook-9999.ebuild
```diff
@@ -46,6 +48,20 @@ DEPEND="${RDEPEND}
 	)
 	"
 
+python_prepare_all() {
+	# disable bundled mathjax
+	sed -i 's/^.*MathJax.*$//' bower.json
+	sed -i 's/mj(/#mj(/' setupbase.py
```
jlec: Always add die to external commands.

flacjacket:
@jlec Added `|| die` to the sed and ln calls

marbre:
I assume we could than also remove line https://github.com/gentoo-science/sci/blob/master/dev-python/ipython/ipython-4.0.0-r1.ebuild#L106 from the ipython ebuild?

flacjacket:
Yeah, the mathjax stuff is in notebook, that line should be removed from
ipython, I can add that to this pr

flacjacket:
@marbre Added

marbre:
Thank you for the pull request. Merged.

PR #481, opened 2015-09-02
Various Jupyter doc and deps fixes

flacjacket:
* Fix some minor dependencies issues with packages and running tests

* Add doc use flags to all Jupyter/IPython packages with sphinx docs

### DIFF
--- a/dev-python/ipywidgets/ipywidgets-9999.ebuild
+++ b/dev-python/ipywidgets/ipywidgets-9999.ebuild
@@ -29,6 +29,7 @@ RDEPEND="
 	"
 DEPEND="${RDEPEND}
 	test? (
+		$(python_gen_cond_dep 'dev-python/mock[${PYTHON_USEDEP}]' python2_7)
 		dev-python/nose[${PYTHON_USEDEP}]
 		dev-python/coverage[${PYTHON_USEDEP}]
 	)
--- a/dev-python/nbconvert/nbconvert-9999.ebuild
+++ b/dev-python/nbconvert/nbconvert-9999.ebuild
@@ -64,3 +64,11 @@ python_install_all() {
 	use doc && HTML_DOCS=( docs/build/html/. )
 	distutils-r1_python_install_all
 }
+
+pkg_postinst() {
+	if ! has_version app-text/pandoc ; then
+		einfo "Pandoc is required for converting to formats other than Python,"
+		einfo "HTML, and Markdown. If you need this functionality, install"
+		einfo "app-text/pandoc."
+	fi
+}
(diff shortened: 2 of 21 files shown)


dev-python/nbconvert/nbconvert-9999.ebuild
```diff
@@ -22,9 +22,10 @@ fi
 
 LICENSE="BSD"
 SLOT="0"
-IUSE="test"
+IUSE="doc test"
 
 RDEPEND="
+	>=app-text/pandoc-1.12.1
```
marbre: For ipython (https://github.com/gentoo/gentoo/blob/master/dev-python/ipython/ipython-3.2.1.ebuild#L144) pandoc was optional. As far as I get it nbconvert is an alternative to pandoc, based on jinja2 templates.

I would like to avoid adding pandoc to the dependencies as this would pull in 90 new packages, mostly from `dev-haskell`. Please recheck, as i can't find any word on pandoc in the nbconvert repository. Further, tests pass without pandoc installed.

marbre: Finally found the docs: https://nbconvert.readthedocs.org/en/latest/install.html#pandoc

Anyway, I would appreciate an optional dependency or a hint like previously in ipython (were I will remove it for 4.0.0). Conversion between python, html and markdown is usable without pandoc.

flacjacket: Alright, I pulled the commit that added this and added one that notes pandoc as an optional dependency.

flacjacket:
With Pandoc installed, some of the otherwise skipped tests are failing, I just added a patch to backport a fix from the current git version.

dev-python/ipywidgets/ipywidgets-9999.ebuild
```diff
@@ -29,6 +29,7 @@ RDEPEND="
 	"
 DEPEND="${RDEPEND}
 	test? (
+		$(python_gen_cond_dep 'dev-python/mock[${PYTHON_USEDEP}]' python2_7)
```
jlec: Is mock only needed for python2.7? Or did you mean ```python_gen_usedep```?

marbre: @jlec This is similar to https://github.com/gentoo/gentoo/blob/master/dev-python/ipython/ipython-3.2.1.ebuild#L67 and I assume mock is really only needed for python 2.7 (https://github.com/ipython/ipywidgets/blob/master/setup.py#L286).

flacjacket: Right, it is only for 2.7 https://github.com/ipython/ipywidgets/blob/master/setup.py#L286, all of the IPython packages that use mock do a try/except ImportError on unittest.mock before importing from mock (https://github.com/ipython/ipywidgets/blob/master/ipywidgets/widgets/tests/test_interaction.py#L8)

PR #483, opened 2015-09-03
dev-python/jupyter: Initial commit

flacjacket:
Here is the Jupyter metapackage. It doesn't install anything on its own, but it pulls in all the various Jupyter packages. There are some docs that can also be built for Jupyter.

Other than the docs, all this provides is some convenience, something similar could be accomplished by adding a use flag for jupyter_console to ipython, since it already has use flags for the notebook and qt console.


jlec:
@marbre What do you think?
@flacjacket Looks good to me.

marbre:
@jlec Since jupyter in the repo is currently not usable with any other kernels than ipython, it is not essential to have a jupyter meta-package. You get nearly the same by installing ipython[notebook]. Anyway, I already started by myself to write a jupyter meta-package. Additional jupyter kernel will hopefully make it into the overlay sooner or later. Therefore merging.. Thanks for the PR @flacjacket.

PR #484, opened 2015-09-07
MathJax.js is in the unpacked/ directory on gentoo

tdaff:
Jupyter notebooks were complaining about missing MathJax, on my system using vanilla mathjax-2.4 ebuilds it is installed in /usr/share/mathjax but MathJax.js is in the unpacked subdirectory.


marbre:
I just upgraded to mathjax-2.4 (version 2.2 gives me Math Processing Erros) and MathJax.js can be found in `/usr/share/mathjax` **and** `/usr/share/mathjax/MathJax`. On my three systems, `jupyter notebook` does not complain about missing MathJax.

jlec:
I also rechecked. In the unpacked dir you will find the not minimized version.

marbre:
Comparing the folder installed by dev-python/notebook if the bundle mathjax is used, it looks like also dev-python/notebook installs the minimized version.

tdaff:
I just checked and it looks like something has corrupted my MathJax installation, re-emerging gives me the minified version again. Sorry for the confusion, and thanks for helping out!

jlec:
Thanks for clarifying.

marbre:
@tdaff, a little confusion isn't always that bad ;-) Without your pull request, I wouldn't have noticed that mathjax-2.2 leads to math parsing errors. Just added a commit to pull in >=mathjax-2.4 as dependency. So feel free to send further pull requests.

PR #486, opened 2015-09-12
sci-libs/rdkit: initial commit

mschubert:
> Open source toolkit for cheminformatics


sci-libs/rdkit/rdkit-2015.03.1.ebuild
```diff
@@ -0,0 +1,47 @@
...
+HOMEPAGE="http://www.rdkit.org/"
+SRC_URI="https://github.com/rdkit/rdkit/archive/Release_${PV//./_}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
+IUSE="+python -static-libs"
+REQUIRED_USE="python? ( ${PYTHON_REQUIRED_USE} )"
+
+RDEPEND="dev-libs/boost
```
jlec: No python support needed in boost?

sci-libs/rdkit/rdkit-2015.03.1.ebuild
```diff
@@ -0,0 +1,47 @@
...
+		-DRDK_INSTALL_INTREE=0
+		-DRDK_BUILD_CPP_TESTS=OFF
+	)
+
+	if ! use static-libs; then
+		mycmakeargs+=( -DRDK_INSTALL_STATIC_LIBS=OFF )
+	fi
+
+	if ! use python; then
+		mycmakeargs+=( RDK_BUILD_PYTHON_WRAPPERS= )
```
marbre: **-D**RDK_BUILD_PYTHON_WRAPPERS=**OFF**? As this request was already merged, I commited ed8fb25203c90f2049a41a421c54713641f7cd66. Anyway, please correct me if I'm wrong.

mschubert: Oh no, copy paste error that I missed :disappointed: 

I'll send a PR for fixing it - but I don't entirely like the syntax, it will break for `+static-libs` if upstream ever sets `RDK_INSTALL_STATIC_LIBS=OFF` by default.

Is there any Gentoo function that just returns `ON` or `OFF` given a USE flag? I haven't seen it in the devmanual so far.

marbre: Unfortunately, not for these. The available arguments are [1]:
* `WITH_`
* `ENABLE_`
* `NO_`
* `WANT_`
* `BUILD_`
* `HAVE_`
* `USE_`

Another option is to name the use flag like the option and to use `-DFOO=ON` (cmake-utils_use).
For this use case, I don't see an alternative. Anyway, that is not a problem for 2015.03.1, but 9999 might break. That's one of the reasons why live ebuilds don't contain keywords ;)



[1] https://devmanual.gentoo.org/eclass-reference/cmake-utils.eclass/

jlec: This is what you are looking for:
```
make-utils_use <USE flag> [flag name]
    Based on use_enable. See ebuild(5).

    `cmake-utils_use foo FOO` echoes -DFOO=ON if foo is enabled and -DFOO=OFF if it is disabled. 
```

mschubert:
Thanks! I added a PR using `cmake-utils_use` here: https://github.com/gentoo-science/sci/pull/487

PR #489, opened 2015-09-14
Remove wxwidgets from ipython dependencies

suhr:


kiwifb:
Thanks for doing something about the issue I reported suhr. The travis problem is that you changed the `wxwidgets` use flag which is a standard use flag to `widgets` which is not. In effect you introduced a new use flag. Such use flags need to be documented by an entry in `metadata.xml` in the same directory. So you should edit `metadata.xml` in the `ipython` folder, there are several entries in the file already that you can model your entry on. I would, personally, also put `ipywidgets` in `RDEPEND` in the `dev-python/notebook` ebuild.

marbre:
This pull request solves Issue #489.

Please add a ChangeLog entry, to document the changes. Furthermore, please use `repoman full` to check in ebuilds in and `repo commit` (which automatically runs repoman) to commit changes in future, as described in https://github.com/gentoo-science/sci/blob/master/CONTRIBUTING.md.

dev-python/ipywidgets/ipywidgets-9999.ebuild
```diff
@@ -25,11 +25,9 @@ IUSE="test"
 RDEPEND="
 	dev-python/traitlets[${PYTHON_USEDEP}]
 	dev-python/ipykernel[${PYTHON_USEDEP}]
-	$(python_gen_cond_dep 'dev-python/wxpython:*[${PYTHON_USEDEP}]' python2_7)
 	"
 DEPEND="${RDEPEND}
 	test? (
-		$(python_gen_cond_dep 'dev-python/mock[${PYTHON_USEDEP}]' python2_7)
```
marbre: Why is this line dropped? @flacjacket Since you added this dependency, maybe you could clarify?

flacjacket: Mock is needed for Python 2.7 to run the tests, I don't think this line should be removed

https://github.com/ipython/ipywidgets/blob/master/ipywidgets/widgets/tests/test_interaction.py#L11

suhr:
I will rebase the branch when everything is ok.

dev-python/ipython/ipython-4.0.0-r1.ebuild
```diff
@@ -63,7 +62,8 @@ DEPEND="${CDEPEND}
 		dev-python/rpy[${PYTHON_USEDEP}]
 		>=dev-python/sphinx-1.1[${PYTHON_USEDEP}]
 		>=www-servers/tornado-4.0[${PYTHON_USEDEP}]
-	)"
+	)
+	wxwidgets? ( dev-python/wxpython[${PYTHON_USEDEP}] )"
```
marbre: Please move to CDEPEND and modify according to in https://github.com/gentoo/gentoo/blob/master/dev-python/ipython/ipython-3.2.1.ebuild#L34

marbre:
Beside the last line note, a [revision bump](https://devmanual.gentoo.org/general-concepts/ebuild-revisions/) to -r2 might be helpful.

suhr:
Looks like everything is fine. Rebase?

kiwifb:
Looks good to me.

suhr:
Rebased.

marbre:
@suhr Thanks for the fix. Merged.

PR #493, opened 2015-09-22
dev-python/notebook: Don't need setuptools for scripts

flacjacket:
See https://github.com/jupyter/notebook/issues/339

### DIFF
--- a/dev-python/notebook/ChangeLog
+++ b/dev-python/notebook/ChangeLog
@@ -2,6 +2,10 @@
 # Copyright 1999-2015 Gentoo Foundation; Distributed under the GPL v2
 # $Id$
 
+  22 Sep 2015; Sean Vig <sean.v.775@gmail.com> notebook-4.0.5.ebuild,
+  notebook-9999.ebuild:
+  dev-python/notebook: Don't need setuptools for scripts
+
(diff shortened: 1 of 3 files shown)


marbre:
~~~
$ ls -la /var/tmp/portage/dev-python/notebook-4.0.4-r3/image/usr/bin
total 8
drwxr-xr-x 2 brehler brehler 4096 Sep 22 16:32 .
drwxr-xr-x 6 brehler brehler 4096 Sep 22 16:32 ..
lrwxrwxrwx 1 brehler brehler   31 Sep 22 16:32 jupyter-nbextension -> ../lib/python-exec/python-exec2
lrwxrwxrwx 1 brehler brehler   31 Sep 22 16:32 jupyter-notebook -> ../lib/python-exec/python-exec2
~~~

 ~~~
$ ls -la /var/tmp/portage/dev-python/notebook-4.0.5/image/usr/bin
ls: cannot access /var/tmp/portage/dev-python/notebook-4.0.5/image/usr/bin: No such file or directory
~~~

flacjacket:
Oh, so I just checked and `scripts/` is not in the source distribution. In that case that bug is still valid and this can be closed until that is fixed.

Issue #514, opened 2015-10-12
Pysurfer-0.6 patch fails


TheChymera:
Hi, I am trying to emerge pysurfer-0.6. The pathc apparently fails with:

```
***** 0.6-headless_build.patch *****
PWD: /var/tmp/portage/sci-biology/pysurfer-0.6/work/PySurfer-0.6
PATCH TOOL: patch -> /usr/bin/patch
VERSION INFO:
GNU patch 2.7.5
Copyright (C) 2003, 2009-2012 Free Software Foundation, Inc.
Copyright (C) 1988 Larry Wall

License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Larry Wall and Paul Eggert

====================================

PATCH COMMAND:  patch -p0 -g0 -E --no-backup-if-mismatch  < '/home/chymera/src/neurogentoo/sci-biology/pysurfer/files/0.6-headless_build.patch'

====================================
can't find file to patch at input line 5
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
|diff --git a/setup.py b/setup.py
|index c8681c2..41c21a5 100644
|--- a/setup.py
|+++ b/setup.py
--------------------------
No file to patch.  Skipping patch.
2 out of 2 hunks ignored

patch program exited with status 1
====================================

PATCH COMMAND:  patch -p1 -g0 -E --no-backup-if-mismatch  < '/home/chymera/src/neurogentoo/sci-biology/pysurfer/files/0.6-headless_build.patch'

====================================
checking file setup.py
Hunk #1 FAILED at 10.
Hunk #2 succeeded at 30 with fuzz 2 (offset 10 lines).
1 out of 2 hunks FAILED

patch program exited with status 1
====================================

PATCH COMMAND:  patch -p2 -g0 -E --no-backup-if-mismatch  < '/home/chymera/src/neurogentoo/sci-biology/pysurfer/files/0.6-headless_build.patch'

====================================
can't find file to patch at input line 5
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
|diff --git a/setup.py b/setup.py
|index c8681c2..41c21a5 100644
|--- a/setup.py
|+++ b/setup.py
--------------------------
No file to patch.  Skipping patch.
2 out of 2 hunks ignored

patch program exited with status 1
[...]

PR #515, opened 2015-10-13
sci-biology/pysurfer: removed deprecated patch section

TheChymera:
Finally fixed my patch issue. One hunk got deprecated. Solves #514


TheChymera:
@jlec ?

PR #521, opened 2015-10-19
dev-python/nilearn: new package

TheChymera:
Great software package, including some long, awaited python plotting functions for neuroimaging.


dev-python/nilearn/nilearn-0.1.4.1.ebuild
```diff
@@ -0,0 +1,27 @@
...
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
+
+DEPEND="
+	dev-python/setuptools[${PYTHON_USEDEP}]
+	"
+RDEPEND="
+	dev-python/numpy[${PYTHON_USEDEP}]
+	sci-libs/scikits_learn
```
marbre: PYTHON_USEDEP missing.

dev-python/nilearn/nilearn-9999.ebuild
```diff
@@ -0,0 +1,28 @@
...
+
+EAPI=5
+
+PYTHON_COMPAT=( python2_7 python3_4 )
+
+inherit distutils-r1 git-r3
+
+DESCRIPTION="Fast and easy statistical learning on NeuroImaging data"
+HOMEPAGE="http://nilearn.github.io/"
+SRC_URI=""
```
marbre: Drop it.

dev-python/nilearn/nilearn-9999.ebuild
```diff
@@ -0,0 +1,28 @@
...
+SLOT="0"
+KEYWORDS=""
+IUSE=""
+
+DEPEND="
+	dev-python/setuptools[${PYTHON_USEDEP}]
+	"
+RDEPEND="
+	dev-python/numpy[${PYTHON_USEDEP}]
+	sci-libs/scikits_learn
```
marbre: Dito.

dev-python/nilearn/nilearn-0.1.4.1.ebuild
```diff
@@ -0,0 +1,27 @@
...
+inherit distutils-r1
+
+DESCRIPTION="Fast and easy statistical learning on NeuroImaging data"
+HOMEPAGE="http://nilearn.github.io/"
+SRC_URI="mirror://pypi/n/${PN}/${PN}-0.1.4.post1.tar.gz -> ${P}.zip"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
```
marbre: Drop empty IUSE: https://devmanual.gentoo.org/ebuild-writing/variables/index.html#iuse

dev-python/nilearn/nilearn-9999.ebuild
```diff
@@ -0,0 +1,28 @@
...
+
+DESCRIPTION="Fast and easy statistical learning on NeuroImaging data"
+HOMEPAGE="http://nilearn.github.io/"
+SRC_URI=""
+EGIT_REPO_URI="git://github.com/nilearn/nilearn"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS=""
+IUSE=""
```
marbre: Dito.

dev-python/nilearn/nilearn-9999.ebuild
TheChymera: are you sure?  I see a lot of live packages mentioning it.

dev-python/nilearn/nilearn-0.1.4.1.ebuild
```diff
@@ -0,0 +1,27 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
+
+EAPI=5
+
+PYTHON_COMPAT=( python2_7 python3_4 )
```
marbre: What about python3_3?

marbre:
`If you are using nilearn plotting functionalities or running the examples, matplotlib >= 1.1.1 is required.` [https://github.com/nilearn/nilearn/blob/master/README.rst]

You may want to add a use flag and an use specific dependency. Furthermore, implementing a test phase could be nice.

dev-python/nilearn/nilearn-0.1.4.1.ebuild
```diff
@@ -0,0 +1,26 @@
...
+
+inherit distutils-r1
+
+DESCRIPTION="Fast and easy statistical learning on NeuroImaging data"
+HOMEPAGE="http://nilearn.github.io/"
+SRC_URI="mirror://pypi/n/${PN}/${PN}-0.1.4.post1.tar.gz -> ${P}.tar.gz"
```
marbre: You should use ${PV} here or add a ${MY_PY} using regex or the versionator eclass to avoid hardcoding the version number.

dev-python/nilearn/nilearn-9999.ebuild
marbre: IMHO it is needless.

dev-python/nilearn/nilearn-0.1.4.1.ebuild
TheChymera: I thought that was deprecated

dev-python/nilearn/nilearn-0.1.4.1.ebuild
TheChymera: I agree, though whatever else I put in here will solve the issue in just as hard-coded a way, and will be no more portable than actually specifying ther version. Future versions will not likely have numbers like `post1`, as past versions also didnot. I think removing the hard-coded version here would do nothing more but solve the issue in a needlessly ellaborate way.

dev-python/nilearn/nilearn-0.1.4.1.ebuild
TheChymera: @marbre are you sure it is not?

TheChymera:
@marbre I added the use flag - there appears to be no test suite shipped with the software.

marbre:
Please check upstreams .travis.yml.

dev-python/nilearn/nilearn-0.1.4.1.ebuild
marbre: AFAIK it is not, at least I know nothing about it. https://www.gentoo.org/support/news-items/2015-07-25-python-targets.html only tells that 3.4 is gentoos new Python 3 default.

TheChymera: @jlec didn't you tell me to stop adding python3_3 support?

jlec: We should start fading it out in favor of python3_5 support.

TheChymera: upstream does not support 3_5 yet. In the mean time, do I add 3_3 or not?

jlec: as you like.

TheChymera: I'll leave it out then.

TheChymera:
[This](https://github.com/nilearn/nilearn/blob/master/.travis.yml) is the upstram `.travis.yml` I don't see where the actual script is. What can I do?

Does [this](https://github.com/nilearn/nilearn/blob/master/.travis.yml#L43) mean it's some generic nosetests?

marbre:
See https://github.com/nilearn/nilearn/blob/master/.travis.yml#L50.

Thus, you might what is call in the Makefile for the`test-code` target,.

TheChymera:
so within an ebuild I use emake for that? e.g. `emake Makefile test-code` ?

kiwifb:
`emake test-code`, `emake` is smart enough to find makefiles normally.

dev-python/nilearn/nilearn-0.1.4.1.ebuild
```diff
@@ -0,0 +1,35 @@
...
+	dev-python/numpy[${PYTHON_USEDEP}]
+	sci-libs/scikits_learn[${PYTHON_USEDEP}]
+	sci-libs/scipy[${PYTHON_USEDEP}]
+	sci-libs/nibabel[${PYTHON_USEDEP}]
+	plot? ( dev-python/matplotlib[${PYTHON_USEDEP}] )"
+
+S="${WORKDIR}/${PN}-${MY_PV}"
+
+src_test() {
+	emake test-code
```
jlec: Missing test dependency.

TheChymera: you mean nosetests? should I use a USE flag for tests? or is this somehow built-in?

jlec: USE=test

TheChymera:
@jlec can we have this now? [nistats](https://github.com/gentoo-science/sci/pull/553) depends on it.

PR #523, opened 2015-10-25
sci-libs/symengine: split out bindings to dev-python/symengine

mschubert:
Upstream split `symengine` into the [C++ library](https://github.com/symengine/symengine), and [external python bindings](https://github.com/symengine/symengine.py).

This ebuild uses a separate package for python support in the development version, that is now compatible with more than one python installation.

(Note: Is is preferable to include the python binding sources in the original ebuild and have a `USE` flag handle its installation instead of a separate package?)

### DIFF
--- a/dev-python/symengine/symengine-9999.ebuild
+++ b/dev-python/symengine/symengine-9999.ebuild
@@ -1,0 +1,32 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
+
+EAPI=5
+
+PYTHON_COMPAT=( python{2_7,3_3,3_4,3_5} )
+
+inherit distutils-r1 git-r3
+
+DESCRIPTION="Python wrappers to the symengine C++ library"
+HOMEPAGE="https://github.com/sympy/symengine.py"
+SRC_URI=""
+EGIT_REPO_URI="https://github.com/symengine/symengine.py.git"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
+
+RDEPEND="
+	dev-python/numpy[${PYTHON_USEDEP}]
+	=sci-libs/symengine-9999"
+DEPEND="${RDEPEND}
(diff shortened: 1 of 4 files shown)


dev-python/symengine/symengine-9999.ebuild
```diff
@@ -0,0 +1,33 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
+
+EAPI=5
+
+PYTHON_COMPAT=( python{2_7,3_3,3_4,3_5} )
+
+inherit git-r3 distutils-r1
```
jlec: If possible, sort alphabetically.

dev-python/symengine/symengine-9999.ebuild
```diff
@@ -0,0 +1,33 @@
...
+
+DESCRIPTION="Python wrappers to the symengine C++ library"
+HOMEPAGE="https://github.com/sympy/symengine.py"
+SRC_URI=""
+EGIT_REPO_URI="https://github.com/symengine/symengine.py.git"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
```
jlec: This is automatically set by the distutils-r1.eclass. Only python-r1 and python-single need it.

dev-python/symengine/symengine-9999.ebuild
```diff
@@ -0,0 +1,33 @@
...
+	=sci-libs/symengine-9999"
+DEPEND="${RDEPEND}
+	dev-python/cython[${PYTHON_USEDEP}]
+	dev-python/setuptools[${PYTHON_USEDEP}]"
+
+# if this is not set python2 .so is linked to python3
+DISTUTILS_IN_SOURCE_BUILD=1
+
+python_install_all() {
+	newdoc README.md ${PN}_py.md.bz2
```
jlec: Don't add the bz2 suffix. Otherwise you end up with e.g. ```${PN}_py.md.bz2.xz```

mschubert:
All fixed.

PR #528, opened 2015-11-08
sci-misc/kaldi: Version bump, switch to OpenBLAS math library by default, add CUDA support

akreal:


sci-misc/kaldi/files/default_rules.mk.a7d9824.patch
```diff
@@ -0,0 +1,48 @@
+*** makefiles/default_rules.mk.orig	2015-10-28 22:11:07.903182744 -0400
+--- makefiles/default_rules.mk	2015-10-29 20:43:41.081851545 -0400
```
jlec: Could you please create a unified diff?

sci-misc/kaldi/kaldi-0_p20151106.ebuild
```diff
@@ -0,0 +1,91 @@
...
+	!atlas? ( sci-libs/openblas[-openmp,-threads] sci-libs/lapack-reference )
+	cuda? ( dev-util/nvidia-cuda-toolkit )"
+DEPEND="${RDEPEND}
+	doc? ( app-doc/doxygen )"
+
+REQUIRED_USE="!atlas? ( !threads )"
+
+# We need write acccess /dev/nvidiactl, /dev/nvidia0 and /dev/nvidia-uvm and the portage
+# user is (usually) not in the video group
+RESTRICT="cuda? ( userpriv )"
```
jlec: Is this really working?

sci-misc/kaldi/kaldi-0_p20151106.ebuild
```diff
@@ -0,0 +1,91 @@
...
+		addwrite /dev/nvidia0
+		addwrite /dev/nvidia-uvm
+	fi
+	default
+}
+
+src_install() {
+	dolib.so $(make print-libfiles)
+	dobin $(make print-binfiles)
+	use doc && dohtml -r html/*
```
jlec: better use 
```ebuild
use doc && dodoc -r html
```
dohtml is deprecated.

jlec:
Is it possible to simply use what the user selects with eselect blas?

sci-misc/kaldi/files/default_rules.mk.a7d9824.patch
akreal: Created.

sci-misc/kaldi/kaldi-0_p20151106.ebuild
akreal: Yes, I've checked it: when `cuda` USE flag is enabled, sandbox is running as root and tests are passing. This is actually needed only when both `cuda` and `test` are enabled, I've updated ebuild.

sci-misc/kaldi/kaldi-0_p20151106.ebuild
akreal: Done.

akreal:
I don't think blas from eselect can be used easily. Upstream configuration adds different compilation flags for different libraries (e.g. `-DHAVE_OPENBLAS`) and different code is compiled depending on them (using `ifdef`s). Because of this ebuild needs to know which library is selected to generate proper compilation flags.

kiwifb:
That's not very good in general but you could do some detection yourself in the ebuild.
```
fbissey@QCD-nzi3 ~ $ eselect blas show
openblas-threads
```
for example. So you can figure out which `{c,}blas` you have from the ebuild. A bit more complicated but doable.

akreal:
Thanks. I will try to use `toolchain-funcs` like in `sci-libs/plasma`.

akreal:
Turned out `toolchain-funcs` is not what I thought, and I need `eselect` as well.
@jlec could you please review this again?

sci-misc/kaldi/kaldi-0_p20151106.ebuild
```diff
@@ -0,0 +1,119 @@
...
+
+	if [[ ${cblas_provider} =~ "atlas" ]]; then
+		append-cxxflags -DHAVE_ATLAS
+	elif [[ ${cblas_provider} =~ "mkl" ]]; then
+		append-cxxflags -DHAVE_MKL
+	elif [[ ${cblas_provider} =~ "openblas" ]]; then
+		append-cxxflags -DHAVE_OPENBLAS $($(tc-getPKG_CONFIG) --cflags lapacke)
+	else
+		die "Build with ${cblas_provider} CBLAS is not supported"
+	fi
```
jlec: it would be nice to phave thsi check in pkg_pretend so the emerge fails a t the start not so late.

sci-misc/kaldi/kaldi-0_p20151106.ebuild
```diff
@@ -0,0 +1,119 @@
...
+	EOF
+
+	if use cuda; then
+		cat <<-EOF >> kaldi.mk
+			CUDA = true
+			CUDATKDIR = "${EPREFIX}"/opt/cuda
+		EOF
+		cat makefiles/linux_x86_64_cuda.mk >> kaldi.mk
+		sed -i \
+			-e "s:CUDA_FLAGS = -g:CUDA_FLAGS = -O2:" \
```
jlec: You need to use the cuda.eclass (https://devmanual.gentoo.org/eclass-reference/cuda.eclass/index.html) and use NVCCFLAGS

sci-misc/kaldi/kaldi-0_p20151106.ebuild
akreal: Done.

sci-misc/kaldi/kaldi-0_p20151106.ebuild
akreal: Done.

Issue #530, opened 2015-11-10
python-future package


TheChymera:
It seems I will need [this package](https://pypi.python.org/pypi/future) as a nipype dependency in the near future. See upstream discussion:

https://github.com/nipy/nipype/issues/1271#issuecomment-155260437

any idea if it is already in gentoo? I tried these packages:

```
* dev-python/futures
     Available versions:  2.2.0 (~)3.0.2 (~)3.0.3 {doc PYTHON_TARGETS="pypy python2_7"}
     Homepage:            https://code.google.com/p/pythonfutures https://pypi.python.org/pypi/futures
     Description:         Backport of the concurrent.futures package from Python 3.2

* virtual/python-futures
     Available versions:  0 {PYTHON_TARGETS="pypy python2_7 python3_3 python3_4"}
     Description:         A virtual for the Python concurrent.futures module
```

But none of them lets me `from future import *` - whic is apparently what the nipype setup.py wants to see.

kiwifb:
Doesn't look like it. I think they mean `http://python-future.org/`

jlec:
no it isn't. the dev-python/futures packages are backports. future is a glue layer like six.

TheChymera:
So I should package that if I want newer nipype versions to work, right?

jlec:
better that

PR #531, opened 2015-11-11
dev-python/future: new package (nipype dep)

TheChymera:
Solves #530


dev-python/future/future-0.15.2.ebuild
```diff
@@ -0,0 +1,22 @@
...
+HOMEPAGE="http://python-future.org/"
+SRC_URI="mirror://pypi/f/${PN}/${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE=""
+
+python_test() {
+	python setup.py test
```
kiwifb: Do not call `python` directly. Better use `esetup.py test` at least.

TheChymera:
@kiwifb  this still seems to fail - any idea why? I think it's actually because of something in a `x11-libs/xview-bin` ebuild, right?

kiwifb:
I don't think that's `xview-bin` at fault and I also don't think it is your fault. Some ebuilds have a broken dependency tree with regards to some keywords.
On your local machine does 
```
ebuild future-0.15.2.ebuild test
```
finish properly? Then what does `repoman -v full` says?

PR #538, opened 2015-11-16
fix dependency on boost[python]

zasdfgbnm:
rdkit should requires that boost is installed with python support.


sci-libs/rdkit/rdkit-2015.03.1.ebuild
```diff
@@ -18,7 +18,7 @@ KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
 IUSE="+python -static-libs"
 REQUIRED_USE="python? ( ${PYTHON_REQUIRED_USE} )"
 
-RDEPEND="dev-libs/boost
+RDEPEND="dev-libs/boost[python]
 	python? (
 		dev-libs/boost[${PYTHON_USEDEP}]
```
marbre: Why not `dev-libs/boost[python,${PYTHON_USEDEP}]`?

zasdfgbnm:
I have fixed... If the rdkit don't have python USE flag, we don't need to require boost built with python support.

PR #544, opened 2015-12-02
Int64 pkgconfig

SoapZaNet:
This PR fixes broken pkg-config files. For instance:

```
$ pkg-config --libs lapack
-L -lreflapack -lrefblas
```
Now:
```
$ pkg-config --libs lapack
-lreflapack -lrefblas
```


sci-libs/blas-reference/blas-reference-20151113.ebuild
```diff
@@ -8,7 +8,7 @@ PYTHON_COMPAT=( python2_7 )
 
 NUMERIC_MODULE_NAME="refblas"
 
-inherit alternatives-2 cmake-utils fortran-2 numeric-int64-multibuild python-any-r1 toolchain-funcs
+inherit alternatives-2 cmake-utils fortran-2 numeric-int64-multibuild python-any-r1 toolchain-funcs eutils
```
jlec: nitpicking thing, but could you please sort this alphabetically?

jlec:
we need to revbump all ebuilds so that we trigger a rebuild on all systems. You can do a simple git mv.

PR #553, opened 2015-12-17
sci-biology/nistats: added new package

TheChymera:
New package (will possibly replace sci-biology/nipy).

### DIFF
--- a/sci-biology/nistats/nistats-9999.ebuild
+++ b/sci-biology/nistats/nistats-9999.ebuild
@@ -18,10 +18,10 @@ SLOT="0"
 IUSE=""
 KEYWORDS=""
 
-DEPEND="
-	dev-python/setuptools[${PYTHON_USEDEP}]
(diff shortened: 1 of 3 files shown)


sci-biology/nistats/nistats-9999.ebuild
```diff
@@ -0,0 +1,30 @@
...
+
+LICENSE="BSD"
+SLOT="0"
+IUSE=""
+KEYWORDS=""
+
+DEPEND="
+	dev-python/setuptools[${PYTHON_USEDEP}]
+	dev-python/numpy[${PYTHON_USEDEP}]"
+RDEPEND="${DEPEND}
```
jlec: Do you really need setuptools at runtime?

TheChymera: I thought it would be shorter than making a separate COMMONDEPEND variable just to hold numpy.

PR #554, opened 2015-12-17
Nilearn directory updae

TheChymera:
sci-biology is certainly more appropriate for this than dev-python. Sorry for the timing, @jlec.


jlec:
Then you need to add an entry into profiles/updates.

TheChymera:
@jlec good?

PR #558, opened 2015-12-31
Nexus 4.4.1

chatcannon:
This version is built with CMake so I have had to make quite a few changes to the ebuild

Python support is now in a separate project, called Nexpy


sci-libs/nexus/nexus-4.4.1.ebuild
```diff
@@ -7,37 +7,37 @@ EAPI=5
 FORTRAN_NEEDED=fortran
 FORTRAN_STANDARD=90
 PYTHON_COMPAT=( python2_7 )
```
jlec: You can drop this if no python is needed anymore.

sci-libs/nexus/nexus-4.4.1.ebuild
```diff
@@ -7,37 +7,37 @@ EAPI=5
...
-inherit fortran-2 flag-o-matic python-single-r1
+inherit cmake-utils fortran-2 java-pkg-opt-2
 
 DESCRIPTION="Data format for neutron and x-ray scattering data"
 HOMEPAGE="http://nexusformat.org/"
-SRC_URI="http://download.nexusformat.org/kits/${PV}/${P}.tar.gz"
+SRC_URI="https://github.com/nexusformat/code/archive/v${PV}.tar.gz"
```
jlec: You must rename the source tarball to avoid name clashes:
```ebuild
SRC_URI="https://github.com/nexusformat/code/archive/v${PV}.tar.gz -> ${P}.tar.gz"
```

chatcannon:
I have rebased the branch on the latest master. Is there anything else that needs to be done before it can be merged?

PR #564, opened 2016-01-11
Added clfft-2.8-r1.

StinGer-ShoGuN:
I tried to build clfft with the `client` USE flag and install phase kept on failing because client was not build. After trying to build it from source, I found that `static-libs` from boost were required, hence this new revision proposal. The only change is the boost line in the DEPEND definition.

Tested on my Core i7.


jlec:
That won't work. Please try to fix the build system to respect shared libs. Using static-libs is bad for several reasons including security. So we cannot use that as a general solution.

StinGer-ShoGuN:
I agree (plus I hate when programs load several time the same code in memory).
Did not know how you would like to handle this. Will issue another pull request then.

PR #569, opened 2016-01-18
sci-biology/psychopy version bump

TheChymera:
1.83.03


sci-biology/psychopy/psychopy-1.83.03.ebuild
```diff
@@ -0,0 +1,47 @@
...
+
+DESCRIPTION="Python experimental psychology toolkit"
+HOMEPAGE="http://www.psychopy.org/"
+SRC_URI=""
+EGIT_REPO_URI="https://github.com/psychopy/psychopy.git"
+EGIT_COMMIT="1.83.03"
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
```
jlec: No KEYWORDS in live ebuilds.

sci-biology/psychopy/psychopy-1.83.03.ebuild
```diff
@@ -0,0 +1,47 @@
...
+
+DEPEND="
+	dev-python/setuptools[${PYTHON_USEDEP}]
+	"
+RDEPEND="${DEPEND}
```
jlec: Is setuptools needed during runtime?

sci-biology/psychopy/psychopy-1.83.03.ebuild
TheChymera: This is not live, it's already versioned.

jlec: live means, you cannot install it without internet connection. In fact, you also need to be able to reach Github. Either do a proper snapshot, pull a snapshot from a tag or drop KEYWORDS.

TheChymera: How do I pull a snapshot from a tag? Apparently there is an archive with the release at this address: https://github.com/psychopy/psychopy/releases/download/1.83.03/PsychoPy-1.83.03.zip Can I link to that with egit? or should I use SRC_URI?

jlec: This is an zip file, so it should be in SRC_URI.

TheChymera: @jlec will that count as non-live?

jlec: sure.

TheChymera: @jlec done

PR #573, opened 2016-01-23
sci-biology/kallisto: version bump 0.42.4

mschubert:


sci-biology/kallisto/kallisto-0.42.4.ebuild
```diff
@@ -2,7 +2,7 @@
 # Distributed under the terms of the GNU General Public License v2
 # $Id$
 
-EAPI=5
+EAPI=6
```
marbre: The `cmake-utils` eclass does not yet support `EAPI=6`. Patches are currently on the way, but that will be the reason for travis failing.

jlec: It does now, so everything should be fine.

jlec:
merged

PR #575, opened 2016-01-23
sci-biology/sailfish: add ebuild

mschubert:


sci-biology/sailfish/files/sailfish-0.9.1-no-boost-static.patch
```diff
@@ -0,0 +1,20 @@
+Do not force static and multithreaded boost libs, use what is available
+
+Patch by Michael Schubert
+
+--- CMakeLists.txt
++++ CMakeLists.txt
```
kiwifb: It would be a good idea for future maintainability if those patch were `git` style
```
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
```
It doesn't have to be `a` and `b` but in the future with EAPI 6 and the `eapply` helper which is replacing `epatch`, `patch` will only be called with `-p1`, i.e. with a leading directory being stripped. The current `epatch` function tries everything between 0 and 10 (I think).

sci-biology/sailfish/sailfish-0.9.1.ebuild
```diff
@@ -0,0 +1,28 @@
+# Copyright 1999-2015 Gentoo Foundation
```
kiwifb: 2016...

mschubert: Thanks. It's enough if you tell me once, then I'll fix it in all ebuilds.

sci-biology/sailfish/sailfish-9999.ebuild
```diff
@@ -0,0 +1,29 @@
...
+inherit cmake-utils git-r3
+
+DESCRIPTION="Rapid Mapping-based Isoform Quantification from RNA-Seq Reads"
+HOMEPAGE="http://www.cs.cmu.edu/~ckingsf/software/sailfish/"
+SRC_URI=""
+EGIT_REPO_URI="https://github.com/kingsfordgroup/sailfish.git"
+
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
```
marbre: Live ebuild don't have keywords.

sci-biology/sailfish/sailfish-0.9.1.ebuild
```diff
@@ -0,0 +1,28 @@
...
+PATCHES=( "${FILESDIR}"/${PN}-0.9.1-no-boost-static.patch )
+
+DEPEND="dev-libs/boost:0
+		dev-libs/jemalloc
+		dev-cpp/tbb"
+RDEPEND="${DEPEND}"
+
+src_install() {
+	cmake-utils_src_install
+	rm -rf "${ED}"/usr/tests || die
```
marbre: Due to `-f` there is no need to led rm die as is always returns true.

sci-biology/sailfish/sailfish-9999.ebuild
```diff
@@ -0,0 +1,29 @@
...
+PATCHES=( "${FILESDIR}"/${PN}-0.9.1-no-boost-static.patch )
+
+DEPEND="dev-libs/boost:0
+		dev-libs/jemalloc
+		dev-cpp/tbb"
+RDEPEND="${DEPEND}"
+
+src_install() {
+	cmake-utils_src_install
+	rm -rf "${ED}"/usr/tests || die
```
marbre: Dito.

sci-biology/sailfish/sailfish-0.9.1.ebuild
```diff
@@ -0,0 +1,28 @@
+# Copyright 1999-2016 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
+
+EAPI=5
```
jlec: EAPI=6.

sci-biology/sailfish/sailfish-9999.ebuild
```diff
@@ -0,0 +1,29 @@
+# Copyright 1999-2016 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
+
+EAPI=5
```
jlec: as above.

sci-biology/sailfish/sailfish-0.9.1.ebuild
mschubert: as of today, `portage-2.2.25` is removed from prefix again, so I've got no way of checking `EAPI=6` ebuilds with `repoman`

jlec: Then go for EAPI=5 as long you have to.

mschubert:
I don't understand the Travis-CI error. `repoman full` does not complain on my system.

kiwifb:
There seem to be a network error, but beside that I believe we are in the process of removing the herds https://blogs.gentoo.org/mgorny/2016/01/16/glep67-or-how-packages-are-going-to-be-maintained/ so it would be better if you removed that field. I believe that's the straw that breaks the camel's back so to speak.

mschubert:
Ok, then this boils down to the same issue as `EAPI=6` for prefix.

I'd prefer to make these adjustments after they are supported by `repoman`, otherwise I can only estimate the changes according to the docs but not run any local validation.

kiwifb:
In that case it is not down to `repoman` but of the version of  `metadata.dtd` currently installed and found by `repoman` on your system. That's the `xml` schema for the `metadata.xml` file. In the one you have locally `herd` is a valid field, in the one travis uses, it is not. If `repoman` downloaded a new version of the schema it may have trouble finding it afterwards because of some portage bugs on prefix. prefix is in a bit of a mess at the moment (I mean more than usual).

kiwifb:
So you don't feel alone, I just tried to update a `metadata.xml` file in the sage-on-gentoo overlay for `dev-libs/ntl` for which I currently have a newer version that is in the tree. Since we were talking about it I checked the `metadata.xml` and thought I should update it. So I copied `metadata.xml` from the main tree to my overlay. `git add metadata.xml` and then `repoman -v commit`:
```
[INFO] checking package dev-libs/ntl
>>> Creating Manifest for /home/fbissey/Work/Overlays/sage-on-gentoo/dev-libs/ntl
!!! metadata.xml is invalid:
!!! I/O error : Attempt to load network entity http://www.gentoo.org/dtd/metadata.dtd
!!! /home/fbissey/Work/Overlays/sage-on-gentoo/dev-libs/ntl/metadata.xml:2: warning: failed to load external entity "http://www.gentoo.org/dtd/metadata.dtd"
!!! <!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
!!!                                                                       ^
!!! /home/fbissey/Work/Overlays/sage-on-gentoo/dev-libs/ntl/metadata.xml:4: element maintainer: validity error : No declaration for attribute type of element maintainer
!!! /home/fbissey/Work/Overlays/sage-on-gentoo/dev-libs/ntl/metadata.xml:8: element maintainer: validity error : No declaration for attribute type of element maintainer
!!! Document /home/fbissey/Work/Overlays/sage-on-gentoo/dev-libs/ntl/metadata.xml does not validate against /scratch/distfiles/metadata.dtd
  metadata.bad [fatal]          1
   dev-libs/ntl/metadata.xml
```
So, currently the master `metadata.dtd` is off line :( and the main tree `metadata.xml` file is invalid when checked against the latest copy of `metadata.dtd` that I have on disc. This is not on prefix.

sci-biology/sailfish/metadata.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+  <herd>sci-biology</herd>
```
jlec: You need to drop herd.

sci-biology/sailfish/metadata.xml
```diff
@@ -0,0 +1,12 @@
...
+  <maintainer>
```
jlec: this needs a type attribute. set it to ```person```

jlec:
merged

PR #581, opened 2016-02-04
dev-python/pydotplus: new package

TheChymera:
Will be required a s a dep by [prov](https://pypi.python.org/pypi/prov) which is [required as a dep for newer nipype versions](https://github.com/nipy/nipype/issues/1336).


dev-python/pydotplus/pydotplus-2.0.2.ebuild
```diff
@@ -0,0 +1,31 @@
...
+	dev-python/setuptools[${PYTHON_USEDEP}]
+	"
+RDEPEND="
+	dev-python/pyparsing[${PYTHON_USEDEP}]
+	"
+
+S="${WORKDIR}/${PN}-${MY_PV}"
+
+src_test() {
+	${PYTHON} -m unittest discover
```
jlec: MIssing ```die```.

dev-python/pydotplus/pydotplus-2.0.2.ebuild
```diff
@@ -0,0 +1,31 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
+
+EAPI=5
+
+PYTHON_COMPAT=( python2_7 python3_4 )
```
jlec: How about adding python 3.3 and 3.5 as well?

TheChymera: it's only listed under python 2.7 and 3.4 [on PyPI](https://pypi.python.org/pypi/pydotplus).

jlec: sure, but upstreams are listing wildest things there. please test with 3.3 and 3.5 as well.

TheChymera: I don't use python 3.* and I don't even have python 3.3 - how would I test them without having to switch?

jlec: If you cannot test then leave it out.

TheChymera:
@jlec?

PR #588, opened 2016-02-16
Intel eclass QA updates

SoapGentoo:
@jlec @mgorny
Major overhaul from the RFC
Points to consider:
- @jlec the `convert2intel` function: why do we need it? it is never actually used.
- `_isdp_big-warning` still lacks a better description. Isn't it obvious from context?
- `eqawarn` remains. Shall we remove the `SKIP_LICENSE` functionality?


jlec:
>    @jlec the convert2intel function: why do we need it? it is never actually used.

I planned to use it here https://github.com/gentoo-science/sci/blob/intel-eclass-qa/eclass/intel-sdp-r1.eclass#L397 to replace some of the INTEL_ARCH logic. Is it worth to use it there? 

>    _isdp_big-warning still lacks a better description. Isn't it obvious from context?

I would think so, plus it is an internal function

>    eqawarn remains. Shall we remove the SKIP_LICENSE functionality?

I think you can use MKL without a license. But we should rename it to something intel specific.

eclass/intel-sdp-r1.eclass
```diff
@@ -330,8 +353,8 @@ intel-sdp-r1_pkg_pretend() {
 	: ${CHECKREQS_DISK_BUILD:=256M}
 	check-reqs_pkg_pretend
 
-	if [[ -z ${INTEL_SKIP_LICENSE} ]]; then
-		if echo ${INTEL_LICENSE_FILE} | grep -q @; then
+	if [[ -z "${INTEL_SKIP_LICENSE}" ]]; then
+		if [[ "${INTEL_LICENSE_FILE}" == *@* ]]; then
```
jlec: I think you can drop the quoting here.

eclass/intel-sdp-r1.eclass
```diff
@@ -367,122 +389,93 @@ intel-sdp-r1_pkg_pretend() {
...
 		done
 
+		local debug_list
 		debug_list="$(IFS=$'\n'; echo ${list[@]} )"
 
 		debug-print "Adding to decompression list:"
 		debug-print ${debug_list}
 
-		tar xvf "${DISTDIR}"/${t} ${list[@]} &> "${T}"/rpm-extraction.log
+		tar -xf "${DISTDIR}"/${t} ${list[@]}
```
jlec: Please readd the verbose logging. That makes debugging so much easier.

eclass/intel-sdp-r1.eclass
```diff
@@ -494,12 +487,14 @@ intel-sdp-r1_src_install() {
...
-	if path_exists opt/intel/"${INTEL_DPN}"*/uninstall; then
+	if path_exists opt/intel/"${INTEL_DIST_NAME}"*/uninstall; then
 		ebegin "Cleaning out uninstall"
-		rm -r opt/intel/"${INTEL_DPN}"*/uninstall || die
+		rm -r opt/intel/"${INTEL_DIST_NAME}"*/uninstall || die
 		eend
 	fi
 
+	echo $INTEL_SDP_DIR
+
```
jlec: Why?

jlec:
@SoapGentoo nice work.

eclass/intel-sdp-r1.eclass
```diff
@@ -95,19 +104,19 @@ esac
 # specify the full path
 #
 # e.g. CLI_install/rpm/intel-vtune-amplifier-xe-cli
```
jlec: Missing array ```()```

eclass/intel-sdp-r1.eclass
```diff
@@ -9,62 +9,71 @@
...
 # e.g. 2016_update1
 #
-# Must be defined before inheriting the eclass
+# Must be defined before inheriting the eclass.
 
 # @ECLASS-VARIABLE: INTEL_TARX
 # @DESCRIPTION:
-# The package extention.
-# To find out its value, see the links to download in
+# The package suffix.
```
mgorny: i suggest to put 'distfile' or sth like that there.

eclass/intel-sdp-r1.eclass
```diff
@@ -94,63 +103,81 @@ esac
...
 #
 # if the rpm is located in a directory different to INTEL_RPMS_DIR you can
 # specify the full path
 #
-# e.g. CLI_install/rpm/intel-vtune-amplifier-xe-cli-common
-: ${INTEL_DAT_RPMS:=()}
+# e.g. ("CLI_install/rpm/intel-vtune-amplifier-xe-cli-common")
+[[ ${INTEL_DAT_RPMS[@]} ]] || INTEL_DAT_RPMS=()
 
 # @ECLASS-VARIABLE: INTEL_SINGLE_ARCH
```
mgorny: Maybe a little DIST here too, since it applies to distfiles and not package features.

eclass/intel-sdp-r1.eclass
```diff
@@ -94,63 +103,81 @@ esac
...
+# @DEFAULT_UNSET
 # @DESCRIPTION:
-# Unset, if only the multilib package will be provided by intel
-: ${INTEL_SINGLE_ARCH:=true}
+# Set to "true" if arches are to be fetched separately, instead of using the combined tarball.
 
-MULTILIB_COMPAT=( abi_x86_{32,64} )
+# @FUNCTION: _isdp_set-version-vars
```
mgorny: Errrrrr, why don't you just use a single big function to set all global variables? You can set global vars from within a function, you know.

eclass/intel-sdp-r1.eclass
```diff
@@ -94,63 +103,81 @@ esac
...
+# @INTERNAL
+# @DESCRIPTION:
+# Sets the internal Intel version variables. This should be used temporarily
+# and unset using _isdp_unset-version-vars when not needed anymore.
+_isdp_set-version-vars() {
+	local i
+	for i in {1..4}; do
+		declare -g _INTEL_PV${i}=$(get_version_component_range ${i})
```
mgorny: Any reason not to use an array here, and `$(get_version_components)` like I s uggested?

eclass/intel-sdp-r1.eclass
```diff
@@ -94,63 +103,81 @@ esac
...
+	done
 
-inherit check-reqs eutils multilib-build versionator
+	local prefixes=('' '-' '.' '.' '-')
```
mgorny: This is not really what I meant. I was wondering if there's a way to make this simpler, not unreadable ;-P.

eclass/intel-sdp-r1.eclass
```diff
@@ -200,18 +225,16 @@ _isdp_big-warning() {
...
 
 		test-failed )
-			echo ""
 			ewarn "Function test failed. Most probably due to an invalid license."
 			ewarn "This means you already tried to bypass the license check once."
 			;;
 	esac
 
-	echo ""
+	ewarn ""
```
mgorny: No need for "". Especially with ryao's idea about empty lines.

eclass/intel-sdp-r1.eclass
```diff
@@ -284,10 +304,11 @@ _isdp_version_test() {
 _isdp_run-test() {
 	debug-print-function ${FUNCNAME} "${@}"
 
-	if [[ -z ${INTEL_SKIP_LICENSE} ]]; then
+	if [[ -z "${INTEL_SKIP_LICENSE}" ]]; then
```
mgorny: No need for quoting. Not that I haveanything against it, just wondering why you added it.

eclass/intel-sdp-r1.eclass
```diff
@@ -339,20 +362,19 @@ intel-sdp-r1_pkg_pretend() {
...
 			ebegin "Checking for a license in: ${dir}"
-			#maybe use nullglob or [[ $(echo ${dir/*lic) != "${dir}/*lic" ]]
-			[[ $( ls "${dir}"/*lic 2>/dev/null ) ]]; ret=$?
-			eend ${ret}
-			if [[ ${ret} == "0" ]]; then
-				warn=${ret}
+			path_exists "${dir}"/*lic && warn=0
+			eend ${warn}
+
+			if [[ ${warn} == 0 ]]; then
```
mgorny: `eend ${warn} && break` would work too, I thikn.

eclass/intel-sdp-r1.eclass
```diff
@@ -94,68 +103,84 @@ esac
...
+# e.g. ("CLI_install/rpm/intel-vtune-amplifier-xe-cli-common")
+[[ ${INTEL_DIST_DAT_RPMS[@]} ]] || INTEL_DIST_DAT_RPMS=()
 
 # @ECLASS-VARIABLE: INTEL_SINGLE_ARCH
 # @DESCRIPTION:
-# Unset, if only the multilib package will be provided by intel
-: ${INTEL_SINGLE_ARCH:=true}
+# Set to "true" if arches are to be fetched separately, instead of using
+# the combined tarball.
+: ${INTEL_SINGLE_ARCH:=false}
```
jlec: Perhaps this could be renamed to something like ```INTEL_SPLIT_ARCH_DIST``` or alike to be more verbose and explanatory on what it is.

PR #608, opened 2016-04-03
sci-biology/abyss: fix automagic haskell deps, bug #534412

trofi:
It's a ::gentoo change a50e1769e82e28dedb16083ac9136f17f9af2297

Reported-by: Toralf Förster
Bug: https://bugs.gentoo.org/534412


SoapGentoo:
Merged

Issue #610, opened 2016-04-06
What do we do about $FSLDIR ?


TheChymera:
`sci-biology/fsl` has a variable called `$FSLDIR`. This is meant to inform FSL functions on where FSL is installed, and FSL functions inside FSL are (afaict) always called like `$FSLDIR/bin/FUNCTIONNAME`.

Apparently we set `FSLDIR=/usr` in `/etc/env.d/99fsl` - but I am thinking handling this a bit differently might help us better deal with issues such as executable name collisions (as suggested [here](https://github.com/gentoo-science/sci/issues/606) we could just change the name, but there are packages soutside of FSL proper which also need to know how to call these renamed functions - see a discussion about that [here](https://github.com/nipy/nipype/issues/1415)).

I am wondering whether it is reasonable to put stuff under `/usr/local/fsl` - and just add a few functions (for which we know there are no conflicts) under `/usr/bin`.

Any ideas on this, @kiwifb ?

TheChymera:
Apparently NeuroDebian installs FSL bianries in `/usr/share/fsl/bin/` (and sets `$FSLDIR` to `/usr/share/fsl/`) - and then creates symlinks to these files in /usr/bin prepending `${P}_` to the name.

TheChymera:
Is this what would do that in portage:

```
src_install() {
    for file in bin/*
    do
        dosym "${ED}"/usr/bin/"${P}"_file file
    done
}
?

kiwifb:
I usually go for `/usr/lib{,64}/fsl/`. I don't think your procedure would work.

TheChymera:
I don't understand what you mean - what do you do with `/usr/lib{,64}/fsl/` ? do you set FSLDIR to that? It seems to me our current installation is installing FSL files here `insinto /usr/share/${PN}`

What procedure exactly do you think would not work?

kiwifb:
`/usr/share/${PN}` is fine, especially if we already do it. I am on a conference call at the same time but your `dosym` stuff looks wrong.

TheChymera:
As of https://github.com/gentoo-science/sci/pull/716 we are addressing the issue of FSLDIR by:

* Setting it to `/usr/share/fsl`
* Placing whatever files belong there in there (currently `data/`)
* Placing all other files in the appropriate filesystem locations
* Adding symlinks under `/usr/share/fsl` to allow FSL and dependencies to find FSL files.

Issue #606, opened 2016-03-30
FSL and graphviz cluster function


TheChymera:
It appears both `media-gfx/graphviz` and `sci-biology/fsl` try to install `/usr/bin/cluster`. Up until now I [removed the file from FSL](https://github.com/gentoo-science/sci/blob/6d2fe5739e0bf0478db40c4429215e78110b3326/sci-biology/fsl/fsl-5.0.8.ebuild#L131) as I thought it was the same thing coming bundled (since graphviz is an FSL dependency).

This turns out to not be the case. Unfortunately both packages decided to provide a binary with a rather collison-prone name. What's the standard/recommended/most elegant way to deal with this?

@kiwifb

jlec:
The best is to rename ```cluster``` for FSL to e.g. ```cluster-fsl```.

kiwifb:
I agree with jlec. But you may want to also check `fsl` scripts for calls to `cluster` and adjust them on the assumption that they want `cluster-fsl`.

TheChymera:
I am actually thinking `fsl_cluster` would be a better name.

kiwifb:
No objections to the name. I wouldn't think there are strict rules.

TheChymera:
How can I actually rename the file before portage tries to put it into `/usr/bin` ?

I tried replacing [my previous fix](https://github.com/gentoo-science/sci/blob/6d2fe5739e0bf0478db40c4429215e78110b3326/sci-biology/fsl/fsl-5.0.8.ebuild#L131) with

`newbin "${ED}"/usr/bin/cluster fsl_cluster || die`

But now the build fails due to the collision.

jlec:
> newbin "${ED}"/usr/bin/cluster fsl_cluster || die

That doesn't work as you expect. The resulting installation would have both files.

```ebuild
mv "${ED}"/usr/bin/{,fsl_}cluster || die
```

is what you are looking for.

TheChymera:
Ok, so apparently this isnn't done yet. Upstream told me that at least main `featreg`, `slices_summary`, `easythresh`, and `autoaq` call cluster, and would have to be informed on the change - but that I also should try grepping for cluster in the entire source to find other packges which maybe also do that.

it trns out the name "cluster" is used all over the place in the fsl surces. Do you know how I can try to single out the places where this is in fact a function call?

kiwifb:
I actually mentioned that you should do that. `grep -r cluster *` is your friend/enemy. You'll have to sort the hits.

[...]

PR #612, opened 2016-04-08
Better fix for FSL executable naming conflicts

TheChymera:
This is a better solution to issue #606 (and any future naming conflicts that might turn up for FSL) and also addresses the concerns of how exactly to deal with `FSLDIR` discussed in #610.

* FSL functions are installed in `FSLDIR=/usr/share/fsl`
* Symlinks are created, as `/usr/bin/fsl_*`


TheChymera:
@jlec @kiwifb I tested this on a number of my machines, and it seems to work nicely. Can you pull it please?

kiwifb:
I can't I don't have access anymore to the official repo, only github. So someone would have to clean after me in any case.

TheChymera:
Why did you lose your access? I always thought this was the official repo - or rather, under `/etc/portage/repos.conf/` I pull from here ^^

kiwifb:
No the github repo is a synchronised mirror of the real thing which is served from gentoo.org. You can, of course, pull from github instead of git.gentoo.org. 
I have lost access during the git migration I believe. Their must have been a review of the various keys or they only migrated the ones with known devs.

sci-biology/fsl/fsl-5.0.8.ebuild
```diff
@@ -108,7 +108,8 @@ src_install() {
 	sed -i "s:\$FSLDIR/tcl:/usr/libexec/fsl:g" \
 		$(grep -l "\$FSLDIR/tcl" tcl/*) || die
 
-	dobin bin/*
+	exeinto /usr/share/${PN}/bin
```
jlec: Better use ```/usr/libexec/${PN}``` here.

TheChymera: Why is that preferred?

TheChymera: @jlec I actually believe that would make things much more difficult. FSL treats `$FSLDIR` like a sort of FSL-only subsystem. It is supposed to contain `$FSLDIR/bin` - but not only. Have a look at an example from the FSL sources:

```
29:FSLTCLSH=$FSLDIR/bin/fsltclsh
30:FSLWISH=$FSLDIR/bin/fslwish
52:FSLCONFDIR=$FSLDIR/config
53:FSLMACHTYPE=`$FSLDIR/etc/fslconf/fslmachtype.sh`
```

kiwifb:
I believe we patch somewhat for that problem `53` doesn't go well with `29` and `30` once you point `FSLDIR` to `EPREFIX/usr`.

TheChymera:
@kiwifb why would it not? we'll just have different directories under `$FSLDIR`

sci-biology/fsl/fsl-5.0.8.ebuild
jlec: > Why is that preferred?

Executables which aren't placed in *bin should go there.

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s07.html

TheChymera: Yes, can we create `usr/libexec/fsl/bin` then? (and also `usr/libexec/fsl/etc` etc.

Otherwise we would pretty much have to hack all of their sources (and expect stuff to bream on updates, and things that depend on FSL not to work - since we are not organizing stuff the way FSL and stuff that depends on it expect to find it).

TheChymera: @jlec @kiwifb any ideas on this?

PR #614, opened 2016-04-11
sci-libs/vtk: new revision with qt5 useflag, updated ebuild to EAPI 6

timeraider4u:
Dear all, 
I have updated the ebuild vtk-6.1.0-r2.ebuild to enable compilation against QT5 and to use the new EAPI. Please let me know your feedback. Thanks!


jlec:
@timeraider4u thanks for the patch. If I haven't merged your PR within a week, please ping me again. 

I would like to diff the ebuilds and have a test build.

timeraider4u:
@jlec 
Attached you can find the diff between the original file and the modified VTK ebuild:
[vtk-6.1.0-r3.patch.txt](https://github.com/gentoo-science/sci/files/234768/vtk-6.1.0-r3.patch.txt)
A build log showing that VTK compiled without problems can be found here: https://github.com/timeraider4u/testing/blob/master/build.log.txt

The only thing I do not know is how to transform line # 239 of the ebuild to the new API:
`# How to convert this to EAPI 6???
+		#$(cmake-utils_use web VTK_WWW_DIR="${ED}/${MY_HTDOCSDIR}")`

jlec:
@timeraider4u Could you please make either of the qt variants as default? e.g. ```IUSE="qt +qt5"```.

timeraider4u:
@jlec: qt5 is now set as a default ;-).

PR #617, opened 2016-04-26
sci-libs/nipype is python 3.4 compatible

TheChymera:


jlec:
Please drop python 3_2 support when you are at it.

PR #625, opened 2016-05-14
sci-biology/ants: install scripts

TheChymera:
ANTs comes with a umber of scripts which call its core functions - and which are what many researchers preferentially use.

### DIFF
--- a/sci-biology/ants/ants-9999.ebuild
+++ b/sci-biology/ants/ants-9999.ebuild
@@ -22,8 +22,7 @@ src_install() {
 	cd "${WORKDIR}/${P}_build/ANTS-build" || die "build dir not found"
 	default
 	cd "${WORKDIR}/${P}/Scripts" || die "scripts dir not found"
-	exeinto /usr/bin
-	doexe *.sh
+	dobin *.sh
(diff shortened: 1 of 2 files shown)


sci-biology/ants/ants-9999.ebuild
```diff
@@ -22,6 +22,9 @@ src_install() {
 	cd "${WORKDIR}/${P}_build/ANTS-build" || die "build dir not found"
 	default
 	cd "${WORKDIR}/${P}/Scripts" || die "scripts dir not found"
+	exeinto /usr/bin
+	doexe *.sh
```
jlec: This is what ```dobin```does.

sci-biology/ants/files/99ants
```diff
@@ -0,0 +1 @@
+ANTSPATH=/usr/bin/
```
jlec: Is this really needed?

TheChymera: yes, sadly it is. Some ANTs scripts check if the variable is set when they are called and fail otherwise.

TheChymera:
@jlec ?

PR #638, opened 2016-06-23
sci-libs/gsl: Revise gsl 2.1 to remove int64 support that doesn't act…

kiwifb:
…ually exist.


kiwifb:
The current `gsl-2.1` ebuild includes int64 support via eclass from the overlay that has no basis in reality and has a number of defects, most of them a consequence of the int64 support. See discussion at https://github.com/cschwan/sage-on-gentoo/issues/429
The current ebuild doesn't pass its test suite with 32bit ABI on a native x86_64 machine, the same tests fail with the current ebuild, so it is not a regression. There are several discussions floating around about the test tolerance not being big enough for 32bit machines. The patch included in fedora 24 doesn't solve the failures observed on my machine.

jlec:
Could you please do a ```git mv``` so that we can see the changes?

kiwifb:
I'll have to do things properly at a later stage. Something funny happened to my key, I'll need to generate a new one.

jlec:
Could you please use the numeric.eclass?

kiwifb:
OK will do that ASAP (probably a few hours from now).

kiwifb:
Doing both 2.1 and 2.3. Still some work to do on the tolerance of the tests. And I switched to using the numeric eclass.

kiwifb:
@jlec I actually will suggest something a little bit funny. About 80% of the `cblas-external` patch is rewiring compilation of tests to a cblas that may or may not be gslcblas. Some of the tolerance problem we see in the tests are the result of using another cblas. I was working on making the tests work with openblas, but what about the other cblas out there? Any and each one of them could need some tolerance tweaking. And it looks like the tests in the `poly` subfolder are just plain disabled.

So, since we always build gslcblas anyway I say instead of allowing an external cblas to be used to run the test suite, we just run it with gslcblas - the way upstream does it. It is unlikely we will uncover anything by using another cblas.

The only thing where we may want to do something is tolerance in x86_32 as upstream doesn't test that anymore and the tolerance are too small in that case even with gslcblas.

kiwifb:
Gave this branch a bit of love. So in summary:

- remove `int64` and therefore switch to `multilib-build`+`numeric` eclasses
- rev-bump because the stuff installed with the current ebuilds is broken (but is less so with a particular combination of use flags) see https://github.com/cschwan/sage-on-gentoo/issues/429
- right now I changed the subslot in gsl 2.3 back to `0/19`. Both gsl 2.1 and 2.3 produce a libgsl with libgsl.so.19 as the soname. If there is a particular reason to increment the subslot, I am happy to put it back.
- changed `gsl-2.3-external-cblas.patch` (and only that one) not to patch the makefiles used by the testsuite. The testsuite will be run with `gslcblas`, even when `external-cblas` is used. 

Note enabling `external-cblas` doesn't prevent `gslcblas` from being build. Testing with `gslcblas` means there are no issues of tolerance in the results from using a different cblas. Upstream only tests on amd64 so other arch may still have tolerance problems. Other distro carry tolerance patches, but it seem to be sensitive to the setup and toolchain. i.e. the tolerance problems I have by testing with `abi_x86_32` on my amd64 gear are different than the one patched by fedora. We may have to restrict testing to `abi_x86_64`.

jlec:
Could you please use `git mv` so that it is easier to spot the diff?

kiwifb:
All right I will do it for gsl-2.1, it's pretty much the same for 2.3. But I insist that those need to be bumped to force rebuild before it is merged.

kiwifb:
Done.

jlec:
THanks. Could you do that for 2.3 as well? In general it LGTM. There are some `cd`which need a `die`

kiwifb:
Actually the `multilib-build` eclass inherits `multibuild` so I will replace those `cd` with `run_in_build_dir`. I thought it was overkill, but it is inherited so I may as well use it.

kiwifb:
All done.

kiwifb:
Hum.... Where did I get all those keywords in 2.3. If they were the ones dropped recently `repoman` should have told me something. I will put it back in line.

kiwifb:
@jlec if you have no further objections, I'll merge on github.

kiwifb:
@marbre out of sync again ;)

marbre:
Not anymore ;)

PR #651, opened 2016-07-22
fix SRC_URI

freeh4cker:
as the src seems to move on an other server


sci-biology/hmmer/hmmer-3.1_beta2.ebuild
```diff
@@ -9,7 +9,7 @@ inherit eutils
 MY_PV="3.1b2"
```
marbre: The version shoudn't be hardcoded. Give the versionator eclass a try instead.

sci-biology/hmmer/Changelog
```diff
@@ -0,0 +1,7 @@
+# ChangeLog for sci-biology/squiz
+# Copyright 1999-2016 Gentoo Foundation; Distributed under the GPL v2
+# $Id$
+
+  22 Jul 2016 Bertrand Néron <freeh4cker@gmail.com> hmmer-3.1_beta2,
+	update SRC_URI
+
```
marbre: With portage now using git, changelogs were dropped. Same for the science overlay, since we have the git history.

freeh4cker: Ok, I did not know that.
I will drop it

sci-biology/hmmer/hmmer-3.1_beta2.ebuild
freeh4cker: I don't modify this part of code, it is like this in master branch.
versionator eclass is smart enough to translate the gentoo version
3.1_beta2 in url version 3.1b2?
or have i to change the version of the ebuild for 3.1b2?

thank's

bertrand

marbre: I know that this was already in the ebuild, but if you send a pull request its welcome if you could further improve (this part of) the ebuild :)
The naming of the ebuild is fine, see https://devmanual.gentoo.org/ebuild-writing/file-format/. However, `MY_PV="3.1b2"` could rewritten using the versionator eclass or using a bash substitution. Please take a look at *Version and Name Formatting Issues* in https://devmanual.gentoo.org/ebuild-writing/variables/. The versionator should be smart enough to handle the translation, but I am not using it that regularly. Please give it a try.

sci-biology/hmmer/hmmer-3.1_beta2.ebuild
```diff
@@ -2,14 +2,16 @@
 # Distributed under the terms of the GNU General Public License v2
 # $Id$
 
-EAPI=5
+EAPI=6
 
 inherit eutils
 
-MY_PV="3.1b2"
+
```
marbre: Running `repoman full` returns:
```
  ebuild.minorsyn               1
   sci-biology/hmmer/hmmer-3.1_beta2.ebuild: Useless blank line on line: 9
```

For future PRs, please use `repoman commit`, which is automatically checking your ebuild (running `repoman full`), updates the Manifest and so on...

marbre:
@jlec @SoapGentoo Instead of several commits, one could shorten the PR. For example commit 0469c54 mainly fixes issues that came with commit dfa3f17, included in the same PR. Do we have a policy for this case? I actually like "clean", incremental commits, but that causes additional effort. Your opinion?

SoapGentoo:
@marbre, your suggestions are absolutely correct. In the end, we have 5 commits changing one (!) ebuild. Squash them all into one, and have the `COMMITMSG` done properly with `repoman commit`. While we don't have a formalised policy on this, the QA spirit of contributions should be, "if there is no rule, do as they do in the main Gentoo tree".

marbre:
@SoapGentoo Thanks for your comment. @freeh4cker Thus we would appreciate if you could squash your changes into a single commit. Just use `git reset --soft ${commithash}`, to remove the commits, without losing yout changes (thus do not use --hard!) and use `repoman commit` to re-commit.

freeh4cker:
sorry, I failed to squash of my commits on this branch, it generate a mess on my repo. So I delete the branch clean my repo and create new hmmer branch and re-submit a pull request   #654, containing only one commit.

PR #654, opened 2016-07-29
fix SRC_URI

freeh4cker:
as the src seems to move on an other server
by the way fix:
- the version is not hardcoded anymore, use shell string substitution and
  gentoo variable PV to compute the version for the url
- bump EAPI to version 6
- fix copyright


SoapGentoo:
@freeh4cker while the commit is fine now, the `COMMITMSG` is still a plain git one. Use repoman to produce a proper QA-compliant commit message.

marbre:
The ebuild moved from the science overlay to the tree. Current hmmr-3.1_beta2 in the tree already includes your fixes. Thanks for your PR.

PR #660, opened 2016-08-26
sci-biology/psychopy: added deps for video rendering

TheChymera:
More deps for psychopy (opencv, moviepy).
Added threads requirements.


TheChymera:
@jlec this is also finished now.

PR #661, opened 2016-08-28
media-libs/avbin-bin: avbin has moved + bump

TheChymera:

fixes https://bugs.gentoo.org/show_bug.cgi?id=592328
partly addresses #663


media-libs/avbin-bin/avbin-bin-8.1.ebuild
```diff
@@ -0,0 +1,28 @@
...
+
+EAPI=6
+
+DESCRIPTION="A thin wrapper around FFmpeg"
+HOMEPAGE="http://avbin.github.io/"
+SRC_URI="
+	amd64? ( mirror://github/AVbin/AVbin/avbin-linux-x86-64-v${PV}.tar.bz2 )
+	x86? ( mirror://github/AVbin/AVbin/avbin-linux-x86-32-v${PV}.tar.bz2 )"
+
+LICENSE="GPL-3 LGPL-3"
```
jlec: Why don't you add the source based ebuild? GPL requires that the source need be accessible. [TLDR legal](https://tldrlegal.com/license/gnu-general-public-license-v3-(gpl-3))

media-libs/avbin-bin/avbin-bin-8.1.ebuild
```diff
@@ -0,0 +1,28 @@
...
+
+pkg_setup(){
+	if use amd64; then
+		S="${WORKDIR}"/avbin-linux-x86-64-v${PV}
+	elif use x86; then
+		S="${WORKDIR}"/avbin-linux-x86-32-v${PV}
+	fi
+}
+src_install() {
+	ln -s libavbin.so.${PV} libavbin.so || die
```
jlec: is this really the soname?

TheChymera: Yes, it is.

media-libs/avbin-bin/avbin-bin-8.1.ebuild
TheChymera: The source build [is a mess](663#issuecomment-243397438).

Is there any way we can "make the source accessible"? AVbin isn't entirely deprecated yet, and `sci-biology/psychopy` video playback (which depends on it) [works considerably better with `=avbin-8.1`](http://discourse.psychopy.org/t/sound-quality-for-moviestim-is-very-bad/955/4?u=horea_christian) - so I think we should definitely bump this independently of whether or not we end up providing a source package.

TheChymera:
@jlec ?

PR #664, opened 2016-08-29
Add support for multilib mpi to mpi.eclass

junghans:
See [bug #519700](https://bugs.gentoo.org/show_bug.cgi?id=519700) for details

### DIFF
--- a/sys-cluster/mpich/mpich-3.2-r2.ebuild
+++ b/sys-cluster/mpich/mpich-3.2-r2.ebuild
@@ -16,7 +16,7 @@ SRC_URI="http://www.mpich.org/static/downloads/${PV}/${P}.tar.gz"
 
 SLOT="0"
 LICENSE="mpich"
-KEYWORDS=""
+KEYWORDS="~amd64 ~hppa ~ppc ~ppc64 ~x86 ~amd64-linux ~x86-linux"
 IUSE="+cxx doc fortran mpi-threads romio threads"
 
 COMMON_DEPEND="
--- a/sys-cluster/mpich/mpich-3.2-r2.ebuild
+++ b/sys-cluster/mpich/mpich-3.2-r2.ebuild
@@ -1,0 +1,120 @@
+# Copyright 1999-2016 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
+
+EAPI=6
+
+FORTRAN_NEEDED=fortran
+FORTRAN_STANDARD="77 90"
+
+inherit fortran-2 multilib-minimal mpi
+
+MY_PV=${PV/_/}
+DESCRIPTION="A high performance and portable MPI implementation"
+HOMEPAGE="http://www.mpich.org/"
+SRC_URI="http://www.mpich.org/static/downloads/${PV}/${P}.tar.gz"
+
+SLOT="0"
+LICENSE="mpich"
+KEYWORDS=""
+IUSE="+cxx doc fortran mpi-threads romio threads"
+REQUIRED_USE="mpi-threads? ( threads )"
+
+COMMON_DEPEND="
+	>=dev-libs/libaio-0.3.109-r5[${MULTILIB_USEDEP}]
+	>=sys-apps/hwloc-1.10.0-r2[${MULTILIB_USEDEP}]
+	romio? ( net-fs/nfs-utils )
+	$(mpi_imp_deplist)"
+
+DEPEND="${COMMON_DEPEND}
+	dev-lang/perl
+	sys-devel/libtool"
+
+RDEPEND="${COMMON_DEPEND}"
+
+S="${WORKDIR}"/${PN}-${MY_PV}
+
+MULTILIB_WRAPPED_HEADERS=(
+	/usr/include/mpicxx.h
+	/usr/include/mpi.h
+	/usr/include/opa_config.h
+)
+
+src_prepare() {
+	default
+
+	# Using MPICHLIB_LDFLAGS doesn't seem to fully work.
+	sed -i 's| *@WRAPPER_LDFLAGS@ *||' \
+		src/packaging/pkgconfig/mpich.pc.in \
+		src/env/*.in \
+		|| die
+}
+
+multilib_src_configure() {
+	# The configure statements can be somewhat confusing, as they
+	# don't all show up in the top level configure, however, they
+	# are picked up in the children directories.
+
+	export MPICHLIB_CFLAGS="${CFLAGS}"
+	export MPICHLIB_CPPFLAGS="${CPPFLAGS}"
+	export MPICHLIB_CXXFLAGS="${CXXFLAGS}"
+	export MPICHLIB_FFLAGS="${FFLAGS}"
+	export MPICHLIB_FCFLAGS="${FCFLAGS}"
+	export MPICHLIB_LDFLAGS="${LDFLAGS}"
+	unset CFLAGS CPPFLAGS CXXFLAGS FFLAGS FCFLAGS LDFLAGS
+
+	ECONF_SOURCE=${S} $(mpi_econf_args) econf \
+		--enable-shared \
+		--sysconfdir="${EPREFIX}/etc/${PN}" \
+		--with-hwloc-prefix="${EPREFIX}/usr" \
+		--enable-threads=$(usex mpi-threads runtime single) \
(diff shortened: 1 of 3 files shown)


sys-cluster/mpich/mpich-3.2-r2.ebuild
```diff
@@ -0,0 +1,120 @@
...
+
+	# fortran header cannot be wrapped (bug #540508), workaround part 1
+	if multilib_is_native_abi && use fortran; then
+		mkdir "${T}"/fortran || die
+		mv "${ED}"usr/include/mpif* "${T}"/fortran || die
+		mv "${ED}"usr/include/*.mod "${T}"/fortran || die
+	else
+		# some fortran files get installed unconditionally
+		rm "${ED}"usr/include/mpif* "${ED}"usr/include/*.mod || die
+	fi
```
SoapGentoo: @junghans is the fortran header thing still a problem? i.e. do the fortran headers differ between ABIs? I'm asking because this problem evaporated with fftw when we properly multilib'd the mpi-dependent stuff.

jsbronder:
```
sudo empi --create sys-cluster/openmpi -C mpi-ompi
```

```
make[2]: Leaving directory '/var/tmp/portage/mpi-ompi/openmpi-1.10.3-r2/work/openmpi-1.10.3-abi_x86_64.amd64'
make[1]: Leaving directory '/var/tmp/portage/mpi-ompi/openmpi-1.10.3-r2/work/openmpi-1.10.3-abi_x86_64.amd64'
mv: cannot stat ‘/var/tmp/portage/mpi-ompi/openmpi-1.10.3-r2/image/usr/include/mpif*’: No such file or directory
 * ERROR: mpi-ompi/openmpi-1.10.3-r2::x-mpi failed (install phase):
 *   (no error message)
 * 
 * Call stack:
 *     ebuild.sh, line  133:  Called src_install
 *   environment, line 4945:  Called multilib-minimal_src_install
 *   environment, line 4082:  Called multilib_foreach_abi 'multilib-minimal_abi_src_install'
 *   environment, line 4276:  Called multibuild_foreach_variant '_multilib_multibuild_wrapper' 'multilib-minimal_abi_src_install'
 *   environment, line 3969:  Called _multibuild_run '_multilib_multibuild_wrapper' 'multilib-minimal_abi_src_install'
 *   environment, line 3967:  Called _multilib_multibuild_wrapper 'multilib-minimal_abi_src_install'
 *   environment, line  698:  Called multilib-minimal_abi_src_install
 *   environment, line 4072:  Called multilib_src_install
 *   environment, line 4498:  Called die
 *           mv "${ED}"usr/include/mpif* "${T}"/fortran || die;
 * 
```

sys-cluster/mpich/mpich-3.2-r2.ebuild
junghans: In fftw, all fortran headers are identical for all abis and hence don't need to wrapped, but that isn't the case for mpi.

junghans:
@jsbronder, 91299b3 should fix the above issue

sys-cluster/mpich/mpich-3.2-r2.ebuild
```diff
@@ -0,0 +1,120 @@
...
+
+	export MPICHLIB_CFLAGS="${CFLAGS}"
+	export MPICHLIB_CPPFLAGS="${CPPFLAGS}"
+	export MPICHLIB_CXXFLAGS="${CXXFLAGS}"
+	export MPICHLIB_FFLAGS="${FFLAGS}"
+	export MPICHLIB_FCFLAGS="${FCFLAGS}"
+	export MPICHLIB_LDFLAGS="${LDFLAGS}"
+	unset CFLAGS CPPFLAGS CXXFLAGS FFLAGS FCFLAGS LDFLAGS
+
+	ECONF_SOURCE=${S} $(mpi_econf_args) econf \
```
jsbronder: Pretty sure you want ```$(mpi_econf_args)``` after ```econf``` here.

sys-cluster/mpich/mpich-3.2-r2.ebuild
```diff
@@ -0,0 +1,120 @@
...
+		--enable-shared \
+		--sysconfdir="${EPREFIX}/etc/${PN}" \
+		--with-hwloc-prefix="${EPREFIX}/usr" \
+		--enable-threads=$(usex mpi-threads runtime single) \
+		--with-thread-package=$(usex threads pthreads none) \
+		--with-pm=hydra \
+		--disable-fast \
+		--enable-versioning \
+		--with-hwloc-prefix=/usr \
+		$(mpi_classed && echo "--docdir=$(mpi_root)/usr/share/doc/${PF}") \
```
jsbronder: This is inverting the previous logic:
```
    if ! mpi_classed; then
        c="${c} --sysconfdir=${EPREFIX}/etc/${PN}"
        c="${c} --docdir=${EPREFIX}/usr/share/doc/${PF}"
    else
        c="${c} --docdir=$(mpi_root)/usr/share/doc/${PF}"
    fi
```

sys-cluster/mpich/mpich-3.2-r2.ebuild
```diff
@@ -0,0 +1,120 @@
...
+	export MPICHLIB_CPPFLAGS="${CPPFLAGS}"
+	export MPICHLIB_CXXFLAGS="${CXXFLAGS}"
+	export MPICHLIB_FFLAGS="${FFLAGS}"
+	export MPICHLIB_FCFLAGS="${FCFLAGS}"
+	export MPICHLIB_LDFLAGS="${LDFLAGS}"
+	unset CFLAGS CPPFLAGS CXXFLAGS FFLAGS FCFLAGS LDFLAGS
+
+	ECONF_SOURCE=${S} $(mpi_econf_args) econf \
+		--enable-shared \
+		--sysconfdir="${EPREFIX}/etc/${PN}" \
```
jsbronder: See comment below, this is a change in behaviour from the previous ebuild.

sys-cluster/mpich/mpich-3.2-r2.ebuild
```diff
@@ -0,0 +1,120 @@
...
+MY_PV=${PV/_/}
+DESCRIPTION="A high performance and portable MPI implementation"
+HOMEPAGE="http://www.mpich.org/"
+SRC_URI="http://www.mpich.org/static/downloads/${PV}/${P}.tar.gz"
+
+SLOT="0"
+LICENSE="mpich"
+KEYWORDS=""
+IUSE="+cxx doc fortran mpi-threads romio threads"
+REQUIRED_USE="mpi-threads? ( threads )"
```
jsbronder: According to note at the end of the section at [1], it's preferable to just do the right thing when we know what the user wants.  In this case, I think the old ebuild got it right by enabling ```threads``` if ```mpi-threads``` was enabled.

1.  https://devmanual.gentoo.org/general-concepts/use-flags/index.html#conflicting-use-flags

jsbronder:
Aside from the comments I've already made, this looks great.  You guys get your multilib in the main tree and I get to easily move ebuilds back and forth.

sys-cluster/mpich/mpich-3.2-r2.ebuild
junghans: @SoapGentoo: that was a port of your edit from gx86.

junghans: Re-added!

sys-cluster/mpich/mpich-3.2-r2.ebuild
SoapGentoo: (Bear with me for the moment) I think the proper long-term solution is to integrate fortran support into the `multilib.eclass` code. While fortran officially has no such thing as a preprocessor, GFortran and ifort both support a C-like preprocessor. I think this would be the best solution, as we're actually breaking the installed headers for one of the ABIs. For the short-term we can leave it as is, but I'd like to fix this properly.

junghans:
@jsbronder: merge when happy!

sys-cluster/mpich/mpich-3.2-r2.ebuild
junghans: `mod` files are binary and cannot be pre-processed. F03 could be pre-processed by renaming them to `F03`, but that would make Gentoo non-standard for fortran included.

I think for now, we cannot support multilib fortran, which means install the native headers and drop the others.

PR #670, opened 2016-09-07
sci-libs/asl: version bump to 0.1.6

otwieracz:


sci-libs/asl/asl-0.1.6.ebuild
```diff
@@ -0,0 +1,48 @@
+# Copyright 1999-2015 Gentoo Foundation
```
jlec: Wrong year

sci-libs/asl/asl-0.1.6.ebuild
```diff
@@ -0,0 +1,48 @@
...
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
+
+EAPI=5
```
jlec: Please use 6

PR #673, opened 2016-09-14
net-misc/cernbox-client: Branded net-misc/owncloud-client for CERN.

olifre:
The branding contains themeing and several special default settings
for the special ownCloud instance of CERN.

The branding is extracted from the official RPM package for CentOS (as can be seen from the SRC URI). The CERNbox team encourages packaging for not-officially-supported distributions by the community. 

The CERNbox ownCloud should only be used with such a branded client due to some specific client settings (e.g. the transferChecksum-setting is changed to match the CERN-EOS backend). 

This package has some additional backported fixes to the shell-integration for branded ownCloud clients (issues either already fixed or reported upstream). 


net-misc/cernbox-client/cernbox-client-2.2.3.ebuild
```diff
@@ -0,0 +1,119 @@
...
+	# Keep tests in ${T}
+	sed -i -e "s#\"/tmp#\"${T}#g" test/test*.cpp || die "sed failed"
+
+	if ! use nautilus; then
+		pushd shell_integration > /dev/null || die
+		cmake_comment_add_subdirectory nautilus
+		popd > /dev/null || die
+	fi
+
+	epatch 	"${FILESDIR}"/shell-integration-nautilus.patch \
```
jlec: Please use `eapply`

olifre: Done, thanks! Should I squash the two commits or do you want to pull them like this?

PR #691, opened 2016-10-14
sci-mathematics/gismo: Version bump to 2016.10.01, added axel-plugin-support

timeraider4u:

added sci-libs/dtk, sci-mathematics/axel and sci-mathematics/axel-vtkview
updated ebuild and patches for sci-mathematics/gismo


timeraider4u:
Closed because a problem after installation has been encountered. Will fix this soon and create a new pull request for that.

PR #696, opened 2016-10-27
net-misc/cernbox-client: many cleanups, release bump

olifre:
Follow along net-misc/owncloud-client: Drop qt4 useflag, and cleanup ebuild.


olifre:
Replaced by #754 .

PR #702, opened 2016-11-23
sci-libs/vtk: deleted sci-libs/vtk because already in portage tree

timeraider4u:
sci-libs/vtk: deleted sci-libs/vtk because already in portage tree
(see https://github.com/gentoo/gentoo/pull/2663)


timeraider4u:
Thanks for merging!

PR #712, opened 2016-12-16
sci-biology/ncbi-genome-download: new ebuild

freeh4cker:
ncbi-genome-dowload first packaging.

Package-Manager: portage-2.3.2


sci-biology/ncbi-genome-download/ncbi-genome-download-0.2.2.ebuild
```diff
@@ -0,0 +1,21 @@
...
+HOMEPAGE="https://github.com/kblin/ncbi-genome-download"
+SRC_URI="https://github.com/kblin/${PN}/archive/${PV}.tar.gz  -> ${P}.tar.gz"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
+
+DEPEND=""
+RDEPEND="${DEPEND}"
```
jlec: Drop these if they are empty.

PR #713, opened 2016-12-17
sci-mathematics/scilab: Require xml-commons-external-1.3, and unmask.

khumba:
Scilab requires xml-commons-external's xml-apis-ext.jar.  xml-commons-external-1.3 provides this but 1.4 has a different JAR name, and Scilab's configure phase breaks as a result (even with both 1.3 and 1.4 installed).  This makes it build again by downgrading the dep to 1.3.

### DIFF
--- a/profiles/package.mask
+++ b/profiles/package.mask
@@ -34,7 +34,6 @@
 # rev dep absent
 sci-biology/genseed
 sci-biology/tgicl
-sci-mathematics/scilab
 
 # Justin Lecher <jlec@gentoo.org> (24 April 2016)
 # rev dep missing keywords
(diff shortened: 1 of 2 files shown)


khumba:
Also, can we unmask Scilab, @jlec?  All revdeps seem to be present here.

TheChymera:
@khumba would you be interested in taking this up again?

khumba:
It's been quite a while since I've used Scilab, and I'm not interested in resurrecting this now.  It looks like there are other missing deps now (dev-java/batik:1.8), and the ebuild probably needs bumping to the latest Scilab version (and also updating to newer EAPI).  Apologies.

TheChymera:
@khumba no worries, we cleaned a lot of things up to remove fatal errors from the overlay. There is no sort of policy to not include those things, just that there were too many things obviously broken for us to fix all of them in any reasonable amount of time. If you want to bring back scilab and its deps at any point, we would be very happy to pull.

PR #718, opened 2017-01-05
sci-biology/fsl: version bump

TheChymera:
Trying to version bump:

* @kiwifb I am running into issues with the patches again. It's [not finding a lot of files](https://bpaste.net/show/360c9da24bfa), but I can find those files in the workdir.
* @jlec coud you reposit the sources in your devspace? Upstream will delete them as soon as a new version comes out.


kiwifb:
```
PATCH COMMAND:  patch -p1 -g0 -E --no-backup-if-mismatch  --dry-run -f < '/home/chymera/src/neurogentoo/sci-biology/fsl/files/fsl-5.0.8-setup.patch'

=================================
checking file build
Hunk #1 succeeded at 12 with fuzz 1 (offset 1 line).
checking file config/common/buildproj
checking file config/common/vars.mk
checking file config/generic/systemvars.mk
checking file extras/build
Hunk #1 FAILED at 60.
1 out of 1 hunk FAILED
```
the file `extras/build` has changed and the patch doesn't apply properly anymore. I cannot tell you more than that without downloading fsl-5.0.9. But only one line was changed by that bit, you should be able to work it out.

jlec:
@TheChymera `https://dev.gentoo.org/~jlec/distfiles/${P}.tar.gz`

TheChymera:
I've re-run all of my pipelines with this without issues, @jlec @kiwifb  - can we bump this?

sci-biology/fsl/fsl-5.0.9.ebuild
```diff
@@ -0,0 +1,147 @@
...
+		$(grep -rl "\${FSLDIR}/etc" src/*)
+
+	sed -i "s:\$FSLDIR/doc:${EPREFIX}/usr/share/fsl/doc:g" \
+		$(grep -rl "\$FSLDIR/doc" src/*)
+
+	sed -i "s:\${FSLDIR}/doc:${EPREFIX}/usr/share/fsl/doc:g" \
+		$(grep -rl "\${FSLDIR}/doc" src/*)
+
+	sed -i "s:\'\${FSLDIR}\'/doc:${EPREFIX}/usr/share/fsl/doc:g" \
+		$(grep -rl "\'\${FSLDIR}\'/doc" src/*)
```
TheChymera: * what do you mean by "move all the seds together"? They are separate statements.
* FSL has very very many files, we use grep to make sure we only sed in the files that match the pattern. this is a lot faster than just using sed.

TheChymera: @jlec ?

TheChymera: @kiwifb  - the `grep` makes sense, right?

kiwifb: Yes, what jlec means is that all those sed commands can be condensed into one, like so
```
sed -e "s:\$FSLDIR/etc:${EPREFIX}/etc:g" \
       -e "s:\${FSLDIR}/etc:${EPREFIX}/etc:g" \
      ....
      -i $(grep -rl "\'\${FSLDIR}\'/doc" src/*)
```
They all apply to the same list of files. In GNU `sed`, `-i` is for "in place" - it does the editing in the file itself (otherwise it spits out on the standard output). I said in GNU `sed` because the POSIX standard for `sed` doesn't include it - if you use IBM AIX instead of linux for example it is not available. But I digress, `-e` is to add a "sed" command, you add many commands together by preceding them with `-e`.

TheChymera:
#thatmomentwhen git blows up in your face, again :-/

PR #724, opened 2017-01-09
Add new gmsh ebuild.

mrkwjc:
Updated to EAPI=6. Added python USE flag for python wrappers building.


sci-libs/gmsh/gmsh-2.16.0.ebuild
```diff
@@ -0,0 +1,92 @@
...
+	cmake-utils_src_configure mycmakeargs
+}
+
+src_install() {
+	cmake-utils_src_install
+
+	# TODO: tutorials get installed twice ATM
+	if use doc ; then
+		cd "${BUILD_DIR}" || die
+		emake pdf
```
mrkwjc: Sorry i didn't get what exactly do you want me to move. Do you mean:

```
src_compile() {
	if use doc ; then
		cd "${BUILD_DIR}" || die
		emake pdf
		cd "${S}" || die
		dodoc doc/texinfo/gmsh.pdf
	fi
}

src_install() {
	cmake-utils_src_install

	if use examples ; then
		insinto /usr/share/doc/${PF}
		doins -r demos tutorial
	fi
}
```

jlec: @mrkwjc yes.

PR #731, opened 2017-02-14
Bump Frama-C to v 14.0 Silicon (20161101) version bump

XVilka:
Bump Frama-C to 20161101 version.
Now it requires OCaml higher or equal than 4.02.3
See also https://bugs.gentoo.org/show_bug.cgi?id=480430


marbre:
You don't use `repoman commit` for commiting. Otherwise you would have noticed that the digest is missing (see travis build failure), the copyright is still 2016 and that you're stil using EAPI=5. Please use `repoman full` to check your commit before commiting!

XVilka:
Thx! Will do this weekend.

TheChymera:
@XVilka would you be interested in submitting a PR with a more recent version?

PR #743, opened 2017-02-28
sci-mathematics/gismo: version bump

timeraider4u:
sci-mathematics/gismo: version bump

Package-manager: portage-2.3.3 repoman-2.3.1


timeraider4u:
thanks!

PR #749, opened 2017-03-05
Intel liberation

kiwifb:
The use of `MULTILIB_COMPAT=( abi_x86_{32,64} )` throughout the blas infrastructure in the overlay prevent any blas/lapack to be used on arch other than x86/amd64. This is perfectly justified for intel mkl and related eclass but not for something like lapack-reference. As per the `multilib-build` man page:

> Note that setting this variable effectively disables support for all other ABIs, including other architectures. For example, specifying abi_x86_{32,64} disables support for MIPS as well.

Since this is implemented as `REQUIRED_USE` restrictions, no amount of `**` in a keyword file will help. Only hacking the eclass and virtual ebuilds will help. You cannot even test the ebuild, it goes much further than signalling no one support the ebuild on other archs.

For the record, I currently have access to ppc64 hardware and have been using the overlay there after modification. And it is perfectly OK and sensible to use `numeric-int64-multibuild` eclass on such architecture. Nothing ground breaking as far as I am concerned.

### DIFF
--- a/eclass/numeric-int64-multibuild.eclass
+++ b/eclass/numeric-int64-multibuild.eclass
@@ -23,8 +23,6 @@ case ${EAPI:-0} in
 	*) die "EAPI=${EAPI} is not supported" ;;
 esac
 
-MULTILIB_COMPAT=( abi_x86_{32,64} )
-
 inherit alternatives-2 eutils fortran-2 multilib-build numeric toolchain-funcs
 
 IUSE="int64"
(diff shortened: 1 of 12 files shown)


jlec:
Is there any change in stallation which you have observed? If not, I am fine with this change. @SoapGentoo Comments?

kiwifb:
Not really any changes that I can tell. Just that I am b***** allowed to install stuff on non-intel class hardware. I left one instance in `gsl-1.16` because 1) I didn't want to mess with my other current PR, 2) Unless there is a hard dependency on that version somewhere, I think it is time to retire it.

SoapGentoo:
looks ok

kiwifb:
I cannot merge this myself. I can merge on github but I cannot merge in gentoo.org anymore, so some else will have to merge it.

marbre:
@kiwifb If you feel that this is ready to merge, just merge it to github. The overlay and github are out of sync often anyway, I'll bring them into sync again as soon as I notice it.

kiwifb:
Done.

marbre:
We are in sync again :)

PR #754, opened 2017-03-14
net-misc/cernbox-client: Bump to 2.3.0,

olifre:
adapt 2.2.4 ebuild to in-tree changes
of owncloud-client, use updated cernbox-branding.

Package-Manager: Portage-2.3.4, Repoman-2.3.2


net-misc/cernbox-client/cernbox-client-2.3.0.ebuild
```diff
@@ -0,0 +1,103 @@
...
+
+# Origin is classic owncloud-client, branded for CERN during compilation.
+ORIGIN_P="owncloudclient"
+BRANDED_P="cernbox"
+BRANDED_PV="2.2.4"
+BRANDED_REL="1.1"
+
+HOMEPAGE="https://cernbox.cern.ch/"
+SRC_URI="http://download.owncloud.com/desktop/stable/${ORIGIN_P}-${PV}.tar.xz
+	 https://cernbox.cern.ch/cernbox/doc/Linux/repo/CentOS_7/src/${PN}-${BRANDED_PV}-${BRANDED_REL}.src.rpm"
```
jlec: Please drop leading whitespace and use `repoman` to commit.

olifre: Done. Note that I also used `repoman` before, as you can see from the automatic `Package-Manager: Portage-2.3.4, Repoman-2.3.2` comment.

PR #757, opened 2017-03-21
sci-mathematics/gismo: version-bump

timeraider4u:
sci-mathematics/gismo: version-bump
sci-mathematics/axel: version-bump

Package-manager: portage-2.3.3 repoman-2.3.1


timeraider4u:
Okay, strange, but the error

 LICENSE.invalid [fatal]       1

   sci-libs/amdlibm/amdlibm-3.1.ebuild: AMD

cannot be caused by this pull-request.

PR #760, opened 2017-03-22
sci-visualization/forge: Make forge use system glm and freetype, they…

j000:
Freetype and glm are already in dependencies. Right now forge tries to download and build own versions, which causes sandbox violations.


marbre:
@j000 You're absolutely right, don't know why this got lost. Will further readd it to the current 0.9x series. Thanks.

PR #762, opened 2017-03-26
sci-biology/nilearn: importing system joblib

TheChymera:
Upstream imports joblib **not** from joblib, but from a scikits_learn bundled version.
We unbundle skits_learn for obvious reasons, so we also have to patch this.

It appears that this strange configuration arises from nilearn and scikits_learn sharing one developer who really wants this to be bundled (he has not explained why). We might be able to address the bundling issue for both packages if we bring this up again/with better arguments/in a different context:

* [trying to get **nilearn** to unbundle](https://github.com/nilearn/nilearn/pull/1398).
* [trying to get **scikits_learn** to unbundle](https://github.com/scikit-learn/scikit-learn/issues/8494)

### DIFF
--- a/sci-biology/nilearn/nilearn-0.2.6.ebuild
+++ b/sci-biology/nilearn/nilearn-0.2.6.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2016 Gentoo Foundation
+# Copyright 1999-2017 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=6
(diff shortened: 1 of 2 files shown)


sci-biology/nilearn/nilearn-0.2.6.ebuild
```diff
@@ -27,6 +27,19 @@ RDEPEND="
 	sci-libs/nibabel[${PYTHON_USEDEP}]
 	plot? ( dev-python/matplotlib[${PYTHON_USEDEP}] )"
 
+# upstream is reluctant to *not* depend on bundled scikits_learn:
+# https://github.com/nilearn/nilearn/pull/1398
+python_prepare_all() {
+	for f in nilearn/{*/*/,*/,}*.py; do
```
jlec: please declare `f` local.

sci-biology/nilearn/nilearn-9999.ebuild
```diff
@@ -28,6 +28,19 @@ RDEPEND="
 	sci-libs/nibabel[${PYTHON_USEDEP}]
 	plot? ( dev-python/matplotlib[${PYTHON_USEDEP}] )"
 
+# upstream is reluctant to *not* depend on bundled scikits_learn:
+# https://github.com/nilearn/nilearn/pull/1398
+python_prepare_all() {
+	for f in nilearn/{*/*/,*/,}*.py; do
```
jlec: same here.

PR #778, opened 2017-04-18
dev-util/emscripten: version bump

timeraider4u:
Package-manager: portage-2.3.3 repoman-2.3.2


dev-util/emscripten-fastcomp/emscripten-fastcomp-1.37.9.ebuild
```diff
@@ -40,6 +40,8 @@ src_configure() {
 	local mycmakeargs=(
 		# avoid clashes with sys-devel/llvm
 		-DCMAKE_INSTALL_PREFIX="/usr/share/${P}"
+		-DLLVM_OCAML_INSTALL_PATH="/usr/share/${P}/ocaml"
```
marbre: Can we make this prefix compatible, putting `${ED}` at the beginning of the path?

timeraider4u: cmake-utils.eclass already adds a prefix in https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/cmake-utils.eclass#n812, see also https://cmake.org/cmake/help/v3.0/variable/CMAKE_INSTALL_PREFIX.html for more details.

marbre: Thanks, wasn't aware of this :)

timeraider4u:
Thanks for merging!

PR #780, opened 2017-04-28
eclass/intel-sdp-r1: fix for the MPI wrapper scripts

abusse:
All MPI wrapper scripts were not working as the I_MPI_SUBSTITUTE_INSTALLDIR string was not replaced by the actual root dir. This commits adds a substitution for I_MPI_SUBSTITUTE_INSTALLDIR trough sed.

### DIFF
--- a/eclass/intel-sdp-r1.eclass
+++ b/eclass/intel-sdp-r1.eclass
@@ -537,6 +537,15 @@ intel-sdp-r1_src_install() {
 		eend
 	fi
 
+	if use mpi; then
+		ebegin "Setting MPI root path"
+		_escaped_root=$(echo $(isdp_get-sdp-edir) | sed 's_/_\\/_g')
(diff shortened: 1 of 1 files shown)


TheChymera:
Hi @abusse , sorry for the long hiatus. The overlay has suffered from very low manpower over the last 2-3 years. We're currently bringing it up to scratch again. Is this PR still relevant in light of recent updates?

abusse:
I wasn't able to determine whether it is still valid with a quick check. I will look into it in more detail as soon as I find some time to spare.

TheChymera:
@abusse no worries, leaving this open for now :)

abusse:
So I did some digging and as a result a created a version bump for the ICC and libraries in #1040. However, Intel seems no longer to include MPI libraries, making this patch obsolete for current packages. However, if someone still has the old ones (which I do not have access to any longer) this patch might still be helpful

epsilon-0:
Intel MPI - https://github.com/gentoo/gentoo/pull/18452

PR #784, opened 2017-05-05
=sci-physics/root-9999: Fix some not-working patches.

eteba:
(NOTE: Sorry for the commit message. Master still doesn't build)

Patching phase failed because of the following patches:

root-5.34.13-unuran.patch
root-6.06.00-nobyte-compile.patch
root-6.00.01-llvm.patch

I modified the first and the second so they apply now, and removed the last one, because ROOT now uses CMake instead of the classic configure/make method.
Maybe, the last patch should be modified to apply on CMakeLists.txt.

It builds succesfully for the "v6-08-00-patches" branch (I will be testing for some time with the "v6-08-06" tag).

However, it fails to build on the master branch: 
"Invalid option '--enable-vdt'. Try ./configure --help"

I leave it here just in case someone find it useful.


olifre:
I believe it would be cool to have this merged. Even if it does not fix things completely, at least it reduces the number of problems, and allows to build 6.08 at least.

PR #792, opened 2017-06-18
sci-libs/itk: bump to 4.12.0

tharvik:
* `EAPI=6`, so no use anymore of `cmake-utils_use*`
* dirty fix for `get_libdir`, maybe a better way to fix it?
* all patches are merged now
* some other simplifications


tharvik:
Travis is failing, the only package related error is because the other ebuilds for `sci-libs/itk` are from an older `EAPI`.

PR #794, opened 2017-06-26
sci-libs/atlas: version bump to 3.11.39

ngg:
Version 3.11.38 is not working on some of the 7th Gen i7 processors (for example on i7-7500U), that is fixed in 3.11.39.

Patch from upstream is required to build the shared library:
https://sourceforge.net/p/math-atlas/bugs/260/


ngg:
It is strange that the default use flags contain `+deprecated -lapack -fortran` but if `+deprecated` is set then both `+lapack +fortran` is required. I didn't change this behavior, but it should be fixed in a later PR, either by disabling `deprecated` or enabling `lapack` and `fortran` by default.

kiwifb:
You have a point about `+deprecated`. This is about including functions in lapack that have been deprecated for a long time and that will soon be removed. I believe those functions are still used in scipy-0.16 IIRC. So of course deprecated is non sense without lapack and lapack requires fortran.

ngg:
Or another solution would be to leave the defaults as they are but remove the `deprecated -> lapack` dependency, so if `lapack` is not set then `deprecated` is a no-op. Building dependencies will require to set the `lapack` use flag manually (as now), but at least the default flags would be a valid combination.

jlec:
@ngg could you please enable both _fortran_ and _lapack_ by default?

sci-libs/atlas/atlas-3.11.39.ebuild
```diff
@@ -0,0 +1,286 @@
+# Copyright 1999-2017 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=5
```
jlec: Bump EAPI to 6

sci-libs/atlas/atlas-3.11.39.ebuild
```diff
@@ -0,0 +1,286 @@
...
+
+DESCRIPTION="Automatically Tuned Linear Algebra Software"
+HOMEPAGE="http://math-atlas.sourceforge.net/"
+SRC_URI="mirror://sourceforge/math-atlas/${PN}${PV}.tar.bz2
+	fortran? ( lapack? ( http://www.netlib.org/lapack/${LAPACKP} ) )"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
+IUSE="+deprecated doc fortran generic ifko lapack static-libs threads"
```
jlec: enable `fortran` and `lapack` by default

sci-libs/atlas/atlas-3.11.39.ebuild
```diff
@@ -0,0 +1,286 @@
...
+		!~X86_P4_CLOCKMOD
+		!~X86_INTEL_PSTATE
+	"
+	ERROR_KERNEL_X86_P4_CLOCKMOD="P4 Clockmod frequency scaling influences tuning and needs to be disabled at compile time."
+	ERROR_KERNEL_X86_INTEL_PSTATE="Intel Pstate frequency scaling influences tuning and needs to be disabled at compile time."
+	linux-info_pkg_setup
+}
+
+src_prepare() {
+	epatch \
```
jlec: Use `PATCHES=()` and drop the whole function.

sci-libs/atlas/atlas-3.11.39.ebuild
ngg: done

sci-libs/atlas/atlas-3.11.39.ebuild
ngg: done

sci-libs/atlas/atlas-3.11.39.ebuild
ngg: done

Issue #793, opened 2017-06-25
virtual/blas is satisfied by dev-cpp/eigen, but eigen doesn't provide blas libraries


rayhogenson:
I have installed virtual/blas and dev-cpp/eigen, satisfying all of the requirements for these packages, but now some packages, for example sci-libs/umfpack won't build, citing the error "configure: error: Cannot find blas libraries".

It does not appear to me that eigen is capable of providing blas, and unless I'm misunderstanding the situation (in which case can someone help me to figure out why these packages won't build), I believe eigen should be removed from the list of dependencies for virtual/blas so that it properly installs something like blas-reference in its place.

jlec:
Are you installing eigen from the overlay or main tree?

rayhogenson:
Both versions exhibit the same behavior.

kiwifb:
Question: Have you enabled the useflag "fortran" for eigen? That could be a mistake in the virtual since blas is a fortran library I am expecting the useflag to be required. However the flag is not present in the dependency in `virtual/blas`.

rayhogenson:
Emerging eigen 3.2.8-r2 from the overlay with the use flag fortran (the newer eigen 3.3.3 from the main tree has no such use flag) changes nothing, the same packages fail to build.

kiwifb:
Can you paste the log or post it somewhere?

rayhogenson:
I assume you mean the build log, here it is:
```
 * Package:    sci-libs/umfpack-5.6.2
 * Repository: gentoo
 * Maintainer: sci@gentoo.org
 * USE:        abi_x86_64 amd64 cholmod elibc_glibc kernel_linux userland_GNU
 * FEATURES:   compressdebug preserve-libs sandbox splitdebug userpriv usersandbox
>>> Unpacking source...
>>> Unpacking umfpack-5.6.2.tar.bz2 to /var/tmp/portage/sci-libs/umfpack-5.6.2/work
>>> Source unpacked in /var/tmp/portage/sci-libs/umfpack-5.6.2/work
>>> Preparing source in /var/tmp/portage/sci-libs/umfpack-5.6.2/work/umfpack-5.6.2 ...
 * Running elibtoolize in: umfpack-5.6.2/
 *   Applying target-nm/2.4.2 patch ...
 *   Applying ppc64le/2.4.2 patch ...
 * Running elibtoolize in: umfpack-5.6.2/config/
 *   Applying portage/1.2.0 patch ...
[...]

PR #795, opened 2017-07-03
virtual/blas: eigen needs the fortran useflag to provide blas.

kiwifb:
Package-Manager: portage-2.3.6


kiwifb:
As discussed in issue #793

PR #800, opened 2017-07-25
Openblas0.2.20

kiwifb:
Move the latest openblas available to the newly released 0.2.20. This new release includes all the patches  that were applied to 0.2.19 - including the big clang-3.9 patch stored on the mirror.
This new version also supports more (and newer) cpus out of the box, which has been a problem I have seen crop up on a regular basis on other forum.


kiwifb:
I limited the PR to move to 0.2.20. Is there any real valid reasons for us to continue carrying the older 0.2.14 and 0.2.15? Especially considering they require significant patching and won't support newer cpus?

kiwifb:
In case that's what holding this up, I have updated my GPG key to reflect my new address. So now the commits are all "verified".

kiwifb:
One commit has been included in the final push to release 0.2.20 that prevents openblas to build on platform not using glibc (OS X for example). See https://github.com/xianyi/OpenBLAS/issues/1258#issuecomment-320223121

I will update the PR with upstream fix for the problem. It will be included in the next release, which for some reason, is slow coming.

kiwifb:
Travis is not happy because of the new repoman which checks for https in `SRC_URI`. I have fixed it for openblas (all versions) in this PR but I haven't touched the other ebuilds for which it fails.

sci-libs/openblas/openblas-0.2.14.ebuild
```diff
@@ -108,12 +108,12 @@ src_compile() {
...
 		emake clean && emake libs shared ${openblas_flags}
-		mkdir -p libs && mv libopenblas* libs/ || die
+		mkdir -p libs && mv libopenblas* libs/ ||??die
 		# avoid pic when compiling static libraries, so re-compiling
 		if use static-libs; then
 			emake clean
 			emake libs ${openblas_flags} NO_SHARED=1 NEED_PIC=
-			mv libopenblas* libs/ || die
+			mv libopenblas* libs/ ||??die
 		fi
```
heroxbd: François, your editor has messed up `die` again.

sci-libs/openblas/openblas-0.2.20.ebuild
```diff
@@ -21,8 +21,6 @@ DEPEND="${RDEPEND}
 	virtual/pkgconfig"
 
 MULTILIB_WRAPPED_HEADERS=(
-	/usr/include/openblas/cblas.h
-	/usr/include/openblas/f77blas.h
 	/usr/include/openblas/openblas_config.h
```
heroxbd: Thanks, that's cleaner!  Could you please add a reference in the commit message to the solution of header problem in bugs.g.o?

sci-libs/openblas/files/openblas-0.2.19-MAKE.patch
```diff
@@ -1,29 +0,0 @@
...
- 	( cd $(NETLIB_LAPACK_DIR)/INSTALL; ./testlsame; ./testslamch; ./testdlamch; \
-         ./testsecond; ./testdsecnd; ./testieee; ./testversion )
-@@ -299,7 +299,7 @@ lapack-runtest:
- 
- blas-test:
- 	(cd $(NETLIB_LAPACK_DIR)/BLAS && rm -f x* *.out)
--	make -j 1 -C $(NETLIB_LAPACK_DIR) blas_testing
-+	$(MAKE) -j 1 -C $(NETLIB_LAPACK_DIR) blas_testing
- 	(cd $(NETLIB_LAPACK_DIR)/BLAS && cat *.out)
-
```
heroxbd: So.. this patch is for openblas-0.20, but the name openblas-0.2.19-MAKE.patch.  Does it break the openblas-0.19 ebuild?

virtual/blas/blas-2.1-r5.ebuild
```diff
@@ -24,7 +24,7 @@ RDEPEND="
 		abi_x86_64? (
 			!abi_x86_32? (
 				|| (
-					>=dev-cpp/eigen-3.1.4
+					>=dev-cpp/eigen-3.1.4[fortran]
 					sci-libs/atlas[fortran]
```
heroxbd: Is this from a different PR?

sci-libs/openblas/openblas-0.2.14.ebuild
kiwifb: You are right. This is quite incredible, is that space some kind of unicode?

sci-libs/openblas/files/openblas-0.2.19-MAKE.patch
kiwifb: You are misreading the commit. I remove the whole patch. In fact I remove 0.2.19 and replace it with 0.2.20.

kiwifb:
@heroxbd The header issue requires #802 to be merged first. Are you on top of that?

sci-libs/openblas/files/openblas-0.2.19-MAKE.patch
heroxbd: Yes, I realized that you have deleted openblas-0.2.19.  Could you please keep the openblas-0.2.19 for a while?

kiwifb: Sure. I will probably restart the branch from scratch.

kiwifb:
@heroxbd Are you going to do a sweep of the overlay to fix the https issue? If you do that, I'll wait for it to be done before recreating the branch.

sci-libs/openblas/files/openblas-0.2.19-MAKE.patch
heroxbd: Thanks again for your work!

sci-libs/openblas/openblas-0.2.14.ebuild
heroxbd: What editor are you using?  I will try to avoid it in the future ;)

kiwifb: I use a mix of kwrite and nano. My money is on kwrite doing something weird, but that implies that there is something special about that space - or its surroundings - to start with.

heroxbd:
> The header issue requires #802 to be merged first. Are you on top of that?

I am digesting it and learning from you.

> Are you going to do a sweep of the overlay to fix the https issue? If you do that, I'll wait for it to be done before recreating the branch.

Why is it important?  Does 301 redirection a big problem?

kiwifb:
the https things makes travis check fail for everyone. That's what I am concerned about. You cannot rely on travis as an extra check.

kiwifb:
Actually once #802 is merged (one way or another). I should fix all the openblas ebuilds in the overlay. Which means they will all get a revbump.

kiwifb:
Hum... I have redone the branch on the current master (minus #802 I had started before you merged it) but it still remembers the point at which I started the PR. Should I just do a new PR or will it do?

Issue #799, opened 2017-07-22
dev-cpp/eigen-3.3.3 version bump


devurandom:
It would be nice if Eigen could be bumped to 3.3, to bring it more in sync with the main Gentoo repository.

Eventually it would be great, if Science Overlay's Eigen could be merged back into Gentoo. That would help with a lot of dependency conflicts I usually need to sort out manually.

kiwifb:
I just looked at eigen 3.3 a little bit. I think there are still the bits to build blas, although it is not in the main documentation I have seen. What I have seen is that you can use a system blas/lapack in eigen. So there are 3 states possibles:
* plain not using blas/lapack and not providing it.
* providing blas
* using system blas/lapack
I don't think the last two are actually mutually exclusive technically (but sanity probably requires it).
Which states should we support?

devurandom:
if possible, all three? You could set `IUSE='blas system-blas'` and `REQUIRED_USE='blas? ( !system-blas ) system-blas? ( !blas )'`. `virtual/blas` would have to be adjusted accordingly (`RDEPEND='|| ( dev-cpp/eigen[blas] )'`). I'd suggest to start with the `USE=blas` case, since you already had that in previous versions.

kiwifb:
A bit like gsl (which has cblas-external), why not. I like the useflag `blas` but that means changing the current flags which is `fortran` (and which is clearly not obvious, changing it would be an improvement).

devurandom:
In hindsight: My suggestion for `system-blas` is not ideal. It is usually used to switch between bundled and system libraries, which is not what it would mean here. Logically it should probably be `IUSE=blas`, i.e. "I will use BLAS", and `IUSE=blas-provider`, which would mean "I provide BLAS"?

kiwifb:
Hum... yes. `blas` for building against blas is more consistent with the other use of the flag. `bls-provider`? Why not, most thing will be better than `fortran`.

kiwifb:
OK now that I have started working in earnest on this, I see things more clearly. I will keep the useflag `blas-provider` I think, because it is appropriate. 
Outside of blas and lapack (I haven't tested lapack, I'll stick to blas as in 3.2 to start with), eigen doesn't produce any libraries. It just installs headers.
[...]

PR #801, opened 2017-08-01
Bump to eigen 3.3.4 for issue #799

kiwifb:


kiwifb:
I am closing this. Since we are set to move to a new framework and there are some issues with this PR, I think it will be best to start fresh later.

PR #802, opened 2017-08-04
Multilib wrapping

kiwifb:
The numeric-int64-multibuild eclass do not currently allow the wrapping of header by ABI. This is because the wrapper function doesn't define the variable `MULTILIB_ABI_FLAG`, this variable is checked by `multilib_prepare_wrappers`, if it is not defined, wrappers are not created.
See 
https://bugs.gentoo.org/show_bug.cgi?id=617312#c11 and the next comment for more details.


kiwifb:
Anyone for this interesting bug?

heroxbd:
Committed.  Thanks for investigating this issue.

I suggest you fix your editor issues. It will save you more time in the future.  This time it has inserted some non-UTF8 characters.

> remote: To github.com:gentoo/sci.git
> remote:  + dd97f69f6...280823319 master -> master (forced update)
> remote: Traceback (most recent call last):
> remote:   File "hooks/post-receive.h02-notify-webhook", line 96, in <module>
> remote:     logmsg = git(['log','--reverse','--format=%an%x09%ae']).split("\n")[0]
> remote:   File "hooks/post-receive.h02-notify-webhook", line 20, in git
> remote:     details = details.decode("utf-8").strip()
> remote: UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe7 in position 81459: invalid continuation byte
>

PR #810, opened 2017-09-08
sci-libs/nipype: added dep version and fixed indentation

TheChymera:
Package-Manager: Portage-2.3.8, Repoman-2.3.3


sci-libs/nipype/nipype-9999.ebuild
```diff
@@ -20,7 +20,7 @@ IUSE="test"
 
 DEPEND="
 	dev-python/future[${PYTHON_USEDEP}]
-	dev-python/prov[${PYTHON_USEDEP}]
+	=dev-python/prov-1.5.0[${PYTHON_USEDEP}]
 	dev-python/numpy[${PYTHON_USEDEP}]
```
heroxbd: Could you please explain why prov-1.5.1 does not work with nipype-9999?

TheChymera: @heroxbd the package explicitly checks for it https://github.com/nipy/nipype/blob/4a0bd3d928f522e16a97526f88b0b9c608db9d34/requirements.txt#L9 I assume newer versions would exhibit undesirable but nonfatal behavior.

heroxbd: Ah-ha, I think we'd better document this dependence in the present version of nipype-0.13.1 instead of a -9999 moving target.

TheChymera: You're welcome to bump, but currently users rely on the 9999 version, and it has this dependency.

heroxbd: Do you mean almost all the users are using the 9999 version, by keywording themselves?  That is a strong reason to bump this package to the latest release.

At the same time, I am not in the position to bump this package, sorry.

TheChymera: It's ok, I don't know if there is any infrastructure to give us definitive quantitative information about usage. All users I know of, are collaborators of mine on some level, and we all need the live version as we work closely with upstream.

Issue #813, opened 2017-09-18
gsl-2.4 version bump


Telemin:
Can we get a version bump to gsl-2.4 please?  This version provides new routines for fixed quadrature integration and hermite polynomials amongst other things (http://git.savannah.gnu.org/cgit/gsl.git/tree/NEWS).

As far as I can tell on my own machines, there are no ebuild changes needed.

kiwifb:
Hum, documentation could be an issue I'll have to see if we were doing anything about it before. Happy that the way I patched previous gsl made it easy to go forward to 2.4 ^_^

kiwifb:
PR #815 is now open for review.

PR #815, opened 2017-09-21
sci-libs/gsl-2.4 version bump for issue #813

kiwifb:
I also fixed https in all ebuild to keep repoman happy. However I didn't migrate 1.16 to EAPI 6 or from autotools-multilib.

Controversial bit: the use of fma instructions leads to failures in the testsuite. So I explicitly disable them if they are available. I cannot just filter them since `-march=native` will enable them when available.


heroxbd:
Could you please forward thism `fma` issue upstream?

kiwifb:
I intend too, although I believe it to be fundamentally the same issue than something already opened since gsl-2.1 (IRC), it was reported for avx, which happens to enable fma where available as well. They probably can do something about the tolerance of that test. The tolerance code is incredibly nebulous to the newcomer otherwise I would have gone that way already.

heroxbd:
Which gsl bug are you referring to?  Let's point to the bug as a reference in the ebuild.

kiwifb:
I am fairly sure it is a related issue to https://savannah.gnu.org/bugs/?49697 but I only have one discrepancy I believe. It is otherwise very similar I believe.

kiwifb:
For `-mno-avx` I don't believe so. `-mno-fma` is more precise than `-mno-avx`. As in, you need avx to enable fma in the first place I think.

PR #816, opened 2017-09-25
net-misc/cernbox-client: Bump to 2.3.3 release, drop old.

olifre:
Following new owncloud-client version.

Package-Manager: Portage-2.3.10, Repoman-2.3.3


olifre:
In case somebody wonders, I just force-pushed some QA changes and a missing RDEP which have also been fixed recently in `net-misc/owncloud-client`.

olifre:
Obsoleted by #831 .

PR #821, opened 2017-10-13
sci-libs/arcas: new ebuild

TheChymera:
A really nice package for the automatic query of academic literature.

Package-Manager: Portage-2.3.8, Repoman-2.3.3


TheChymera:
Checks apparently fail due to `sci-mathematics/dolfin/dolfin-2017.1.0.ebuild` - nothing to do with this PR.

PR #825, opened 2017-10-31
sci-misc/openfst: Version bump

akreal:


akreal:
OpenFST bump is merged with Kaldi bump (#826) for better consistency.

PR #835, opened 2017-11-28
[DO NOT MERGE] blas/lapack eclasses

mgorny:
Per https://bugs.gentoo.org/632624. I'm creating a pull request to make it possible to comment easier.

### DIFF
--- a/eclass/blas.eclass
+++ b/eclass/blas.eclass
@@ -78,6 +78,23 @@
 # BLAS_USE_CBLAS=1
 # @CODE
 
+# @ECLASS-VARIABLE: BLAS_REQ_USE
+# @DESCRIPTION:
+# Adds a USE-flag requirement to every BLAS-implementation we depend on
+# (mainly useful for multilib, since the implementations share nearly no
+# other flags)
+# This must be set before the call to inherit blas
+#
+# Example usage:
+# @CODE
+# BLAS_REQ_USE="int64"
+# @CODE
+# 
+# Will result in DEPENDs like.
+# @CODE
+# DEPEND="blas_refblas? ( sci-libs/blas-reference[int64] )
+# @CODE
+
 
 
 BLAS_IMPLS=(refblas openblas gotoblas mkl)
@@ -137,15 +154,22 @@ function _blas_usedep(){
 function _blas_get_depends(){
 	local impl
 	local cblas
+	local requse
 	for impl in "${BLAS_SUPP_IMPLS[@]}"
 	do
+		if [[ $BLAS_REQ_USE ]]
+		then
+			requse="[${BLAS_REQ_USE}]"
+		else
+			requse=""
+		fi
 		if [[ $BLAS_USE_CBLAS ]]
 		then
 			eval "cblas=\$_cblas_provider_$impl"
 		else
 			cblas=""
 		fi
-		eval "echo \"$(_blas_useflag_by_impl $impl)? ( \$_blas_provider_$impl $cblas )\""
+		eval "echo \"$(_blas_useflag_by_impl $impl)? ( \${_blas_provider_$impl}${requse} ${cblas}${requse} )\""
 	done
 }
 
--- a/eclass/blas.eclass
+++ b/eclass/blas.eclass
@@ -169,7 +169,11 @@ function _blas_get_depends(){
 		else
 			cblas=""
 		fi
-		eval "echo \"$(_blas_useflag_by_impl $impl)? ( \${_blas_provider_$impl}${requse} ${cblas}${requse} )\""
+		if [ -n "$cblas" ]
+		then
+			cblas="${cblas}${requse}"
+		fi
+		eval "echo \"$(_blas_useflag_by_impl $impl)? ( \${_blas_provider_$impl}${requse} ${cblas} )\""
 	done
 }
 
@@ -204,7 +208,6 @@ function _blas_set_globals(){
 	BLAS_USEDEP="$(_blas_usedep)"
 	BLAS_DEPS="$(_blas_get_depends)"
 	BLAS_REQUIRED_USE="^^ ( $(_blas_useflag_by_impl "${BLAS_SUPP_IMPLS[@]}") )"
-
 	if [[ ${BLAS_CONDITIONAL_FLAG} ]]
 	then
 		REQUIRED_USE="${BLAS_CONDITIONAL_FLAG}? ( ${BLAS_REQUIRED_USE} )"
--- a/eclass/blas.eclass
+++ b/eclass/blas.eclass
@@ -106,8 +106,8 @@ function _blas_set_globals(){
 	
 	BLAS_USEDEP="$(_blas_usedep)"
 	BLAS_DEPS="$(_blas_get_depends)"
-	BLAS_REQ_USE="^^ ( $(_blas_useflag_by_impl "${BLAS_SUPP_IMPLS[@]}") )"
-	
+	BLAS_REQUIRED_USE="^^ ( $(_blas_useflag_by_impl "${BLAS_SUPP_IMPLS[@]}") )"
+
 }
 
 _blas_set_globals
--- a/eclass/blas.eclass
+++ b/eclass/blas.eclass
@@ -15,6 +15,11 @@ _blas_provider_openblas="sci-libs/openblas"
 _blas_provider_gotoblas="sci-libs/gotoblas2"
 _blas_provider_mkl="sci-libs/mkl"
 
+_cblas_provider_refblas="sci-libs/cblas-reference"
+_cblas_provider_openblas=""
+_cblas_provider_gotoblas=""
+_cblas_provider_mkl=""
+
 function _blas_impl_valid(){
 	local impl
 	for impl in "${BLAS_IMPLS[@]}"
@@ -58,9 +63,16 @@ function _blas_usedep(){
 
 function _blas_get_depends(){
 	local impl
+	local cblas
 	for impl in "${BLAS_SUPP_IMPLS[@]}"
 	do
-		eval "echo \"$(_blas_useflag_by_impl $impl)? ( \$_blas_provider_$impl )\""
+		if [[ $BLAS_USE_CBLAS ]]
+		then
+			eval "cblas=\$_cblas_provider_$impl"
+		else
+			cblas=""
+		fi
+		eval "echo \"$(_blas_useflag_by_impl $impl)? ( \$_blas_provider_$impl $cblas )\""
 	done
 }
 
--- a/eclass/blas.eclass
+++ b/eclass/blas.eclass
@@ -1,0 +1,104 @@
+# @ECLASS: blas.eclass
+# @MAINTAINER:
+# schmidom@student.ethz.ch
+# @BLURB: Selects a proper BLAS implementation to build the package
+# @DESCRIPTION:
+# Here a BLAS-implementation is chosen from the cutset of available,
+# compatible and implementations in the USE-flags
+# The implementation is then enforced upon every dependency of the package
+
+BLAS_IMPLS=(refblas openblas gotoblas mkl)
+BLAS_SUPP_IMPLS=()
+
+_blas_provider_refblas="sci-libs/blas-reference"
+_blas_provider_openblas="sci-libs/openblas"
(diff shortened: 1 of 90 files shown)


eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+# This variable contains a list of BLAS implementations the package
+# supports. It has to be an array. Either it or BLAS_COMPAT_ALL must be
+# set before the `inherit' call. BLAS_COMPAT_ALL overrides BLAS_COMPAT.
+#
+# Example:
+# @CODE
+# BLAS_COMPAT=( refblas openblas gotoblas mkl )
+# @CODE
+
+# @ECLASS-VARIABLE: BLAS_COMPAT_ALL
```
mgorny: 1. Is there a reason to have two variables? The 'standard' practice would be to allow all if `BLAS_COMPAT` is not set. However…
2. Allowing 'all' is problematic by design (this is why python-r1 doesn't do it). It means that whenever a new implementation is supported by the eclass, all packages instantly start depending on it. This means that before adding a new implementation, you need to keyword it on all relevant architectures and test all reverse dependencies.
3. How frequent are BLAS/LAPACK incompatibilities?

eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+# @REQUIRED
```
mgorny: Marking it `@REQUIRED` is confusing since it's only conditionally required.

eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+# RDEPEND="sci-libs/foo[${BLAS_USEDEP}]"
+# @CODE
+#
+# Example value:
+# @CODE
+# refblas?,openblas?
+# @CODE
+
+
+# @ECLASS-VARIABLE: BLAS_CONDITIONAL_FLAG
```
mgorny: 1. This is a horrible trap, and you already fell in it. Again, python-r1 doesn't do stuff like this for a good reason. You add the basic form and it works for 95% of the packages. Then you add more complexity and it works for 95% of the rest. Then you…
2. Implicit OR is totally confusing. What if someone needs AND instead?

eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+# @CODE
+#
+# Then the eclasses REQUIRED_USE and DEPENDs look like this:
+# @CODE
+# DEPEND="blas? ( blas_refblas? ( sci-libs/blas-reference ))"
+# RDEPEND="${DEPEND}"
+# REQUIRED_USE="blas? ( ^^ ( blas_refblas blas_openblas ) )"
+# @CODE
+
+# @ECLASS-VARIABLE: BLAS_USE_CBLAS
```
mgorny: Do you always use CBLAS along with BLAS, or is there a use case for CBLAS alone?

eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+# @CODE
+# DEPEND="blas_refblas? ( sci-libs/blas-reference[int64] )
+# @CODE
+
+
+
+BLAS_IMPLS=(refblas openblas gotoblas mkl)
+BLAS_SUPP_IMPLS=()
+
+_blas_provider_refblas="sci-libs/blas-reference"
```
mgorny: Is this split actually beneficial? Usually the only result of such variables is that people have to scroll the code back and forth to figure things out.

eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+_blas_provider_openblas="sci-libs/openblas"
+_blas_provider_gotoblas="sci-libs/gotoblas2"
+_blas_provider_mkl="sci-libs/mkl"
+
+_cblas_provider_refblas="sci-libs/cblas-reference"
+_cblas_provider_openblas=""
+_cblas_provider_gotoblas=""
+_cblas_provider_mkl=""
+
+function _blas_impl_valid(){
```
mgorny: nit: don't use `function`, and add a space between `()` and `{`.

eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+	local impl
+	for impl in "${BLAS_IMPLS[@]}"
```
mgorny: nit: we generally do `for ...; do` on a single line, so you may want to do that for consistency.

eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+	do
+		if [ "$1" == "$impl" ]
```
mgorny: nit: don't use non-bash `[ ... ]` form, always use full bash `[[ ... ]]` which doesn't require quoting (except for very special cases that are not relevant here). Also always use `${var}` instead of `$var`.

eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+		then
+			return 0
+		fi
+	done
+	return 1
```
mgorny: This whole thing seems equivalent to `has "${1}" "${BLAS_IMPLS[@]}"`

eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+	
+	die "No matching implementation found"
+}
+
+function _blas_useflag_by_impl(){
+	echo "${@/#/blas_}"
+}
+
+function _blas_name_by_impl(){
+	eval "echo \"\${_blas_name_$1}\""
```
mgorny: Don't ever use `eval`. You're probably looking for `${!foo}` syntax, except you really want to remove the completely needless indirection, and e.g. use `switch` instead.

eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+}
+
+function _blas_usedep(){
+	local tmp
+	tmp=("$(_blas_useflag_by_impl "${BLAS_SUPP_IMPLS[@]/%/?}" )" )
```
mgorny: The variable inside the array is quoted, so this array will always have one element. In other words, you can drop the outer `()` for the same effect.

eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+	echo "${tmp// /,}"
+}
+
+function _blas_get_depends(){
+	local impl
```
mgorny: You can list multiple variables in a single `local` call.

eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+		BLAS_SUPP_IMPLS=( "${BLAS_IMPLS[@]}" )
+	else
+		if [ "${#BLAS_COMPAT[@]}" -eq 0 ]
+		then
+			die "No BLAS implementations set in BLAS_COMPAT"
+		fi
+		
+		for impl in "${BLAS_COMPAT[@]}"
+		do
+			if [ "$impl" == "*" ]
```
mgorny: Now I'm lost. I don't think this was documented.

eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+			then
+			
+				break
+			elif _blas_impl_valid "$impl"
+			then
+				BLAS_SUPP_IMPLS+=( "$impl" )
```
mgorny: Why add them one-by-one? If one is invalid, you fail anyway, so you can as well just check them, and then assign all.

eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+	DEPEND="${RDEPEND}"
+}
+
+_blas_set_globals
+unset -f _blas_set_globals
+
+
+function blas_pkg_setup(){
+	BLAS_IMPL="$(_blas_best_impl)"
+	mkdir "${T}/pkg-config/"
```
mgorny: All external (system) commands need `|| die` to account for errors.

eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+	local NORMAL_PKGCONFIG="${EROOT}/usr/lib/pkgconfig"
```
mgorny: This is certainly wrong and doesn't account for multilib. Not to mention I don't think it's ever necessary.

eclass/blas.eclass
```diff
@@ -0,0 +1,241 @@
...
+	ln -s "${EROOT}/usr/lib/pkgconfig/$BLAS_IMPL.pc" "${T}/pkg-config/blas.pc"
```
mgorny: This doesn't handle multilib.

eclass/blas.eclass
kiwifb: Incompatibilities are few and far between. You could possibly have trouble with intel MKL because it does more than just provide blas/lapack. It provides FFTW and gsl like functionalities. But so long has you are interested in the published blas/lapack functions defined in the reference package there shouldn't be incompatibilities.

This is why I suggested we follow the multilib model rather than the python model.

eclass/blas.eclass
kiwifb: I am not sure what you mean here. But usually you one or the other. BLAS is a fortran interface and CBLAS is a C interface. For historical reason the fortran interface is the one most used and some implementation like openblas combine the two in one fat library.

kiwifb:
Apart from the lack of multilib support mentioned by @mgorny we should aim to support the int64 variant for packages that wants it or support it. Right the following implementations can provide int64: blas-reference, lapack-reference, openblas, MKL. Not too sure about cblas-reference.
Packages wanting int64: julia IIRC.

Doeme:
Oh wow, thank you very much for the thorough review @mgorny @kiwifb .
Unfortunately, this is an old version of the eclasses, the new ones have a pull request here: https://github.com/gentoo/sci/pull/837

I probably should have announced that more clearly, sorry about that.
I will go through the comments and point out stuff that has been fixed and correct the things in the other branch that haven't.

eclass/blas.eclass
Doeme: @mgorny 
1. has been fixed in 
2. "All" means that the program only requires the common functions of BLAS, i.e. everything in refblas. A package that doesn't provide these core-functions is no BLAS implementation.
3. Some implementations have extra functions like openblas_set_num_threads (or similar). If a program uses those, it is incompatible with the other implementations

@kiwifb If a program uses these functions, the eclass probably has to depend on MKL directly.

eclass/blas.eclass
Doeme: What would be the proper way to handle it? define a BLAS_DEPEND that the eclass-dev then can add to DEPEND themselves?

eclass/blas.eclass
Doeme: You're probably right. There was an implementation once that needed the selected implementations to be sorted in a predefined order, there it made sense, but not any more.

eclass/blas.eclass
Doeme: Another option would be to use associative arrays, but they're only supported in >=bash-4.0, I believe

eclass/blas.eclass
Doeme: The ${EROOT}?
I think it is necessary for cross-emerge, which I occasionally do.

eclass/blas.eclass
Doeme: This changed in the new version https://github.com/gentoo/sci/pull/837
The new *_COMPAT syntax is described here https://github.com/gentoo/sci/pull/837/files#diff-7f6f012dc8db266ba1e7b5f7a931734cR14

Doeme:
Sorry to be getting on your nerves again, but I would like to have the three questions answered in order to continue working on the eclasses:

* https://github.com/gentoo/sci/pull/835#discussion_r155259879
* https://github.com/gentoo/sci/pull/835#discussion_r155254885 (i.e. is it OK to use associative arrays?)
* https://github.com/gentoo/sci/pull/835#discussion_r155252969

Additionally, I'd like to mention that, as you might have guessed, I'm doing this in the scope of a semester project at my university. The project ends with this year, and it's highly unlikely that I will find the motivation to further this work in my spare time.

Here be highlights @mgorny @kiwifb @TheChymera

heroxbd:
Thanks guys.  So as in 2018, @Doeme are you going to work on it or leave it half-done?

Doeme:
I have a few non-pushed modifications (usage of assoc arrays) and a few questions unanswered. I'm not unwilling to work on it a bit more, but it now has a rather low priority in my project schedule.

PR #836, opened 2017-11-29
Fix libmed multilib-strict compilation issue.

robertmbaker:
This solution provides a patch to change the appropriate entries in two CMakeLists.txt files to install library files into the appropriate architecture library directory.

The upstream package uses a hard-coded default of "lib" the patch changes this to ${CMAKE_INSTALL_FULL_LIBDIR} which is a macro provided in GNUInstallDirs which the patch also adds include statements to pull in.

Finally the patch has been added to the libmed ebuild.

### DIFF
--- a/sci-libs/libmed/files/libmed-3.1.0-cmake-libdir.patch
+++ b/sci-libs/libmed/files/libmed-3.1.0-cmake-libdir.patch
@@ -1,0 +1,95 @@
+diff -Naur med-3.1.0_SRC/src/CMakeLists.txt med-3.1.0_SRC-patched/src/CMakeLists.txt
+--- med-3.1.0_SRC/src/CMakeLists.txt	2015-09-07 11:24:04.000000000 -0500
++++ med-3.1.0_SRC-patched/src/CMakeLists.txt	2017-11-28 22:54:19.677235182 -0600
+@@ -19,6 +19,8 @@
+ INCLUDE_DIRECTORIES(${PROJECT_BINARY_DIR}/include 
+     ${PROJECT_SOURCE_DIR}/include)
+ 
++INCLUDE(GNUInstallDirs)
++
+ # Get pure C intermediary targets:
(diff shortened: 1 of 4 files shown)


kiwifb:
Looks good to me. Do we need a revbump for one of these? It changes the record of where the files are installed if not the actual location.

robertmbaker:
Per the guidelines here:
https://devmanual.gentoo.org/general-concepts/ebuild-revisions/index.html

We can probably do this without a revision bump since users who managed to build and install this previously will not run into the issue without attempting a rebuild of the package.

mrkwjc:
Seems that sci-mathematics/dolfin suffers similar 'multilib-strict' issue...

robertmbaker:
With my current build sci-mathematics/dolfin won't compile unless I unmask the 2017.1.0 version, so I didn't bother looking at the earlier CMakeLists.txt files, but so long as nothing major changed there across versions the fix should be the same.

See https://github.com/gentoo/sci/pull/838 for a solution to the same issue on sci-mathematics/dolfin.

robertmbaker:
Forgive me, I accidentally included the changes for sci-mathematics/dolfin in this PR too. I don't think I can undo that with my permissions, perhaps a project collaborator can.

TheChymera:
@robertmbaker would you be interested in providing a more up-to-date version of this? Regarding dolfin, it has recently been removed due to fatal QA issues in its dependency stack: https://gitweb.gentoo.org/proj/sci.git/commit/?id=d51c2ced7b3d8950894746f66da1c07712a00feb

Should you be interested in bringing that back as well, I would be happy to help, and it could be pretty quick.

TheChymera:
@robertmbaker feel free to reopen or resubmit if this becomes relevant again.

waebbl:
There's a more up-to-date ebuild in ::gentoo as sci-libs/med

PR #838, opened 2017-12-01
Fix dolfin lib dir for multilib-strict.

robertmbaker:
This package also came up while discussing multilib-strict issues for libmed.

The issue is largely the same with this package. Library defaulting to install in /usr/lib. The solution is only slightly different because upstream sets a variable to "lib" and the "install(TARGET" function calls that variable. So I chose to leave the "instalI(TARGET" call alone, and set the variable it uses to ${CMAKE_INSTALL_FULL_LIBDIR}.

Again I added an include statement to CMakeLists.txt for GNUInstallDirs to gain access to the ${CMAKE_INSTALL_FULL_LIBDIR} macro.


TheChymera:
@robertmbaker we have temporarily suspended dolfin from Gentoo Science as per a recent cleanup effort. If this package is still of interest, we would be happy to bring it back pending a PR for a recent version.

PR #839, opened 2017-12-07
forge-libdir patch

the-mozart:


the-mozart:
Remove the patch please. It doesn't transfer right directory to the CMake confgi dirs/

PR #845, opened 2018-01-05
Rutils.eclass

cloc3:
R modules, such as Rcpp (http://rcpp.org/) are not present in portage, but must be installed directly using R commands.

a new eclass may be useful to help portage to correctly install these modules.

this class may also be useful to correctly build packages depending from R.
for example, actually sci-physics/root fails to build against R extra-use flag if RInside and Rcpp are not manually installed.

these patch propose a way how to easily solve this problem.


cloc3:
found a repoman issue.
I'try to reopen thie pull request later.

Issue #867, opened 2018-04-22
sci-biology/kat: need help with calling setup.py


mmokrejs:
Hi,
  the kat package uses configure/make approach in general but also contains setup.py to install python file. Although I import `distutils-r1` something is still wrong with the ebuild. Please have a look into https://github.com/gentoo/sci/blob/master/sci-biology/kat/kat-8888.ebuild and clean it. Thank you.

mmokrejs:
Right, thank you.

[build.log.txt](https://github.com/gentoo/sci/files/1936191/build.log.txt)

mmokrejs:
It registered with only one python version for some reason. Usually multiple are supported. It should have recognized 3.5 and 3.6 per PYTHON_COMPAT=( python3_{5,6} ) in the *.ebuild files.

kiwifb:
I need to take a proper look at the source code.

kiwifb:
Hum... I almost feel like this should be split into two. But we can certainly keep it as one ebuild with a bit of thought. There may be a difficulty in testing the python bits. So this is a autotools+python build system where usually only one python is supported. This is a autotools limitation because the autotools macros AX_PYTHON* were never meant to support more than one python.

However it looks like all the python bits are independent and can be shipped separately. Beurk double options for build and install are not useful in the build system.

I do not know why dependency to python-3.5+ is not ensured. Someone else needs to check your system. You appear to be running some sort of prefix so I am not sure what is happening there.

First step, pass `--disable-pykat --disable-pykat-install` so that we don't get autotools pollution in the script directory. I'll start looking seriously at the ebuild side.

kiwifb:
Underlinked libraries? Really? Parallel make on doc broken? Sure... 
I cannot test on multiple python3 because I only have 3.6 right now. But what I have, does what I expect it to do.

You want all that in a PR? I am only doing 2.4.1 - I leave it to you to migrate the fix to 9999.

kiwifb:
Trying to run the tests. Needs sys-process/time for the jellyfish ones. Interesting bug for kat. In the `test.sh_sect.sh` test there seems to be a limit on the size of the argument. If the command line ends up being too long we have failure with an unhelpful message (unsupported format).

mmokrejs:
Thanks @kiwifb for all your work. Please edit the ebuild right away, no need for PR. The -9999 ebuild differs from -2.4.1 just in the `SRC_URI` line and `git-r3` import. On my normal Gentoo install on a laptop I have these:

```
[...]

PR #868, opened 2018-04-23
sci-biology/kat: various fixes for issue #867

kiwifb:
Package-Manager: Portage-2.3.24, Repoman-2.3.6


mmokrejs:
Thanks for your work, actually I realized I have no rights over https://github.com/gentoo/sci some maybe @jlec or @SoapGentoo would help here? The only thing I could is to take the patch and commit it straight into git+ssh://git@git.gentoo.org/proj/sci.git . I shall better wait for others, it probably won't take too long anyway.

mmokrejs:
Here is at least output from `MAKEOPTS=-j1 FEATURES=noclean emerge =sci-biology/kat-9999`.

[build.log.txt](https://github.com/gentoo/sci/files/1957899/build.log.txt)

mmokrejs:
Updating output from MAKEOPTS=-j1 FEATURES=noclean emerge =sci-biology/kat-9999.

[build.log.txt](https://github.com/gentoo/sci/files/1957976/build.log.txt)

and

`equery files =sci-biology/kat-9999 > /tmp/files.txt`

[files.txt](https://github.com/gentoo/sci/files/1957978/files.txt)

kiwifb:
OK, so there is still only one python installed for you. I will have to get a minimal python3.5 install to test this myself. Anyone has documentation on fixing the rpath?

kiwifb:
Yes rpath problem still present in pure gentoo. Now let's give some attention to your multi-python problem. I have a feeling it will require a split or some very complicated gymnastics. Can you replace `default` in `src_prepare` by `distutils-r1_src_preprare` and see if it improves things.

mmokrejs:
> Can you replace default in src_prepare by distutils-r1_src_prepare and see if it improves things.

I do not think so.

[build.log.txt](https://github.com/gentoo/sci/files/1958238/build.log.txt)

mmokrejs:
@kiwifb Please check whether https://github.com/gentoo/sci/commit/f373141b6dd59db5eaf1bce4104e5ba6085958ac is correct/complete. I tried to squash and rebase a few times, resolve merge conflicts, I wish I did not screw up this. The `DEPPEND` was I guess a typo. Why did not `repoman` catch that?

kiwifb:
Was a typo, and repoman should check that `DEPEND` is set. It doesn't care about a random variable it doesn't know anything about. Spotting that it is a typo is hard unless you put some kind of matching engine behind.

The only wrong with the overlay right now is that there are patches for kat-2.4.1 when there is no kat-2.4.1 ebuild anymore and the patches are not used by any other ebuilds.

mmokrejs:
But the patches **are** used by 2.4.2 and -9999. Yes, I could have renamed them, true. Yes, the old patches I dropped but I thought that was because of my tree being outdated, not because of your branch.

mmokrejs:
Aha, I wanted to say two patches are still used (probably due to me screwing the merges they stay with the old filenames). In manual resolving of the failed merge I had to bring the patch file lines back in ebuilds. I will rename them on disk, I thought you objected not even the two remaining patches should be used anymore.

kiwifb:
Yeah, I thought about redo-ing the branch from scratch. May be I should have :)

mmokrejs:
Aha, I am really sleepy, sorry. Now I see what you said. There are still extra the '2.4.1' versions. Removed in 2ae549fb69309b8ab8537b3e9a8ea158d375cf1d .

PR #869, opened 2018-05-05
sci-mathematics/scilab: Fix compilation w/ gcc-8

vaeth:
The argument arrays in subroutine tridv were larger than needed and larger
than passed when called.


TheChymera:
@vaeth sorry for the extremely late reply. Gentoo Science has suffered from a lack of proper access for maintainership during the past 2 years. GCC 8 is now no longer current, and the package still doesn't build, in addition to this version being deprecated.

Would you be interested in submitting a PR for a more current version (latest is `6.1.0` https://github.com/scilab/scilab/releases). We how have the manpower to review this in a timely fashion.

vaeth:
Sorry, meanwhile I have hardly any free time, and in foreseeable future no need for scilab anymore.

TheChymera:
@vaeth I fully understand ;) should you spontaneously take an interest, we've made good progress on addressing new PRs quickly, so it should run smoothly.

PR #872, opened 2018-06-22
sci-libs/openblas: version bump to 0.3.0

kiwifb:
Package-Manager: Portage-2.3.40, Repoman-2.3.9
Notable ebuild changes:
* replaced the sed replacement in src_prepare by standard variables in openblas_flags
* Added NO_STATIC=1 for building shared library (may not be necessary)
* removed TARGET=GENERIC for the dynamic use flag. It actually prevents some targets to compile properly.


Kagami:
:+1:
Might want to update PR to [0.3.1](https://github.com/xianyi/OpenBLAS/releases/tag/v0.3.1).

kiwifb:
Thanks for the heads up. I may just shutter that one and make a new one. Hopefully someone will eventually pick it up.

PR #876, opened 2018-08-01
sci-libs/openblas: version bump to 0.3.2

kiwifb:
Package-Manager: Portage-2.3.40, Repoman-2.3.9


xhub:
Version 0.3.3 is available

kiwifb:
I know, I'll update the PR when I have a bit of time.

leezu:
Besidse incrementing the version number, what needs to be done before this can be merged? I incremented the version number in my fork and successfully compiled/installed openblas 0.3.3 via this ebuild. Compiling numpy against the new version of openblas was also successful, so I wonder if there is anything missing? Thanks for your work on this @kiwifb .

kiwifb:
I am sure it just works increasing the version number. While we can just do that for now, there are tweaks that would be nice to have that should be done at some point regarding generic hardware support - especially for older hardware.

As for merging, well it seems merging is on hiatus unless you provide a serious bug fix.

kiwifb:
Simple version bump.

grafi-tt:
This PR also fixes <https://bugs.gentoo.org/584534>.

I bought Ryzen 2 and noticed OpenBLAS 0.3.2 or later is required. Whereas I will work around using my own overlay, merging the PR certainly helps many users.

stefantalpalaru:
openblas-0.3.5 is available in my overlay (simple rename): https://github.com/stefantalpalaru/gentoo-overlay

kiwifb:
For the record there is one such ebuild as well in the sage-on-gentoo overlay, and has been for a while since no one is doing anything here. Multiplication of ebuilds in third party overlays is no good.

stefantalpalaru:
> Multiplication of ebuilds in third party overlays is no good.

It's how we cope with lazy devs ;-)

kiwifb:
Said devs will blame me for refusing to join for years when I obviously have what it takes. So that's not a good argument I could make. For your information RL can be tough on dev time.

heroxbd:
> > Multiplication of ebuilds in third party overlays is no good.
> 
> It's how we cope with lazy devs ;-)

Hi @stefantalpalaru, sorry for being lazy.  Are you interested submitting some patches and PR to make sci overlay more up-to-date and bug-free?

heroxbd:
Hi, as you know.  We are migrating OpenBLAS to the gentoo main repository.  Would you please give that a test? https://bugs.gentoo.org/584534#c15

I am closing this PR as it is obsolete.

stefantalpalaru:
> We are migrating OpenBLAS to the gentoo main repository. Would you please give that a test?

That's nice, but you've added 0.3.5 to the main repo while I have 0.3.6 in my overlay (released on April 29, added on May 1st).

https://github.com/xianyi/OpenBLAS/releases

kiwifb:
Sorry for the long silence. We need 0.3.6, it fixes issues with gcc-9.1.0.

kiwifb:
I am looking at the ebuild in the main tree and I cannot see any provision for int64 builds. Is it a problem in the current framework? Julia favor int64 blas if my memory serves me right, I don't think that it is something we should leave behind. MKL would become the sole provider of that kind of capabilities.

heroxbd:
@cdluminate could you please share you plans for eselect-ldso of blas and lapack int64?

kiwifb:
@heroxbd It seems like 0.3.6 is a trivial bump from 0.3.5. For my sanity can a space be inserted in 
```
emake -Cinterface shared-blas-lapack $(openblas_flags)
```
make understand `-C` separately from `interface` but it still spooks me.

cdluminate:
> @cdluminate could you please share you plans for eselect-ldso of blas and lapack int64?

It would be problematic to provide int64 ABI/API through `libblas.so` or `liblapack.so` since that would lead to a mess with the int32 ABI/API. Adding the int64 USE flags to existing packages will make it hard to write ebuilds for users who want e.g. both int32 and int64 openblas at the same time. My personal recommendation is to provide a new set of packages, say `eselect-{blas64,lapack64}` + `sci-libs/lapack64`, which are clearly isolated with the int32 ones. These new packages should change SONAMEs into say `libblas64.so` and `liblapack64.so`. This is the same as what I'm doing for Debian's 64bit-indexing BLAS/LAPACK packages.

kiwifb:
Sounds fair to me. Especially considering that not all packages do int64, so different virtuals.

cdluminate:
Ok, let me bump openblas::gentoo first. Afterwards I'll try to find a better solution for the in64 variants.

heroxbd:
> > @cdluminate could you please share you plans for eselect-ldso of blas and lapack int64?
> 
> It would be problematic to provide int64 ABI/API through `libblas.so` or `liblapack.so` since that would lead to a mess with the int32 ABI/API. Adding the int64 USE flags to existing packages will make it hard to write ebuilds for users who want e.g. both int32 and int64 openblas at the same time. My personal recommendation is to provide a new set of packages, say `eselect-{blas64,lapack64}` + `sci-libs/lapack64`, which are clearly isolated with the int32 ones. These new packages should change SONAMEs into say `libblas64.so` and `liblapack64.so`. This is the same as what I'm doing for Debian's 64bit-indexing BLAS/LAPACK packages.

I agree with the overall plan, but one thing.  We can make int64 as a use flag and by enabling it we install too sets of libraries: libblas.so and libblas64.so.  What do you think?

kiwifb:
I know why you want that @heroxbd - you want to avoid ebuild duplication. On the face of it it is less ebuilds to maintain in exchange of more complex ebuilds. However I think it would still be wise to have at least separate virtuals - that means dependencies on int32 and int64 are completely separate and allow a different set of implementations without impinging on each other.

Whether to go for doubling the ebuilds or making them more complex is up to the maintainer. And we have to deal with the headers in any case.

cdluminate:
I agree with @kiwifb . At least we need `virtual/{blas64,cblas64,lapack64,lapacke64}`. For the implementations we could introduce two new USE flags: `int32` and `int64`, where each of them could trigger a separate full build, and `int32` is enabled by default. Same for the eselect packages.

PR #887, opened 2018-10-01
sci-biology/fsl: bump to 5.0.10 (now 5.0.11)

TheChymera:
Package-Manager: Portage-2.3.50, Repoman-2.3.11

I am trying to bump FSL to 5.0.10 . This is particularly relevant for dependency integration (e.g. fslpy) with fsleyes (the new FSL viewer).

One patch needed updating, but the changes were mostly line placement.

However the [build fails](http://bpaste.net/show/ad1d785cd6ab).

I am a bit unsure about the snippet, it seems to fail after it cannot `rm`, but the error message does not seem to relate to that and instead cites being unable to build a number of functions:

``` 
make[4]: Leaving directory '/var/tmp/portage/sci-biology/fsl-5.0.10/work/fsl/src/mist-clean'
make[3]: Leaving directory '/var/tmp/portage/sci-biology/fsl-5.0.10/work/fsl/src/mist-clean'
make[2]: Leaving directory '/var/tmp/portage/sci-biology/fsl-5.0.10/work/fsl/src/mist-clean'
make[1]: Leaving directory '/var/tmp/portage/sci-biology/fsl-5.0.10/work/fsl/src/mist-clean'
Makefile:55: warning: overriding recipe for target 'clean'
/var/tmp/portage/sci-biology/fsl-5.0.10/work/fsl/config/common/rules.mk:32: warning: ignoring old recipe for target 'clean'
rm common/gibbsshapemodel.o common/mrfshapemodel.o common/mvnshapemodel.o common/plotting.o common/profilefilters.o common/profilemixtures.o common/profilemodel.o common/profilepriors.o common/serialisation.o common/shape.o common/shapemodel.o common/stats.o common/transformation.o mist/mist.o mist/mist
rm: cannot remove 'common/gibbsshapemodel.o': No such file or directory
rm: cannot remove 'common/mrfshapemodel.o': No such file or directory
rm: cannot remove 'common/mvnshapemodel.o': No such file or directory
rm: cannot remove 'common/plotting.o': No such file or directory
rm: cannot remove 'common/profilefilters.o': No such file or directory
rm: cannot remove 'common/profilemixtures.o': No such file or directory
rm: cannot remove 'common/profilemodel.o': No such file or directory
rm: cannot remove 'common/profilepriors.o': No such file or directory
rm: cannot remove 'common/serialisation.o': No such file or directory
rm: cannot remove 'common/shape.o': No such file or directory
rm: cannot remove 'common/shapemodel.o': No such file or directory
[...]

PR #889, opened 2018-10-13
sci-libs/ciftilib: new ebuild

TheChymera:
Package-Manager: Portage-2.3.49, Repoman-2.3.10

hopefully helping with https://github.com/gentoo/sci/pull/887

but still failing with: http://bpaste.net/show/dc8fe9f41bd0


kiwifb:
I am not sure why the cmake module to find libxml++ breaks from reading it. I will have to experiment for myself. I would have gone the qt5-core (the only qt component needed) route to be honest.

TheChymera:
As mentioned in the other thread, upstream seems to like libxml++, and as long as it works for them I don't think we need to worry. Personally, I feel qt is wayy too large to pull in and certainly not already present on headless systems (and also probably not present on all desktop systems except the subset of them which uses KDE).

kiwifb:
OK but the current ebuild will have to do something more. At the moment if QT5-core is present it will be used regardless. You would have to set `IGNORE_QT` to completely do without it.

Anyway if things are packaged properly upstream shouldn't care which back end is used. It will solve itself "magically". So I wouldn't just ignore QT but either put it as an option or at least having a `||` block in DEPEND. `qtcore` is not that big anyway.

TheChymera:
In what phase would I set IGNORE_QT?

kiwifb:
In `src_configure` you have to read the cmake-utils eclass page to set the switch to FALSE/OFF when cmake is called.

TheChymera:
@kiwifb I wasn't able to find any examples fo this. From what little I can deduce from the eclass, I tried:

```
src_configure() {
    local mycmakeargs=(
        IGNORE_QT="FALSE"
    )
    cmake-utils_src_configure
}
```

and

```
src_configure() {
    IGNORE_QT="FALSE"
    cmake-utils_src_configure
}
```

And I go the following log for both: http://bpaste.net/show/23b61148e446

kiwifb:
`-DIGNORE_QT=FALSE`. Example ebuild sci-libs/vtk and sci-visualization/paraview. First option please. I can see you never manually used cmake. I have to say the most useful stuff in the eclass helps when you have useflags but you are electing to just disable a path.

TheChymera:
@kiwifb with the newetst commit I still get this:

/var/tmp/portage/sci-libs/ciftilib-1.5.3/work/CiftiLib-1.5.3

:(

but doesn't it look like the error is from libxml++ rather than QT?

kiwifb:
Sorry I wasn't implying that there currently was a problem with qtcore, just that if it was installed it would automagically be found and used before doing anything about libxml++. Why the libxml++ detection fails is a bit of a mystery to me at this stage.

kiwifb:
I notice that thos ebuild doesn't specify libxmlpp:2.6. Have you by any chance switched to libxmlpp:3.0? And removed the 2.6 slot? The package does require 2.6 as far as I can see.
https://github.com/Washington-University/CiftiLib/blob/master/cmake/Modules/Findlibxml%2B%2B.cmake#L14
If you only have libxmlpp:3.0 that won't work.

TheChymera:
ah, ok, in addition to having pasted the path to my local log instead of the URL I seem to have messed up the deps. Yes, the libxmlpp slot was the issue.

It now works. Any idea if the QT deselection also works? Can you test the most recent commit out?

kiwifb:
If you show me a configure log, I'll have a good idea if QT deselection works by comparing with before.

TheChymera:
@kiwifb To me it still looks like it's actively trying to find Qt and moves on only when it does not.... http://bpaste.net/show/79e3dad868ef

kiwifb:
Correct you shouldn't see this. I'll do a couple of experiments locally.

kiwifb:
This is embarrassing. I got the variable setting in reverse. You want to ignore QT so it should not be set to FALSE but either TRUE or ON.

TheChymera:
yeah, dear god, I also copy-pasted like a robot XD

kiwifb:
Humpf. Tests all passed in a manual build. I will have to see from the ebuild. There are a few minor versions differences because I am on arch while it looks like you are on ~arch.

kiwifb:
Interesting. Slightly different for me 
```
83% tests passed, 6 tests failed out of 35

Total Test time (real) =   0.65 sec

The following tests FAILED:
	  5 - rewrite-big-md5-Conte69.MyelinAndCorrThickness.32k_fs_LR.dscalar.nii (Failed)
	 12 - rewrite-big-md5-Conte69.MyelinAndCorrThickness.32k_fs_LR.dtseries.nii (Failed)
	 19 - rewrite-big-md5-Conte69.MyelinAndCorrThickness.32k_fs_LR.ptseries.nii (Failed)
	 26 - rewrite-big-md5-Conte69.parcellations_VGD11b.32k_fs_LR.dlabel.nii (Failed)
	 33 - rewrite-big-md5-ones.dscalar.nii (Failed)
	 35 - datatype-md5-ones.dscalar.nii (Failed)
``` 
One less failure.

kiwifb:
Testsuite doesn't pass when run in parallel add the following block
```
src_test(){
	local myctestargs=(
                -j1
        )
	cmake-utils_src_test
}
```

TheChymera:
@kiwifb Thank you for your help. Any other advice before I merge this and we move back to bumping FSL?

kiwifb:
Nope.

TheChymera:
resolved by 6eb5168dec6ae536a582c35586f3d0587089c422

PR #895, opened 2018-10-21
sci-biology/fsl: upstream bump

kiwifb:
Package-Manager: Portage-2.3.49, Repoman-2.3.11


kiwifb:
Alternative to #887 includes all my current fixes:
* work with overlay ciftilib compiled with either qtcore or libxmlpp
* leverage latest overlay ciftilib to overcome incompatibility with internal nifti_io
* unbundle most of what can be
* fix parallel build issues in fslsurface
* fix language issue
* fix double definitions between sub-projects

kiwifb:
I think we should with this now. Unbundling `fslpy` certainly can be done later. And we have a working fsl ebuild in the meantime.

I can have a look at the rest but cannot help about merging.

TheChymera:
Ok, merged c7f45b539af87c6ba4fe535181d5d4afa25baa66

PR #910, opened 2019-04-04
sys-cluster/openmpi: specify external libraries and version bump

swordencao:
* Version 4.0.* have both `hwloc` and `libevent` external libraries support, which find and use the external libraries by default (https://github.com/open-mpi/ompi/issues/5031). However, the new implementations for configuring search for `lib64` and use it as the library directory which is not correct and incompatible with `multilib.eclass`.  
https://github.com/open-mpi/ompi/blob/master/opal/mca/hwloc/external/configure.m4
https://github.com/open-mpi/ompi/blob/master/opal/mca/event/external/configure.m4  
Patch by specifying external libraries explicitly.

* Version bumps to 4.0.1.

Closes: https://bugs.gentoo.org/671510
Package-Manager: Portage-2.3.62, Repoman-2.3.12
Signed-off-by: Jian Cao sworden.cao@gmail.com


jsbronder:
Awesome work!  Only a couple of lines changed but I know it was a lot to debug.  I changed the spaces to tabs in the ebuild and merged.
`9b561f5c7 sys-cluster/openmpi: specify external libraries and version bump`

PR #937, opened 2019-12-31
dev-python/nbdime: add PYTHON_COMPAT version 3.6

y0rune:
Signed-off-by: Marcin Wozniak y0rune@aol.com


TheChymera:
nbdime is now no longer provided in Gentoo Science, but only in Gentoo Main. Ideally we would like to avoid shadowing packages, unless absolutely necessary. If you need a live ebuild, please consider submitting a PR to https://github.com/gentoo/gentoo

PR #941, opened 2020-02-10
PyTorch and Torchvision packages

aclex:
Here's a merge of the current state of PyTorch, Torchvision and some other required packages ebuilds as in development at https://github.com/aclex/pytorch-ebuild. I've been suggested to make a pull request to merge it here, to join the efforts of the development, so I've decided to try.

Could you please take a look at the ebuilds and consider them for maybe merge to your repository? Please, let me know, if something should be fixed for that. Thanks in advance!


app-doc/pytorch-docs/pytorch-docs-1.3.1.ebuild
```diff
@@ -0,0 +1,54 @@
+# Copyright 2019-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python2_7 python3_{5,6,7} )
```
heroxbd: We are going rid of Python 2.7 and 3.5. Please drop them.  The same applies to all other ebuilds.

app-doc/pytorch-docs/pytorch-docs-1.3.1.ebuild
```diff
@@ -0,0 +1,54 @@
...
+"
+
+PATCHES=(
+	"${FILESDIR}/0001-Don-t-prerender-TeX-parts.patch"
+	"${FILESDIR}/0002-Use-MAKE-environment-variable-instead-of-make.patch"
+)
+
+src_unpack() {
+	unpack ${A}
+	mv "${WORKDIR}/${PYTORCH_NAME}-${PV}" "${S}"
```
heroxbd: `mv` is an external command.  Should append with `|| die`.

app-doc/pytorch-docs/pytorch-docs-1.3.1.ebuild
```diff
@@ -0,0 +1,54 @@
...
+}
+
+src_compile() {
+	local doc_build_dir="${S}/docs"
+	cd "${doc_build_dir}"
```
heroxbd: || die

app-doc/pytorch-docs/pytorch-docs-1.3.1.ebuild
```diff
@@ -0,0 +1,54 @@
...
+	emake html-stable
+}
+
+src_install() {
+	local doc_build_dir="${S}/docs"
+
+	mkdir -p "${D}/usr/share/doc/${P}"
```
heroxbd: `|| die`

app-doc/pytorch-docs/pytorch-docs-1.3.1.ebuild
```diff
@@ -0,0 +1,54 @@
...
+	cp -a "${doc_build_dir}/build/html" "${D}/usr/share/doc/${P}/html"
```
heroxbd: `|| die`.  You get the idea.

dev-python/javalang/javalang-0.12.0.ebuild
```diff
@@ -0,0 +1,23 @@
...
+
+inherit distutils-r1
+
+DESCRIPTION="Pure Python Java parser and tools"
+HOMEPAGE="https://github.com/c2nes/javalang"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~mips ~ppc ~ppc64 ~s390 ~sparc ~x86 ~amd64-linux ~x86-linux ~x64-solaris"
```
heroxbd: Only add keywords that you have tested.

dev-python/javalang/javalang-0.12.0.ebuild
```diff
@@ -0,0 +1,23 @@
...
+
+RDEPEND="
+	dev-python/six"
+
+DEPEND="
+	${RDEPEND}
+	${PDEPEND}
```
heroxbd: Where is PDEPEND?

dev-python/javalang/javalang-0.12.0.ebuild
```diff
@@ -0,0 +1,23 @@
...
+DESCRIPTION="Pure Python Java parser and tools"
+HOMEPAGE="https://github.com/c2nes/javalang"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~mips ~ppc ~ppc64 ~s390 ~sparc ~x86 ~amd64-linux ~x86-linux ~x64-solaris"
+
+RDEPEND="
+	dev-python/six"
```
heroxbd: After dropping Python 2.7, `six` will be unnecessary.

dev-python/javasphinx/javasphinx-0.9.15.ebuild
```diff
@@ -0,0 +1,32 @@
...
+HOMEPAGE="https://github.com/bronto/javasphinx"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~mips ~ppc ~ppc64 ~s390 ~sparc ~x86 ~amd64-linux ~x86-linux ~x64-solaris"
+
+RDEPEND="
+	>=dev-python/javalang-0.10.1
+	dev-python/lxml
```
heroxbd: All the Python dependencies should have the same Python version USE flags.  cf. https://wiki.gentoo.org/wiki/Project:Python/distutils-r1#PYTHON_USEDEP

The same applies to all others.

sci-libs/pytorch/pytorch-1.3.1.ebuild
```diff
@@ -0,0 +1,246 @@
...
+	"${FILESDIR}/0008-Fix-include-directory-variable-of-rocThrust.patch"
+	"${FILESDIR}/0009-Prefer-lib64-for-some-ROCm-packages.patch"
+	"${FILESDIR}/0010-Remove-conversion-ambiguity-in-ternary-operators.patch"
+)
+
+src_prepare() {
+	cmake-utils_src_prepare
+
+	if use rocm; then
+		python "${S}/tools/amd_build/build_amd.py"
```
heroxbd: `${EPYTHON}` cf. https://wiki.gentoo.org/wiki/Project:Python/python-utils-r1#Python_build_environment

sci-libs/pytorch/pytorch-1.3.1.ebuild
```diff
@@ -0,0 +1,246 @@
...
+		libgloo.a
+		libshm.so
+		libsleef.a
+		libcaffe2_detectron_ops.so
+		libtorch_python.so
+	)
+
+	for file in ${multilib_failing_files[@]}; do
+		mv -f "${D}/usr/lib/$file" "${D}/usr/lib64"
+	done
```
heroxbd: This works.  But we'd better fix the build system.

sci-libs/pytorch/pytorch-1.3.1.ebuild
```diff
@@ -0,0 +1,246 @@
...
+		rm -rfv "${D}/usr/include/hip"
+	fi
+
+	cp -rv "${WORKDIR}/${P}/third_party/pybind11/include/pybind11" "${D}/usr/include/"
+
+	rm -fv "${D}/usr/lib64/libtbb.so"
+	rm -rfv "${D}/usr/lib64/cmake"
+
+	rm -rfv "${D}/usr/share/doc/mkldnn"
+
```
heroxbd: Patching the build system would be better and cleaner.

heroxbd: But we can improve it gradually if your time is limited.

heroxbd: Same applies to 1.4.0.

sci-libs/torchvision/torchvision-0.4.2.ebuild
```diff
@@ -0,0 +1,85 @@
...
+	unpack ${A}
+	mv "${WORKDIR}/${P//torch/}" "${S}"
+}
+
+src_configure() {
+	cmake-utils_src_configure
+
+	if use python; then
+		FORCE_CUDA=$(usex cuda 1 0) \
+		CUDA_HOME=$(usex cuda ${CUDA_HOME} "") \
```
heroxbd: Where is `CUDA_HOME` set?

aclex:
@heroxbd thank you very much for this hard work of reviewing all these ebuilds, Benda! Many thanks for the detailed comments and references! I'll try to fix them all and update the request as soon, as possible.

heroxbd:
> @heroxbd thank you very much for this hard work of reviewing all these ebuilds, Benda! Many thanks for the detailed comments and references! I'll try to fix them all and update the request as soon, as possible.

Hi Alex, are you working on this?  If you are busy, I can implement what I have commented.

aclex:
@heroxbd Hi, Benda! Yes, unfortunately I didn't have a chance to work on it last weekend, so it goes slower, than I expected. I hope to submit the fixes soon, maybe this evening, except for the things related to the build system (those artifacts which are moved or removed manually). As for them, I tried to do it the proper way, but they seem to be quite tricky both to implement and test, especially given the upstream strongly discourage any changes to the building configuration.

Sorry again for the delay, I hope to return with the results soon.

heroxbd:
> @heroxbd Hi, Benda! Yes, unfortunately I didn't have a chance to work on it last weekend, so it goes slower, than I expected. I hope to submit the fixes soon, maybe this evening, except for the things related to the build system (those artifacts which are moved or removed manually). As for them, I tried to do it the proper way, but they seem to be quite tricky both to implement and test, especially given the upstream strongly discourage any changes to the building configuration.
> 
> Sorry again for the delay, I hope to return with the results soon.

No problem, we can work together on this.  You can update what you have and I can continue from there to make a working version.  Then we can keep improving it.

app-doc/pytorch-docs/pytorch-docs-1.3.1.ebuild
aclex: Yes, I also planned this change after reading the news. Now fixed.

app-doc/pytorch-docs/pytorch-docs-1.3.1.ebuild
aclex: Thanks for pointing out on this! Fixed.

app-doc/pytorch-docs/pytorch-docs-1.3.1.ebuild
aclex: Fixed.

app-doc/pytorch-docs/pytorch-docs-1.3.1.ebuild
aclex: Fixed.

app-doc/pytorch-docs/pytorch-docs-1.3.1.ebuild
aclex: Fixed.

dev-python/javalang/javalang-0.12.0.ebuild
aclex: Change it to the most tested range. The package seems to be pure Python, so is obviously safe on the platforms.

dev-python/javalang/javalang-0.12.0.ebuild
aclex: Slipped in from somewhere :) Fixed.

dev-python/javalang/javalang-0.12.0.ebuild
aclex: Yes, thank you, removed it.

dev-python/javasphinx/javasphinx-0.9.15.ebuild
aclex: I'd like to make some more thorough tests on this, as I seem to remove it after it led to some failures. But not sure why. Will test it and sort it out.

sci-libs/pytorch/pytorch-1.3.1.ebuild
aclex: Fixed.

sci-libs/pytorch/pytorch-1.3.1.ebuild
aclex: Unfortunately, this needs some more experiments and tests, as we kind of go against the upstream here.

sci-libs/pytorch/pytorch-1.3.1.ebuild
aclex: Agree, but not changed yet.

sci-libs/torchvision/torchvision-0.4.2.ebuild
aclex: I've just put it following the expectations in the `setup.py`, but it seems to require some more environment preparation, like here: https://github.com/pytorch/vision/blob/master/packaging/pkg_helpers.bash

aclex:
@heroxbd Benda, so I've pushed the most trivial portion of fixes for now, so I'll continue to look at the remaining ones, though it's going to take some more time. Please, feel free to make any changes you like, no problem with it at all. Still I don't really know, how it can be done within the pull request, but I think I can share the request branch with you, if it's needed.

dev-python/javasphinx/javasphinx-0.9.15.ebuild
aclex: Tested and fixed it as well in the affected ebuilds.

aclex:
I also would like to apologize for my commit messages, which, as I see, don't tally well with the history of the repository. I tried to follow the rules in recent messages, though it's not always easy, when the commit affect several ebuilds at once. Anyway, feel free to correct me.

sci-libs/torchvision/torchvision-0.4.2.ebuild
heroxbd: I see.

sci-libs/pytorch/pytorch-1.4.0.ebuild
```diff
@@ -0,0 +1,253 @@
...
+		-DUSE_ZMQ=$(usex zeromq ON OFF)
+		-DUSE_MPI=$(usex mpi ON OFF)
+		-DUSE_GLOO=$(usex gloo ON OFF)
+		-DBUILD_NAMEDTENSOR=$(usex namedtensor ON OFF)
+		-DBLAS=${blas}
+		-DBUILDING_SYSTEM_WIDE=ON # to remove insecure DT_RUNPATH header
+	)
+
+	local CC="clang"
+	local CXX="clang++"
```
heroxbd: Why is clang required?  Is it because of ROCm?

aclex: No, not really, it's because of the code in version 1.4.0, in `c10`, if I remember correctly (haven't saved the failure log for some reason), which interferes with the [regression in GCC 9.1 and 9.2](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90538). It's fixed in 9.3 and 10 and doesn't reproduce in 8.x and Clang, hence the change. It can be more fine-grained, allowing safe GCC versions, but require detecting compiler version in ebuild, which I'm not so comfortable with.

heroxbd: I will revert this.  We cannot switch to clang unconditionally.

aclex: Then it's probably better not to include ebuild for version 1.4.0 for now, as otherwise it would lead to compilation failure, as soon as GCC-9 is selected as default compiler (which is the default for modern profiles).

heroxbd:
Hi, I have committed only pytorch CPU version to the tree.  We still need to add more features like ROCm and CUDA, besides torchvision, etc.  Please rebase against the newest tree and send atomic pull requests.  

This one is too big to handle smoothly.   And please use `repoman` for the ebuild commits.

aclex:
OK, thank you. At first glance it looks quite deeply modified, so I'll try to look closer at it later.

sci-libs/pytorch/pytorch-1.4.0.ebuild
heroxbd: I can't reproduce this with `gcc-9.2.0-r4` though.

aclex: It's fairly possible, especially given, that, as I can see, the merged version of the ebuild have little similar parts to one in this request. I think, I have `gcc-9.2.0-r2`, the latest stable version, but haven't tested building the merged version, yet.

aclex:
Benda, I've looked through the commit you've mentioned, and, to be honest, quite disappointed, as it's silently pretty much another version of what I've put in this request. It's written in the commit as if authored by me, but there're hardly a few lines I wrote actually) To me it looks kind of contradictory to your suggestion of adding things atomically. Perhaps, it might have more sense, if we elaborated some joint version together somewhere externally, with all the little changes and fixes gradually accumulated from both versions. But OK, that's the way it is.

I'm actually quite confused about the next steps on this, as I don't know your plan, what you didn't included as wasteful or unnecessary and what was just postponed for future. As you wrote, ROCm, torchvision and other things were excluded for now, so apparently that's pretty much it. Anyway, I'll backport the fixes on your suggestions, so please feel free to pick any other parts you consider worth merging, I'd be glad to help you with that.

My initial plan after your suggestion was to merge all the tree fully here and render my branch to read-only, but, I think, I just underestimated the amount of efforts and pain of the process, following all the packaging rules, given there's a lot of work to be done in the ebuild itself to keep it stable enough after each version bump, while free time on this is very limited. So, I think, it would be the best way to continue it separately. I'm going to concentrate on ROCm support now, as there're major issues there. Again, I'm open for any interaction, to not do the same work twice etc.

Anyway, thank you very much again for your help, suggestions and guidance! Was useful experience for me.

heroxbd:
> Benda, I've looked through the commit you've mentioned, and, to be honest, quite disappointed, as it's silently pretty much another version of what I've put in this request. It's written in the commit as if authored by me, but there're hardly a few lines I wrote actually) 

Not really, the CPU version is based on your `ebuild`, with the exception of `ROCm` and `SRC_URI` parts.  The biggest point of your contribution is the `cmake` driven build and related patches, which is buried underneath `setup.py`.  I put you on the author on purpose, because you have more code than me.

I am sorry if that confused you.

> To me it looks kind of contradictory to your suggestion of adding things atomically. 

By atomically, I mean putting in CPU, AMD, nVidia versions one at a time. And when it comes to AMD, land the ROCm ebuilds first and then an update to the pytorch ebuild.  Finally the reverse dependencies of pytorch.

> Perhaps, it might have more sense, if we elaborated some joint version together somewhere externally, with all the little changes and fixes gradually accumulated from both versions. But OK, that's the way it is.

The science overlay itself is an working place for the gentoo main tree.  Ultimately every ebuild that matured in science overlay should land into gentoo.  So let's polish the joint version on science overlay.

> I'm actually quite confused about the next steps on this, as I don't know your plan, what you didn't included as wasteful or unnecessary and what was just postponed for future. As you wrote, ROCm, torchvision and other things were excluded for now, so apparently that's pretty much it. Anyway, I'll backport the fixes on your suggestions, so please feel free to pick any other parts you consider worth merging, I'd be glad to help you with that.

Please see the atomic comment above.

> My initial plan after your suggestion was to merge all the tree fully here and render my branch to read-only, but, I think, I just underestimated the amount of efforts and pain of the process, following all the packaging rules, given there's a lot of work to be done in the ebuild itself to keep it stable enough after each version bump, while free time on this is very limited. So, I think, it would be the best way to continue it separately. I'm going to concentrate on ROCm support now, as there're major issues there. Again, I'm open for any interaction, to not do the same work twice etc.

Thank you!  I am interested in ROCm ebuilds as well.  If you could revise them according to be the standards, I am more than happy to land them to the science overlay.

I understand there is a steep learning curve for packaging.  Requiring a quality standard does not seem to make much sense as long as it `just works`, but the quality standards are accumulated wisdom and proved best efforts in the history.  They will pay off in the long run. 

> Anyway, thank you very much again for your help, suggestions and guidance! Was useful experience for me.

You are welcome.  I apologize again for not synchronizing the big picture of pytorch/ROCm with you promptly.

aclex:
Thank you for the reply and clarification of your view, Benda!

> The science overlay itself is an working place for the gentoo main tree. Ultimately every ebuild that matures in science overlay should land into gentoo. So let's polish the joint version on science overlay.

This is so attractive aim, especially in case of PyTorch, as, to my knowledge, there's no version of it in repositories of many other distributions.

The main problem is that the recommended building process just ignores the system-wide installation, FHS, building separated from the installation and all that stuff and very clearly discourage anyone to try doing these things. I don't think, that's for bad will, but obviously rather because of complex build configuration historically fused of several different packages. To make it system compliant we actually need to violate its rules, which usually leads to quite subtle problems. In that case, incremental steps are very useful and valuable for the same reasons we do separate commits instead of one squashed. For example, I carefully selected all the header files in the image, checking if they are included in some Torch public headers or not, to not break the compilation of C++ part and packages which depend on it (like, as I remember, `torchvision`). As I can see, at least `pybind11` headers are commented out from installation in the submitted version, so it's unclear, if it works in these scenarios at all. But it's already there. Such things could easily be avoided, if we tried to observe all the proposals together, discussing each one separately. Two heads are always better, than one, they say)

> …when it comes to AMD, land the ROCm ebuilds first and then an update to the pytorch ebuild. Finally the reverse dependencies of pytorch.

The ROCm-enabled build relies on the ebuilds in the separate repository: https://github.com/justxi/rocm/. As I know, packages from there are accepted to the main Gentoo repository directly, but that's not finished process, about a half of the packages are not yet moved. My PyTorch ebuild just mentions the dependencies from there and implies this repository has been registered beforehand, when `rocm` USE flag enabled. Effectively this means it depends on non-existent packages, so if it's not appropriate case for SCI repository, the ROCm-enabled configuration would have to wait for all the ROCm packages added to the main Gentoo repository.

> I understand there is a steep learning curve for packaging. Requiring a quality standard does not seem to make much sense as long as it just works, but the quality standards are accumulated wisdom and proven best efforts in the history. They will pay off in the long run.

I surely agree with you on the rules importance and always try to respect and follow the rules I'm aware of. Not a good thing is just when they are quite spread and there're a lot of moments that all checks are passed and you thing it's all good, while there are so much more requirements violated. I mean, I just want to share my results with everyone, to save them from doing the same work, making it easy to use as much, as possible. I have no power to keep everyone happy with every single of my commit messages :)

heroxbd:
> Thank you for the reply and clarification of your view, Benda!
> 
> > The science overlay itself is an working place for the gentoo main tree. Ultimately every ebuild that matures in science overlay should land into gentoo. So let's polish the joint version on science overlay.
> 
> This is so attractive aim, especially in case of PyTorch, as, to my knowledge, there's no version of it in repositories of many other distributions.

Indeed.

> The main problem is that the recommended building process just ignores the system-wide installation, FHS, building separated from the installation and all that stuff and very clearly discourage anyone to try doing these things. I don't think, that's for bad will, but obviously rather because of complex build configuration historically fused of several different packages. To make it system compliant we actually need to violate its rules, which usually leads to quite subtle problems. In that case, incremental steps are very useful and valuable for the same reasons we do separate commits instead of one squashed. For example, I carefully selected all the header files in the image, checking if they are included in some Torch public headers or not, to not break the compilation of C++ part and packages which depend on it (like, as I remember, `torchvision`). As I can see, at least `pybind11` headers are commented out from installation in the submitted version, so it's unclear, if it works in these scenarios at all. But it's already there. Such things could easily be avoided, if we tried to observe all the proposals together, discussing each one separately. Two heads are always better, than one, they say)

That's a very smart move to use PyPI package to figure out which headers to keep.

We are now using `pybind11` from the system, not bundled as far as I remember.

> 
> > …when it comes to AMD, land the ROCm ebuilds first and then an update to the pytorch ebuild. Finally the reverse dependencies of pytorch.
> 
> The ROCm-enabled build relies on the ebuilds in the separate repository: https://github.com/justxi/rocm/. As I know, packages from there are accepted to the main Gentoo repository directly, but that's not finished process, about a half of the packages are not yet moved. My PyTorch ebuild just mentions the dependencies from there and implies this repository has been registered beforehand, when `rocm` USE flag enabled. Effectively this means it depends on non-existent packages, so if it's not appropriate case for SCI repository, the ROCm-enabled configuration would have to wait for all the ROCm packages added to the main Gentoo repository.

Good to know of the ongoing ROCm efforts.

> > I understand there is a steep learning curve for packaging. Requiring a quality standard does not seem to make much sense as long as it just works, but the quality standards are accumulated wisdom and proven best efforts in the history. They will pay off in the long run.
> 
> I surely agree with you on the rules importance and always try to respect and follow the rules I'm aware of. Not a good thing is just when they are quite spread and there're a lot of moments that all checks are passed and you thing it's all good, while there are so much more requirements violated. I mean, I just want to share my results with everyone, to save them from doing the same work, making it easy to use as much, as possible. I have no power to keep everyone happy with every single of my commit messages :)

Sure, no problem at all.

aclex:
> We are now using pybind11 from the system, not bundled as far as I remember.

Ah, indeed, haven't noticed it's available in the system. Then that's obviously the best option, and I am probably wrong here, hopefully it's going to work then.

> That's a very smart move to use PyPI package to figure out which headers to keep.

Unfortunately, it's not a source reliable enough, as e.g. for C++ development [it's not even mentioned](https://pytorch.org/cppdocs/installing.html), with preference to the official ZIP binary archive unpacked locally. Another problem is PyPI packages are not always proved to be absolutely consistent, for example, CPU version of Torchvision goes with box operations linked to non-extistent CUDA library, leading to runtime linker failure, for several versions in a row already. Nobody seems to care yet :) So yes, I try to check all that cases known to me manually.

heroxbd:
Alexey Chernov <notifications@github.com> writes:

>>  We are now using pybind11 from the system, not bundled as far as I
>>  remember.
>
> Ah, indeed, haven't noticed it's available in the system. Then that's
> obviously the best option, and I am probably wrong here, hopefully
> it's going to work then.

It works as far as I can remember.

>> That's a very smart move to use PyPI package to figure out which
>> headers to keep.
>
> Unfortunately, it's not a source reliable enough, as e.g. for C++
> development it's not even mentioned, with preference to the official
> ZIP binary archive unpacked locally. Another problem is PyPI packages
> are not always proved to be absolutely consistent, for example, CPU
> version of Torchvision goes with box operations linked to
> non-extistent CUDA library, leading to runtime linker failure, for
> several versions in a row already. Nobody seems to care yet :) So yes,
> I try to check all that cases known to me manually.

I can see it is a great endeavour to do so.  Thank you!

epsilon-0:
@heroxbd @aclex , is this still being worked on?
I wanted to get pytorch from the central system in my cluster as opposed to letting my lab users install it manually for themselves, mostly for performance and consistency reasons.
I'm hoping this can be added soonish, thanks a lot for your work :)

TheChymera:
@epsilon-0 the package has kindly been provided by @heroxbd and @aclex , witht the current version in the overlay superseding the one in this PR.

The package works well, other than it will not build with the `cuda` USE flag in any of the labs I am in contact with. Help would very much be appreciated on that front https://bugs.gentoo.org/742617

PR #949, opened 2020-03-20
sci-libs/lib2bit: fix testing and qa warnings

jiegec:
Package-Manager: Portage-2.3.88, Repoman-2.3.18
Signed-off-by: Jiajie Chen <noc@jiegec.ac.cn>

### DIFF
--- a/sci-libs/lib2bit/lib2bit-0.2.1.ebuild
+++ b/sci-libs/lib2bit/lib2bit-0.2.1.ebuild
@@ -1,13 +1,13 @@
-# Copyright 1999-2017 Gentoo Foundation
+# Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=7
 
 inherit eutils
 
 DESCRIPTION="C library for accessing 2bit files"
 HOMEPAGE="https://github.com/dpryan79/lib2bit"
-SRC_URI="https://github.com/dpryan79/lib2bit/archive/0.2.1.tar.gz -> ${P}.tar.gz"
+SRC_URI="https://github.com/dpryan79/lib2bit/archive/${PV}.tar.gz -> ${P}.tar.gz"
 
 LICENSE="MIT"
 SLOT="0"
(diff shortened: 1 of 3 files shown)


sci-libs/lib2bit/lib2bit-0.2.1.ebuild
```diff
@@ -17,15 +17,20 @@ IUSE="static"
...
 	default
-	epatch "${FILESDIR}"/${P}_respect_DESTDIR.patch
-	sed -e 's#/usr/local#/usr#' -i Makefile || die
 }
 
 src_install(){
-	emake DESTDIR="${ED}" install
-	if not use static; then
-		rm "${ED}"/usr/lib/lib2bit.a || die
+	emake
```
heroxbd: I doubt it work for Prefix.

sci-libs/lib2bit/lib2bit-0.2.1.ebuild
```diff
@@ -17,15 +17,20 @@ IUSE="static"
...
+PATCHES=(
+	"${FILESDIR}"/${P}_respect_LDFLAGS.patch
+	"${FILESDIR}"/${P}_test_support_python3.patch
+)
+
 src_prepare(){
 	default
-	epatch "${FILESDIR}"/${P}_respect_DESTDIR.patch
-	sed -e 's#/usr/local#/usr#' -i Makefile || die
 }
```
heroxbd: Remove src_prepare when there is only `default` left in the function.

sci-libs/lib2bit/lib2bit-0.2.1.ebuild
```diff
@@ -17,15 +17,20 @@ IUSE="static"
...
 
+PATCHES=(
+	"${FILESDIR}"/${P}_respect_LDFLAGS.patch
+	"${FILESDIR}"/${P}_test_support_python3.patch
+)
+
 src_prepare(){
 	default
-	epatch "${FILESDIR}"/${P}_respect_DESTDIR.patch
-	sed -e 's#/usr/local#/usr#' -i Makefile || die
```
heroxbd: Is it taken care of elsewhere?

sci-libs/lib2bit/lib2bit-0.2.1.ebuild
jiegec: `dolib` and `doheader` are used. If they work for Prefix, this would as well.

sci-libs/lib2bit/lib2bit-0.2.1.ebuild
jiegec: We no longer use the make install target, but rather install the libraries and headers by ourselves.

sci-libs/lib2bit/lib2bit-0.2.1.ebuild
jiegec: Done in a new commit.

PR #966, opened 2020-08-24
sci-biology/open-ephys-gui: new package

malfatti:
Hi!

Open-ephys Plugin-GUI is a GUI for processing, recording, and visualizing multichannel electrophysiological data, very relevant for experimental neuroscience. This commit add ebuilds to install open-ephys gui master (9999) and development (99999999) versions. The optional plugins are added via USE flags.

Cheers :)


TheChymera:
@malfatti thank you for another nice contribution :)
Could you please add a static version as well?

I know it is tempting for the user-developer to just go with 9999 (most of my early packages started like that as well, and seamless support for live versions was the reason I switched all our lab machines to Gentoo) — but these versions tend to be very unstable and unsustainable, particularly if you do not have push access to the overlay which distributes them, and can't fix build instructions immediately when you spot them.

We can keep them ofc, but at least one static version on top of that would really be best.

malfatti:
Hi!

Again, sorry for the delay. Totally agree. Added the last static version, and fixed the use of the `jack` flag (it was not actually compiling with jack support).

sci-biology/open-ephys-gui/open-ephys-gui-9999.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	SubDir=${P}
+else
+    SRC_URI="https://github.com/open-ephys/plugin-GUI/archive/v${PV}.tar.gz"
+	Suffix=${PV}
+	SubDir="plugin-GUI-${PV}"
+	S="${WORKDIR}/$SubDir"
+fi
+
+SLOT="${PV}"
+KEYWORDS="~amd64 ~x86"
```
TheChymera, line 31: please remove keywords from 9999 and above

sci-biology/open-ephys-gui/open-ephys-gui-99999999.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	SubDir=${P}
+else
+    SRC_URI="https://github.com/open-ephys/plugin-GUI/archive/v${PV}.tar.gz"
+	Suffix=${PV}
+	SubDir="plugin-GUI-${PV}"
+	S="${WORKDIR}/$SubDir"
+fi
+
+SLOT="${PV}"
+KEYWORDS="~amd64 ~x86"
```
TheChymera, line 31: please remove keywords from 9999 and above

sci-biology/open-ephys-gui/open-ephys-gui-0.5.2.2.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	SubDir="plugin-GUI-${PV}"
+	S="${WORKDIR}/$SubDir"
+fi
+
+SLOT="${PV}"
+KEYWORDS="~amd64 ~x86"
+IUSE="jack +KWIKFormat +NetworkEvents +NWBFormat"
+
+BDEPEND="
+	<sys-devel/gcc-9
```
TheChymera, line 35: can we not have nice things? :( it's ok, if we can't, but are you sure that this constraint is necessary

TheChymera (changes requested):
minor changes and newer GCC question.

sci-biology/open-ephys-gui/open-ephys-gui-9999.ebuild
malfatti: Done

sci-biology/open-ephys-gui/open-ephys-gui-99999999.ebuild
malfatti: Done

sci-biology/open-ephys-gui/open-ephys-gui-0.5.2.2.ebuild
malfatti: Open-ephys depends on JUCE, which is bundled upstream. The current version bundled does not appear to support >gcc-8 (at least I cannot compile it in my machine unless I switch to gcc-8). I use and update this software regularly and I always try to compile with the newest gcc, so I'll for sure drop this restriction as soon as I'm able to compile it with a newer version.

TheChymera: ok :)

sci-biology/open-ephys-gui/open-ephys-gui-0.5.2.2.ebuild
```diff
@@ -0,0 +1,65 @@
...
+inherit cmake
+
+DESCRIPTION="Processing, recording, and visualizing multichannel ephys data"
+HOMEPAGE="http://www.open-ephys.org/gui/"
+LICENSE="GPL-3"
+
+if [[ ${PV} == "9999" ]] ; then
+	EGIT_REPO_URI="https://github.com/open-ephys/plugin-GUI"
+	EGIT_BRANCH="master"
+    inherit git-r3
```
TheChymera, line 14: make sure indents are always tabs, here they're spaces

sci-biology/open-ephys-gui/open-ephys-gui-0.5.2.2.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	media-libs/alsa-lib
+	media-libs/freeglut
+	media-libs/freetype
+	x11-libs/libXrandr
+	x11-libs/libXcursor
+	x11-libs/libXinerama
+	jack? ( media-sound/jack-audio-connection-kit )
+	KWIKFormat? ( >sci-libs/hdf5-1.8.12[cxx(+)]	)
+	NetworkEvents? ( =net-libs/zeromq-3* )
+	NWBFormat? ( <sci-libs/hdf5-1.8.21 =sci-libs/hdf5-1.8*	)
```
TheChymera, line 47: These versions no longer exist in ::gentoo.

sci-biology/open-ephys-gui/open-ephys-gui-0.5.2.2.ebuild
```diff
@@ -0,0 +1,65 @@
...
+src_prepare() {
+	cmake_src_prepare
+
+	use jack && sed -i 's/JUCE_APP_VERSION=/JUCE_JACK=1\n    JUCE_APP_VERSION=/' "${WORKDIR}/$SubDir/CMakeLists.txt" || die "Sed failed!"
+}
+
+src_install() {
+	dodir opt/open-ephys-$Suffix
+	cp -R "${BUILD_DIR}"/Gentoo/* "${ED}"/opt/open-ephys-"$Suffix"/
+	dosym /opt/open-ephys-"$Suffix"/open-ephys usr/bin/open-ephys-"$Suffix"
```
TheChymera, line 63: You can use relative paths

TheChymera (changes requested):
@malfatti sorry for not noticing these the first time around.

The dependency part needs to be fixed, otherwise the ebuild is plain broken with current tree versions.

Additionally, you can run `repoman full` in the directory of your new ebuild before committing. It will point out most things me or other reviewers will notice, and possibly more :)

Lastly, when everything is fixed and `repoman full` gives you either no issues or at least no fatal ones, please rebase everything into one commit. Thanks ^^

sci-biology/open-ephys-gui/open-ephys-gui-0.5.2.2.ebuild
malfatti: Doh! That's true :) I missed this because I personally do not need those flags. Unfortunately, the code upstream depends on those specific versions (if the plugins are enabled). How should I proceed? Should I add back those ebuilds for zeromq and hdf5 in this same commit? Or should I drop these flags until upstream upgrades the code? The software works perfectly fine without those plugins, they just provide extra file formats for recording data to disk .

sci-biology/open-ephys-gui/open-ephys-gui-0.5.2.2.ebuild
malfatti: You mean `opt/open-ephys-"$Suffix"/open-ephys` instead of `/opt/open-ephys-"$Suffix"/open-ephys`?

sci-biology/open-ephys-gui/open-ephys-gui-0.5.2.2.ebuild
malfatti: Done :)

sci-biology/open-ephys-gui/open-ephys-gui-0.5.2.2.ebuild
TheChymera: Yes, that should work. Best to also try it out, though :)

malfatti: It didn't, sorry :/ see comment below.

malfatti:
Hi!

To my amusement, upstream now deals with plugins internally, so I do not need to compile them in the instalation process, making my life a lot easier :D 

I removed the plugins use flags and the dependencies. Now `repoman full` reports no fatal issues, however there are still two things to consider:

1. Repoman tells me I `could use relative path` in  the `dosym` inside `src_install`. However, if I change it to the relative path, after install I have a link `/usr/bin/open-ephys-{version}` pointing at `opt/open-ephys-{version}/open-ephys`, which do not exist (note the missing `/`.  Strangely, portage reports that the link points to `/usr/bin/opt/open-ephys-{version}/open-ephys`, which is not true.

2. I didn't notice this before, but portage is reporting `QA Notice: Pre-stripped files found` . I could not find a way to prevent the source from stripping files.

Of course, none of these are fatal and the software works perfectly.

sci-biology/open-ephys-gui/open-ephys-gui-0.5.2.2.ebuild
malfatti: Fixed upstream :)

TheChymera:
Ok, well, in the interest of giving fixing those issues a try, I would recommend:

1. `${D}/opt/`
2. asking Gentoo dev help (on freenode IRC) regarding the file stripping. I haven't encountered that before, so I'm not fully sure.

TheChymera:
@malfatti any updates on this? :)

malfatti:
Hey @TheChymera, happy new year :)

About 1), your suggestion worked fine, but I changed to ${ED} instead. It also worked :) I still didn't have time to work on 2).

AndrewAmmerlaan:
This looks good, could you please add a sign-off line to your commit. Can't merge without a proper sign-off line due to Gentoo's copyright policy, see:
https://wiki.gentoo.org/wiki/Talk:Gentoo_git_workflow#Sign-Off
https://www.gentoo.org/glep/glep-0076.html

sci-biology/open-ephys-gui/open-ephys-gui-0.5.2.2.ebuild
```diff
@@ -0,0 +1,75 @@
...
+	SubDir=${P}
+else
+	SRC_URI="https://github.com/open-ephys/plugin-GUI/archive/v${PV}.tar.gz"
+	Suffix=${PV}
+	SubDir="plugin-GUI-${PV}"
+	S="${WORKDIR}/$SubDir"
+fi
+
+SLOT="${PV}"
+KEYWORDS="~amd64 ~x86"
```
AndrewAmmerlaan, line 31: I would put the KEYWORDS inside the if statement, this makes version bumps easier because you can just copy version x+1 to the -9999 ebuild without having to edit it.

malfatti: Done, thanks :)

malfatti:
Hi @AndrewAmmerlaan,

I didn't know that, thanks for the links :) I made the changes and rebased. I hope everything is fine, because I actually could not push the signed commit. I did the following:

```bash
## added DCO_SIGNED_OFF_BY and PORTAGE_GPG_KEY to /etc/portage/make.conf, then
$ git config --local pull.ff only
$ git config --local pull.rebase merges
$ git config --local push.default simple
$ git config --local user.signingkey 0x29707BA965771A82
$ git config --local commit.gpgsign 1
$ git config --local push.gpgsign 1
$ repoman commit
$ git rebase -i HEAD~2
$ git push -f
[...]
fatal: the receiving end does not support --signed push
fatal: The remote end hung up unexpectedly
$ git push -f --no-signed
## [All done :)]
```

By the way, would you have a guess about the `QA Notice: Pre-stripped files found`? I still could not solve this (but of course the software installs and runs perfectly).

AndrewAmmerlaan:
>   By the way, would you have a guess about the QA Notice: Pre-stripped files found? I still could not solve this (but of course the software installs and runs perfectly).

This warning just tells you that some files installed to the system where already stripped when emerge tried to strip all the installed files. You can prevent the stripping entirely with `RESTRICT="strip". However, that is not what we want here because we want the installed files to be stripped of the debug symbols. The fact that some of these files where already stripped is just something that tells you that some of the installed files were pre-built or stripped by something else. If there is some compile/install script that strips this, than this should be patched to not do that. However, I suspect that the reason this warning is shown is because some of these files were prebuilt, in this case you can just add a list of these files to `QA_PREBUILT` or `QA_PRESTRIPPED`. See also the "QA control variables" section of the devmanual: https://devmanual.gentoo.org/eclass-reference/ebuild/index.html . Note that you can use a `*` if everything in the sources is prebuilt, e.g. `QA_PREBUILT="*"`.


> ```shell
> fatal: the receiving end does not support --signed push
> fatal: The remote end hung up unexpectedly
> ```

Either the remote is configured to point to the wrong repository (but I suspect not, because than it wouldn't work at all). Or your GitHub account is not configured for signed pushes (You need to upload that key to your GitHub account so it can actually verify the signed push). Either way it is not really a problem, because you are not pushing directly to the repo, as long as the commit is signed and contains a signed-off line it is fine.

malfatti:
Thanks @AndrewAmmerlaan! Fixed the QA_ warnings and also fixed the package name in SRC_URI.

AndrewAmmerlaan:
Awesome, I'll test this soon (currently waiting for gcc-8 to compile)

2 minor things I'm wondering about:
1) The [udev eclass](https://devmanual.gentoo.org/eclass-reference/udev.eclass/index.html) adds the `udev_reload` command which is exported to `pkg_postinst` and automatically reloads udev rules/databases, would that be sufficient for this package or is a full restart of udev required?
2) Replacing `"${ED}"/opt/open-ephys-"$Suffix"/open-ephys usr/bin/open-ephys-"$Suffix"` with `"../../opt/open-ephys-$Suffix/open-ephys" "usr/bin/open-ephys-$Suffix" should get rid of the repoman/pkgcheck warning about dosym.

```
2021-02-05T11:47:50.1889266Z 
2021-02-05T11:47:50.1890813Z sci-biology/open-ephys-gui
2021-02-05T11:47:50.1892709Z   DuplicateFiles: duplicate identical files in FILESDIR: 'files/open-ephys-gui-0.5.2.2.patch', 'files/open-ephys-gui-9999.patch', 'files/open-ephys-gui-99999999.patch'
2021-02-05T11:47:50.1894504Z   RedundantLongDescription: metadata.xml longdescription closely matches DESCRIPTION
2021-02-05T11:47:50.1895987Z   AbsoluteSymlink: version 0.5.2.2: dosym called with absolute path on line 73: dosym "${ED}"/opt/open-ephys-"$Suffix"/open-ephys
2021-02-05T11:47:50.1897329Z   DuplicateEclassInherit: version 9999: duplicate eclass inherit 'git-r3', line 18
2021-02-05T11:47:50.1898930Z   AbsoluteSymlink: version 9999: dosym called with absolute path on line 73: dosym "${ED}"/opt/open-ephys-"$Suffix"/open-ephys
2021-02-05T11:47:50.1900437Z   DuplicateEclassInherit: version 99999999: duplicate eclass inherit 'git-r3', line 18
2021-02-05T11:47:50.1901995Z   AbsoluteSymlink: version 99999999: dosym called with absolute path on line 73: dosym "${ED}"/opt/open-ephys-"$Suffix"/open-ephys
2021-02-05T11:47:50.2150997Z 
```

AndrewAmmerlaan:
I'm getting:

```
* ERROR: sci-biology/open-ephys-gui-0.5.2.2::localrepo failed (prepare phase):
*   Sed failed!
* 
``

The reason is:

```
use jack && sed -i 's/JUCE_APP_VERSION=/JUCE_JACK=1\n    JUCE_APP_VERSION=/' "${WORKDIR}/${SubDir}/CMakeLists.txt" || die "Sed failed!"
```

If use is not jack, sed will not be executed, but die will be, resulting in that this cannot be emerged with USE="-jack".
A solution would be to put this in a full if statement:
```
if use jack; then
     sed -i 's/JUCE_APP_VERSION=/JUCE_JACK=1\n    JUCE_APP_VERSION=/' "${WORKDIR}/${SubDir}/CMakeLists.txt" || die "Sed failed!"
fi
```

malfatti:
Doh! Yes, I'll put the sed in the full statement and replace the `"${ED}"` as suggested. About the udev, a full restart appears to be needed. I don't know exactly why, but if the board is already plugged to the computer not even plug-out-plug-in works. The software only recognizes the board after `/etc/init.d/udev restart`.

AndrewAmmerlaan:
Perfect! Merged

Thank you for your contribution!

PR #968, opened 2020-08-26
sci-libs/pytorch: fix in patch name

Fougeroc:
Fix a bug in pytorch-1.6.0 ebuild.


TheChymera:
@Fougeroc thank you for this. The issue was also discussed on bugzilla: https://bugs.gentoo.org/738592 

the fix has not been pushed yet because the maintainers have not yet responded. Generally we're not supposed to modify peoples' packages without their consent, unless they're unresponsive for over 4 weeks after the bug was reported on bugzilla. So 2 more weeks left :)

Still, even with this fix there remain issues when building with the CUDA flag (without which the package is basically dead-slow and borderline unusable). Would you be interested in looking into this as well? https://bugs.gentoo.org/738682

TheChymera:
this fix has now been applied by the original authors.

heroxbd:
@Fougeroc Thank you!  I am sorry that I overlooked this PR.

PR #972, opened 2020-09-11
dev-libs/optix: Add optix v7.x

kshots:
Blender utilizes optix as an alternative to CUDA to utilize the
raytracing cores in the RTX series of video cards. version 7.0 has been
tested with Blender 2.8.4.

Package-Manager: Portage-3.0.4, Repoman-3.0.1


dev-libs/optix/optix-7.0.0.ebuild
```diff
@@ -0,0 +1,37 @@
+# Copyright 2020 Gentoo Authors
```
epsilon-0, line 1: Do we really need a `7.0.0` installation?
I would image a minor version bump should be backwards compatible, but then this *is* nvidia.

dev-libs/optix/optix-7.1.0.ebuild
```diff
@@ -0,0 +1,37 @@
...
+
+SLOT="0/7"
+KEYWORDS="~amd64"
+RESTRICT="fetch"
+LICENSE="NVIDIA-r2"
+
+RDEPEND="dev-util/nvidia-cuda-toolkit
+	virtual/opengl
+	media-libs/freeglut"
+DEPEND="${RDEPEND}"
```
epsilon-0, line 18: Remove `DEPEND`, installation is just an extraction, no need for any libraries.

dev-libs/optix/optix-7.1.0.ebuild
```diff
@@ -0,0 +1,37 @@
...
+HOMEPAGE="https://developer.nvidia.com/optix"
+SRC_URI="NVIDIA-OptiX-SDK-${PV}-linux64-x86_64.sh"
+
+SLOT="0/7"
+KEYWORDS="~amd64"
+RESTRICT="fetch"
+LICENSE="NVIDIA-r2"
+
+RDEPEND="dev-util/nvidia-cuda-toolkit
+	virtual/opengl
```
epsilon-0, line 16: Please order them alphabetically.

dev-libs/optix/optix-7.1.0.ebuild
```diff
@@ -0,0 +1,37 @@
...
+
+src_unpack() {
+	tail -n +223 "${DISTDIR}"/${A} | tar -zx || die
+}
+
+src_install() {
+	insinto /opt/${PN}
+	dodoc -r doc
+	doins -r include SDK
+}
```
epsilon-0, line 37: Otherwise look OK.

epsilon-0 (changes requested):
Thanks for the update.

dev-libs/optix/optix-7.0.0.ebuild
kshots: media-gfx/blender-2.8.4 does not compile with dev-libs/optix-7.1.0, but does compile with dev-libs/optix-7.0.0. However, media-gfx/blender-2.9.0 compiles with dev-libs/optix-7.1.0. So, depending on which version of blender you want, you may need to stay < 7.1.0

dev-libs/optix/optix-7.1.0.ebuild
kshots: Agreed - when I first poked at this, I thought the SDK needed to be compiled. Then I discovered the installer doesn't compile it itself, so why should we?

dev-libs/optix/optix-7.1.0.ebuild
kshots: I'm a bit mystified here... order what alphabetically?

dev-libs/optix/optix-7.0.0.ebuild
epsilon-0: thanks nvidia...

dev-libs/optix/optix-7.1.0.ebuild
epsilon-0: Sorry, maybe I was unclear.
Keep the `RDEPENDS`, only remove the `DEPENDS`.
This is important for cross compilation.

kshots: Oh! Putting them back now...

epsilon-0: RDEPENDS = packages needed during running time
DEPEND = packages needed during linking time
BDEPEND = packages needed to be **run** to compile.

epsilon-0:
You need to add a Sign-Off to the commit - https://wiki.gentoo.org/wiki/Talk:Gentoo_git_workflow#Sign-Off

Also squash the commits into one instead of so many.

kshots:
Thanks for the link, rebased with --signoff enabled to combine all three commits into one.

epsilon-0:
Thanks, committed.

PR #976, opened 2020-09-21
sci-biology/cutadapt - update + dependencies

epsilon-0:
should be pretty straight forward...


epsilon-0:
Got tests for `xopen` and `cutadapt`, passing for both.

`dnaio` does not distribute tests with tarball.

PR #978, opened 2020-10-03
sci-libs/libsufr: Fix build failure with USE=static-libs.

MarcvdSluys:
sci-libs/libsufr-0.7.5 fails to build with USE=static-libs. Using emake as the CMake Makefile
generator, instead of ninja, fixes this.

Package-Manager: Portage-3.0.4, Repoman-3.0.1

### DIFF
--- a/sci-libs/libsufr/libsufr-0.7.5-r1.ebuild
+++ b/sci-libs/libsufr/libsufr-0.7.5-r1.ebuild
@@ -5,6 +5,7 @@ EAPI=7
 
 CMAKE_BUILD_TYPE=Release
 inherit cmake fortran-2
+CMAKE_MAKEFILE_GENERATOR="emake"
 
 DESCRIPTION="LIBrary of Some Useful Fortran Routines"
 HOMEPAGE="http://libsufr.sourceforge.net/"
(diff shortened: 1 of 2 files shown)


TheChymera:
@MarcvdSluys thank you! Could you please merge these changes into one commit? This can be done by running `git rebase -i HEAD~2` and following the instructions.

MarcvdSluys:
Well, yes, I combined the two commits into one, but then I can't push before pulling.  After pulling it jumps from 1 to 3 commits....

MarcvdSluys:
@TheChymera Allright, I simply force pushed, since it's my branch anyway, and this seems to have worked.  So now I 1) fixed the ebuild, 2) revision bumped and 3) updated the metadata to adopt the package.

TheChymera:
Closed by https://gitweb.gentoo.org/proj/sci.git/commit/?id=55ae2a0e636be6ef170c2dc1b29f246daa8e5db7

Yes, after rebasing (in this case, i.e. “merging” commits), you need to force push, since you are in fact re-writing the branch history.

Thank you for the contribution! Also, if you are interested in helping provide a reliable solution for this package, please consider submitting a follow-up PR enabling its test suite. I know, prima facie it looks like superfluous extravagance, but it really helps greatly because it allows (a) users to verify whether issues they encounter are due to the package distribution in Gentoo or due to their usage of it, (b) developers who might not use the package to easily ascertain whether it still “works”.

MarcvdSluys:
It took me a long time to figure out what the problem (and hence the solution) was, but it turns out to be that the author of the package uses CMake, which works as expected, whereas Gentoo uses Ninja by default, which apparently doesn't work.

PR #986, opened 2020-11-20
sci-libs/h5zstd: Add zstd compress filter for hdf5

littlewu2508:
Introduce the Zstandard compression filter plugin for HDF5 to Gentoo via this ebuild. Also upport Gentoo-Prefix.

Package-Manager: Portage-3.0.5, Repoman-3.0.1
Signed-off-by: Benda Xu <heroxbd@gentoo.org>


epsilon-0:
Why does this have heroxbd's sign off?
You should sign this with your own name and email.

heroxbd:
> Why does this have heroxbd's sign off?
> You should sign this with your own name and email.

+1

@littlewu2508 I will sign-off on my own, after I review your ebuild and before I commit it to the tree.

heroxbd:
I think using the upstream name `HDF5Plugin-Zstandard` would be better.

littlewu2508:
> 
> 
> I think using the upstream name `HDF5Plugin-Zstandard` would be better.

OK. Because the package name changed, I created a new pull request: [](https://github.com/gentoo/sci/pull/987).

> 
> 
> Why does this have heroxbd's sign off?
> You should sign this with your own name and email.

Signing message corrected in the new pull request.

PR #988, opened 2020-11-21
sci-libs/pg2plplot: sci-libs/pg2plplot: version bump.

MarcvdSluys:
Package-Manager: Portage-3.0.9, Repoman-3.0.2


sci-libs/pg2plplot/pg2plplot-5.13.0.ebuild
```diff
@@ -1,9 +1,11 @@
-# Copyright 1999-2015 Gentoo Foundation
+# Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=5
+EAPI=7
 
-inherit cmake-utils fortran-2
+CMAKE_BUILD_TYPE=Release
```
epsilon-0, line 6: Can you see if this can be patched out?
Generally it should be a small 2-3 line patch.
 
The problem is that using `CMAKE_BUILD_TYPE=Release` overrides the `CFLAGS,FFLAGS` from `/etc/portage/make.conf` by using the default cmake flags.

sci-libs/pg2plplot/pg2plplot-5.13.0.ebuild
```diff
@@ -14,19 +16,20 @@ SLOT="0"
 KEYWORDS="~amd64 ~x86"
 IUSE="X png postscript static-libs"
```
epsilon-0, line 17: The `USE` flags are a bit weird.
They don't seem to be used in the configuration of the current package?
Does it automagically pick up that the dependencies are built with said flag or not?
Consider this - I don't enable `png` and `postscript` flags on `pg2plplot` but I do build `plplot` with `cairo`.
The building of `pg2pltplot` will pick up that `plplot` has `cairo` and will build the `png` and `postscript` parts, but if I later rebuild `plplot` without `cairo` I will get preserved libraries that will need to be rebuilt using `@preserved-rebuild`.

My proposal - remove `png`, `postscript` and `X` flags and make the dependency be `plplot[cairo,fortran,X]`.

sci-libs/pg2plplot/pg2plplot-5.13.0.ebuild
```diff
@@ -14,19 +16,20 @@ SLOT="0"
...
-	 png? ( sci-libs/plplot[cairo] )
-	 postscript? ( sci-libs/plplot[cairo] )"
+		 sci-libs/plplot[fortran,X?]
+		 png? ( sci-libs/plplot[cairo] )
+		 postscript? ( sci-libs/plplot[cairo] )"
 
 src_configure() {
 	local mycmakeargs=(
-		$(cmake-utils_use static-libs CREATE_STATICLIB)
+		-DCREATE_STATICLIB="$(usex static-libs)"
```
epsilon-0, line 30: Can you just remove the `static-libs` option?
Unless specifically needed by another package, it is recommended to disable the static-libs option.

Users can always enable it locally by using `CMAKE_EXTRA_CACHE_FILE` for specifying `static-libs`.

epsilon-0 (changes requested):

epsilon-0:
Also add a comment saying that it **Closes: https://bugs.gentoo.org/755812 **

sci-libs/pg2plplot/pg2plplot-5.13.0.ebuild
MarcvdSluys: I'd like to try, but lack the knowledge.  Can you point me to some documentation?

epsilon-0: No documentation specifically for this.
Look here - https://sourceforge.net/p/pg2plplot/code/ci/master/tree/CMakeLists.txt#l28
Just remove this check and the `set (CMAKE_CONFIGURATION_TYPES ...` line.

sci-libs/pg2plplot/pg2plplot-5.13.0.ebuild
MarcvdSluys: I can, but I'm not sure how to build the static libs then.  I would normally test this using `USE=static-libs ebuild pg2plplot-5.13.0.ebuild clean manifest install`.  How can I test it now?  I've googled on replacing the Gentoo `USE` flags with `CMAKE_EXTRA_CACHE_FILE` but I get very few hits.

epsilon-0: An example for creating a cmake cache file which will enable `BUILD_STATIC_LIBS` for any **<category/package>**.

- Create a folder `/etc/portage/cache`
- Create a file `/etc/portage/cache/CMAKE_STATIC` with the contents
```
set(BUILD_STATIC_LIBS ON CACHE BOOL "")
```
- Create a file `/etc/portage/env/cmake-static` with the contents
```
CMAKE_EXTRA_CACHE_FILE=/etc/portage/cache/CMAKE_STATIC
```
- Create a file `/etc/portage/package.env/<package>` with the contents
```
<category/package>    cmake-static
```

This will enable this option for **<category/package>**.

To enable this for any package, just do the `/etc/portage/package.env/<package>` step for it.   
In fact, using such a cache file, you can override all default arguments set in the ebuild.

epsilon-0:
Thanks a lot for the version bump!
I've merged it with static-libs for now.

PR #994, opened 2020-12-16
sys-kernel/dkms: dynamic kernel module support

AndrewAmmerlaan:


epsilon-0:
OK, this is looking good.
I'm not going to be able to merge right now cuz my build server got "semi-borked" but I'll look at the other PRs.

epsilon-0:
Hmm, after sleeping on it I had a thought. Maybe there should be a kernel config check that module loading and unloading support is enabled?

AndrewAmmerlaan:
> Hmm, after sleeping on it I had a thought. Maybe there should be a kernel config check that module loading and unloading support is enabled

Good idea! On it

AndrewAmmerlaan:
> Hmm, after sleeping on it I had a thought. Maybe there should be a kernel config check that module loading and unloading support is enabled?

Fixed, this now checks for CONFIG_MODULES, which should be sufficient for this package I think.

epsilon-0:
This killed the build because it could not find the .config in the chroot :open_mouth:
We want a warning not a kill, to do a warning you should do `CONFIG_CHECK="~MODULES"`, the `~` makes it non fatal.

AndrewAmmerlaan:
> This killed the build because it could not find the .config in the chroot open_mouth
> We want a warning not a kill, to do a warning you should do `CONFIG_CHECK="~MODULES"`, the `~` makes it non fatal.

Fixed, after reading the modules article on the Gentoo wiki I also added: `RDEPEND="sys-apps/kmod"`.

PR #995, opened 2020-12-16
Sci-ni: new catagory and scripts

AndrewAmmerlaan:


TheChymera:
@AndrewAmmerlaan thanks for the contribution.
(1) given the general scope of the package, going by the metadata description, don't you think it would be better suited for ::gentoo?
(2) Why `sci-ni`?

AndrewAmmerlaan:
> @AndrewAmmerlaan thanks for the contribution.
> (1) given the general scope of the package, going by the metadata description, don't you think it would be better suited for ::gentoo?

Well I don't think I'll ever get a PR of this size, containing some packages that violate Gentoo's FSH policy through proxy-maint, especially considering that proxy-maint gives low priority to PRs that introduce new packages. Besides the proxy-maint team is busy enough as is, I'd not want to burden them with a PR containing package that are only relevant to the subset of Gentoo users that also happens to be a scientist, or happen to have to use GPIB for other reasons. I could put this in ::guru, but to be honest I think thematically it fits better in the science overlay.

> (2) Why `sci-ni`?

Because these are scientific (sci) packages from National Instruments (ni).

TheChymera:
> Because these are scientific (sci) packages from National Instruments (ni).

Are you sure we need a special package category for this?

AndrewAmmerlaan:
> > Because these are scientific (sci) packages from National Instruments (ni).
> 
> Are you sure we need a special package category for this?

IMO it would be a good idea to have them in a separate category, because 1) there are a lot of them, and 2) since these are generated from a script it seems to me like a good idea to keep them isolated from the rest, see also: https://github.com/gentoo/sci/pull/993

TheChymera:
> because 1) there are a lot of them, and 2) since these are generated from a script it seems to me like a good idea to keep them isolated from the rest

ok, if you're looking to maintain this longer-term I guess it's your call how to organize it :)

AndrewAmmerlaan:
> > because 1) there are a lot of them, and 2) since these are generated from a script it seems to me like a good idea to keep them isolated from the rest
> 
> ok, if you're looking to maintain this longer-term I guess it's your call how you want it organized :)

I hope I can maintain these packages and the VISA/GPIB related packages long term, the script I wrote should make it easy :D

PR #996, opened 2020-12-16
eclass/rpm-extended: add eclass

AndrewAmmerlaan:

### DIFF
--- a/eclass/rpm-extended.eclass
+++ b/eclass/rpm-extended.eclass
@@ -1,0 +1,132 @@
+# Copyright 1999-2020 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+# @ECLASS: rpm-extended.eclass
+# @MAINTAINER:
+# Andrew Ammerlaan <andrewammerlaan@riseup.net>
+# @AUTHOR:
+# Andrew Ammerlaan <andrewammerlaan@riseup.net>
+# Extension of the rpm.eclass by Alastair Tse <liquidx@gentoo.org>
+# @SUPPORTED_EAPIS: 6 7
+# @BLURB: An eclass that helps automating the creation of ebuilds for software only distributed through rpm
+# @DESCRIPTION:
+# This extends the rpm eclass to also installs *all* files in the rpm,
+# and extracts post/pre(un)install scripts and runs them.
+#
+# This is mostly useful when creating ebuilds for rpm files in bulk.
+# In cases when it is not doable to write a Gentoo specific src_install
+# and pkg_post/pre/install/rm functions for each ebuild.
+
+# The rpm eclass provides the pkg_unpack function we need
+inherit rpm
+
+case "${EAPI:-0}" in
+	0|1|2|3|4|5|6)
+		die "Unsupported EAPI=${EAPI:-0} (too old) for ${ECLASS}"
+		;;
+	7)
+		;;
+	*)
+		die "Unsupported EAPI=${EAPI} (unknown) for ${ECLASS}"
+		;;
+esac
+
+# Need rpm to extract scripts
+BDEPEND="app-arch/rpm"
+
+# Otherwise we get the S does not exist error
+S="${WORKDIR}"
+# Everything in the rpm is prebuilt
+QA_PREBUILT="*"
+# Distributing binaries of rpm packages makes no sense
+# and it will likely not work, as the pkg_pre/post/inst/rm
+# phases will fail to extract the script from the rpm file
+# as it is not guaranteed to exist when using binpkgs
+RESTRICT-"bindist"
+
+# @FUNCTION: rpm-extended_src_compile
+# @DESCRIPTION:
+# As everything in the rpm file is prebuilt
+# we do not need to compile anything, therefore
+# we add an empty src_compile function to avoid
+# errors.
+rpm-extended_src_compile() {
+	# Nothing to do here
+	return
+}
+
+# @FUNCTION: rpm-extended_src_install
+# @DESCRIPTION:
+# The rpm has been extracted by the pkg_unpack
+# function from rpm.eclass. Now we just copy
+# everything over to the image directory, while
+# preserving permissions. If something is installed
+# to /usr/share/doc we move everything in there to
+# the correct Gentoo specific location: /usr/share/doc/${PF}
+rpm-extended_src_install() {
+	cp -a "${S}"/* "${ED}"
+	if [ -d "${ED}/usr/share/doc/" ]; then
+		# If this package contains docs, move to gentoo specific dir
+		mkdir "${ED}/usr/share/doc/${PF}" || die
+		for file in "${ED}/usr/share/doc"/* ; do
+			if [[ "${file}" == "${ED}/usr/share/doc/${PF}" ]]; then
+				continue
+			fi
+			mv "${file}" "${ED}/usr/share/doc/${PF}" || die
+		done
+	fi
+}
+
+# @FUNCTION: rpm-extended_pkg_preinst
+# @DESCRIPTION:
+# Some rpm files contain a preinstall script,
+# we extract this script, write it to a file
+# and execute it in the correct phase
+rpm-extended_pkg_preinst() {
+	for x in ${A}; do
+		rpm -qp --scripts "${DISTDIR}/${x}" | sed -n '/preinstall scriptlet (using \/bin\/sh):/,/scriptlet (using \/bin\/sh)/{//!p;}' > "preinst-${x}.sh"
+		chmod +x "preinst-${x}.sh"
+		bash "preinst-${x}.sh"
+	done
+}
+
+# @FUNCTION: rpm-extended_pkg_postinst
+# @DESCRIPTION:
+# Some rpm files contain a postinstall script,
+# we extract this script, write it to a file
+# and execute it in the correct phase
+rpm-extended_pkg_postinst() {
+	for x in ${A}; do
+		rpm -qp --scripts "${DISTDIR}/${x}" | sed -n '/postinstall scriptlet (using \/bin\/sh):/,/scriptlet (using \/bin\/sh)/{//!p;}' > "postinst-${x}.sh"
+		chmod +x "postinst-${x}.sh"
+		bash "postinst-${x}.sh"
+	done
+}
+
+# @FUNCTION: rpm-extended_pkg_prerm
+# @DESCRIPTION:
+# Some rpm files contain a preuninstall script,
+# we extract this script, write it to a file
+# and execute it in the correct phase
+rpm-extended_pkg_prerm() {
+	for x in ${A}; do
+		rpm -qp --scripts "${DISTDIR}/${x}" | sed -n '/preuninstall scriptlet (using \/bin\/sh):/,/scriptlet (using \/bin\/sh)/{//!p;}' > "prerm-${x}.sh"
+		chmod +x "prerm-${x}.sh"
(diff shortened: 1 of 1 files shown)


AndrewAmmerlaan:
I just now made one small change to the eclass, instead of doing `source ....sh` I now dow `bash ....sh`, this way emerge doesn't complain if the script calls exit, which means that I don't have to perform some sed magic to remove all exit commands.

epsilon-0:
Without looking too much into the eclass yet, maybe rpm-extended isn't the best name?
Case in point, there is the intel-sdp-r1 eclass, which also does a bunch of rpm things, but the name is specific to make it clear it's doing intel specific rpm things. Maybe the name here should be reflective of the sci-ni things that are happening.
This is easily fixable so no need to change just yet, we can do this at the last part.

AndrewAmmerlaan:
> Without looking too much into the eclass yet, maybe rpm-extended isn't the best name?
> Case in point, there is the intel-sdp-r1 eclass, which also does a bunch of rpm things, but the name is specific to make it clear it's doing intel specific rpm things. Maybe the name here should be reflective of the sci-ni things that are happening.
> This is easily fixable so no need to change just yet, we can do this at the last part.

In principle this eclass should work for *any* rpm file, all it does extra is extract the rpm's post/pre(un)install scripts and runs them, it is not really specific to rpm files from Nation Instruments, this is on purpose, I tried to write in such a way that it is universally applicable. I suppose I could still rename it to something like `auto-rpm.eclass`.

I suppose I could (re)write the eclass in such a way that _is_ specific to these NI ebuilds, in which case I could move some stuff that is now in the script that generates the ebuilds to the eclass, such as `LICENSE`, `KEYWORDS`, `pkg_nofetch`, but IMO it is better to keep eclases as universally as possible.

eclass/rpm-extended.eclass
```diff
@@ -0,0 +1,76 @@
+# Copyright 1999-2020 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
```
epsilon-0, line 3: For cross-reference, the python-single-r1.eclass is a good place to look at for my comments: https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/python-single-r1.eclass

eclass/rpm-extended.eclass
```diff
@@ -0,0 +1,76 @@
...
+# Author : Andrew Ammerlaan <andrewammerlaan@riseup.net> (16-12-2020)
```
epsilon-0, line 4: add `@AUTHOR`

eclass/rpm-extended.eclass
epsilon-0, line 3: add `@ECLASS`, `@MAINTAINER` and basically all the things from the header from the `python-single-r1` eclass.

eclass/rpm-extended.eclass
```diff
@@ -0,0 +1,76 @@
...
+# post/pre(un)install scripts and runs them.
+#
+# This is mostly useful when creating ebuilds for rpm files in bulk.
+# In cases when it is not doable to write a Gentoo specific src_install
+# and pkg_post/pre/install/rm functions for each ebuild.
+
+inherit rpm
+
+# Need rpm to extact scripts
+BDEPEND="app-arch/rpm"
```
epsilon-0, line 18: This should already be there from `inherit rpm`

eclass/rpm-extended.eclass
```diff
@@ -0,0 +1,76 @@
...
+#
+# Extension of the rpm.eclass by Alastair Tse <liquidx@gentoo.org>
+#
+# This also installs *all* files in the rpm, and extracts
+# post/pre(un)install scripts and runs them.
+#
+# This is mostly useful when creating ebuilds for rpm files in bulk.
+# In cases when it is not doable to write a Gentoo specific src_install
+# and pkg_post/pre/install/rm functions for each ebuild.
+
```
epsilon-0, line 14: add the supported `EAPI` checks.

eclass/rpm-extended.eclass
```diff
@@ -0,0 +1,76 @@
...
+# Need rpm to extact scripts
+BDEPEND="app-arch/rpm"
+
+# Otherwise we get the S does not exist error
+S="${WORKDIR}"
+# Everything in the rpm is prebuilt
+QA_PREBUILT="*"
+
+rpm-extended_src_compile() {
```
epsilon-0, line 25: maybe add a very small documentation header saying it does nothing like: https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/python-single-r1.eclass#n519

eclass/rpm-extended.eclass
```diff
@@ -0,0 +1,76 @@
...
+	# Nothing to do here
+	return
+}
+
+rpm-extended_src_install() {
```
epsilon-0, line 30: As I am unfamiliar with how RPMs work (and I am sure there might be others) can you add small documentations for what all the functions do?

epsilon-0 (changes requested):
Mostly comments on adding documentation.

AndrewAmmerlaan:
> Hmm, unfortunately, I am unfamiliar with how RPMs are packaged and what their structure is so I am unable to comment and check if this eclass works perfectly for all RPMs.
> I've asked @SoapGentoo and lets see if he has any comments.


In principle a RPM file is an archive with some metadata. The metadata contains thins like, name, version, revision, description, url and stuff, we can extract this information with `rpm --info`. This is basically what the script in the other PR is doing. Along with this metadata 5 scripts can be included, preinstall, postinstall, preuninstall, postuninstall and verify, we can view these scripts with `rpm --scripts`. Unfortunately there is no way (that I know) to have rpm output only the script for a specific stage, instead it outputs them all at once. Therefore we use some sed magic to separate each script from this output, write it to a file, and execute it. And this is basically the part that the eclass does. Together the ebuild-generation-script and the eclass, create something which is more or less like an "ebuild-wrapper" around the rpm file, which does the exact same thing that would happen if you would install the rpm file using `rpm --install` except now dependencies and files are tracked by portage, which is more user friendly than having to run two separate package managers (portage and rpm) on one system.

eclass/rpm-extended.eclass
AndrewAmmerlaan: On it

eclass/rpm-extended.eclass
AndrewAmmerlaan: The rpm eclass does not actually add a dependency on rpm, instead it adds a dependency on rpm2targz, see: https://gitweb.gentoo.org/repo/proj/prefix.git/tree/eclass/rpm.eclass?id=d7fc8cf65c536224bace1d22c0cd85a526490a1e

epsilon-0: :open_mouth:

AndrewAmmerlaan:
@epsilon-0 I added some documentation in the eclass, and improved the overall style of the thing to, let me know if that clarifies things. :)

eclass/rpm-extended.eclass
```diff
@@ -0,0 +1,131 @@
...
+# and pkg_post/pre/install/rm functions for each ebuild.
+
+# The rpm eclass provides the pkg_unpack function we need
+inherit rpm
+
+case "${EAPI:-0}" in
+	0|1|2|3|4|5)
+		die "Unsupported EAPI=${EAPI:-0} (too old) for ${ECLASS}"
+		;;
+	6|7)
```
epsilon-0, line 27: Do we need 6? We can simplify life by just keeping it 7 yea?

AndrewAmmerlaan: Well no we don't really need 6, but from what I understand there are some eclasses that are not compatible with 7 yet (gnome-something if I remember correctly), by keeping 6 in we make sure this eclass is compatible with those eclasses.

epsilon-0: I'd say, let's remove 6. Saves future work. 
Highly unlikely we are going to be crossing paths with gnome.

AndrewAmmerlaan: Alright, 6 is gone :)

epsilon-0:
OK, documentation wise lgtm.
Still going to wait for other comments.

epsilon-0:
OK, this is not logically correct. pkg_* stages don't have access to the distfiles. What if package is being installed as a binpkg?

AndrewAmmerlaan:
> OK, this is not logically correct. pkg_* stages don't have access to the distfiles. 

What do you mean? The variable DISTDIR exists in these phases and the rpm command can access and read the files in it successfully.

> What if package is being installed as a binpkg?

All these ebuilds are RESTRICT="mirror bindist" so there shouldn't be any binpkg's of these package's. That being said, perhaps RESTRICT="bindist" should be moved into the eclass instead of leaving it up to the ebuild to set it? Since distributing binaries of something that already is a binary does not make any sense anyway?

epsilon-0:
Not sure if mirror restriction is needed but bindist is for sure. The pkg_* functions are assuming access to the rpm package which is not available is done in a bindist.
A possible solution would be to install the scripts as part of the package in a directory like `/usr/share/<pkg>/<script>.sh` so that they are available properly in the pkg_* phases. But that still doesn't take care of preinst...

AndrewAmmerlaan:
Not sure if mirror restriction is needed but bindist is for sure. The pkg_* functions are assuming access to the rpm package which is not available is done in a bindist.

The mirror restriction is necessary because of the license, but that can stay in the ebuilds and doesn't have to go in the eclass I guess. Is there a way to force portage to download the contents of SRC_URI even if it is a binpkg? Otherwise I'll just add RESTRICT="bindist" to the eclass, which should prevent the problem from occurring. 

> A possible solution would be to install the scripts as part of the package in a directory like /usr/share/<pkg>/<script>.sh so that they are available properly in the pkg_* phases. But that still doesn't take care of preinst...

I could for example extract the scripts in pkg_unpack, and save them in some directory e.g. ${S}/var/tmp/${PF}/scripts/, this way they will be available if the package is a binpkg, and they would also be available in pkg_preinst, since pkg_unpack is run before pkg_preinst. 

Another solution would be to extract the scripts in the ebuild generation script and save them in ${FILESDIR}, but I'm pretty sure NI's license also covers the rpm-scripts meaning that we cannot redistribute those scripts in the repo. So probably this solution is legally impossible.

What do you think? add RESTRICT="bindist", or move the extraction of the scripts to the pkg_unpack phase and install them somewhere in ${S}/ (which will then later be installed to ${ED} and thus to the system)? I'm thinking the later will be less error-prone, but the disadvantage is that the script are installed to the file system.

AndrewAmmerlaan:
> What do you think? add RESTRICT="bindist", or move the extraction of the scripts to the pkg_unpack phase and install them somewhere in ${S}/ (which will then later be installed to ${ED} and thus to the system)? I'm thinking the later will be less error-prone, but the disadvantage is that the script are installed to the file system.

Alright, I did some testing and it seems the latter solution is impossible. The DISTDIR does not exist in `pkg_setup`, and I can't use portageq to get it because emerge won't let me do that, which means I can't access the rpm file in this phase. I can't access the rpm command in the other phases, because it wants to write a lock file to /var/lock which results in ACCESS DENIED. The earliest I can do this is the pkg_preinst phase, but that is run *after* a hypothetical binpkg has been generated, so it does not solve our problem. It looks like we'll just have to RESTRICT="bindist" to solve the problem, which is not that bad of a solution since making a binary of the binary makes no sense to begin with.

epsilon-0:
OMG OMG OMG, mistake, mistake, undo undo undo

PR #998, opened 2020-12-20
sci-libs/ni-visa: cleanup

AndrewAmmerlaan:
@epsilon-0 ::natinst overlay is online at https://github.com/AndrewAmmerlaan/natinst . I'm currently waiting for my PR to api-gentoo-org to be approved: https://github.com/gentoo/api-gentoo-org/pull/348 .

In the meantime, this PR should cleanup things here, let me know if I missed something :smile:


epsilon-0:
Over a quick glance it seems OK. I'll do a full repoman and pkgcheck analysis soon.
I'll merge it after you PR to the overlay list is merged.
I want to put a news item here which tells about your overlay in case there are some people who are using the packages still and would need to know.

PR #999, opened 2020-12-29
*/* misc fixes to make repoman happier

AndrewAmmerlaan:
- http to https where possible
- remove some blank lines
- quote unquoted variables
- '=' to '~' where applicable
- add slot operator where necessary
- correct one dependency that has been renamed in ::gentoo (dev-util/Hermes to dev-util/hermes)
- fdo-mime to xdg-utils (fdo-mime has been removed from ::gentoo, functionality is identical)
- clean up profiles/ (there's no need to keep a copy of `arch.list` or `categories` in ::sci if we are not adding arches or categories, it got out of date and lead to annoying warnings)
- clean up licenses/ (remove licenses that are in ::gentoo, clean up `license_groups`)

This fixes most of the smaller trivial issues that repoman finds in this repository. Once these are fixed I might see about fixing some of the bigger non-trivial ones if I have the time.


epsilon-0:
Thanks so much !!
So, I fixed the hermes/Lmod things as they were in need of a few more complex changes (moving to slotted lua).

Could you remove those?
I'm still taking a look at the other things :heart:

app-doc/scalapack-docs/scalapack-docs-2.0.ebuild
```diff
@@ -1,10 +1,10 @@
-# Copyright 1999-2016 Gentoo Foundation
+# Copyright 1999-2020 Gentoo Authors
```
epsilon-0, line 1: I deleted the package - it was from dev space and it is no longer there.

AndrewAmmerlaan:
> Could you remove those?

Removed.

```
RDEPEND.suspect               2
sys-cluster/lmod/lmod-8.4.19.ebuild: 'virtual/pkgconfig'
sys-cluster/lmod/lmod-9999.ebuild: 'virtual/pkgconfig'
```

Should lmod RDEPEND on pkgconfig?

epsilon-0:
[](url)

> > Could you remove those?
> 
> Removed.
> 
> ```
> RepoMan scours the neighborhood...
> RDEPEND.suspect               2
> sys-cluster/lmod/lmod-8.4.19.ebuild: 'virtual/pkgconfig'
> sys-cluster/lmod/lmod-9999.ebuild: 'virtual/pkgconfig'
> ```
> 
> Should lmod RDEPEND on pkgconfig?

Yes, it uses pkg-config in scripts to mess around with LD_PRELOAD and LD_LIBRARY_PATH.

epsilon-0:
OK, so a quick glance makes it look fine.
We can try to split it up now, maybe one thing at a time, so first one that looks fine would be the license cleanup.
Could you create a PR for that? 
That should be pretty easy to merge, only one package ebuild changes which uses CeCILL-2.0, rest are deletions and the license group cleanup.

AndrewAmmerlaan:
> OK, so a quick glance makes it look fine.
> We can try to split it up now, maybe one thing at a time, so first one that looks fine would be the license cleanup.
> Could you create a PR for that?
> That should be pretty easy to merge, only one package ebuild changes which uses CeCILL-2.0, rest are deletions and the license group cleanup.

Alrighty, this PR here only cleansup the licenses: https://github.com/gentoo/sci/pull/1000

epsilon-0:
Awesome, next can you do the profiles/* cleanup? 
That one also looks straightforward enough to merge fast.

AndrewAmmerlaan:
> Awesome, next can you do the profiles/* cleanup?
> That one also looks straightforward enough to merge fast.

PR for cleaning up profiles/: https://github.com/gentoo/sci/pull/1001

epsilon-0:
Great I merged that.
The http -> https one is something that I'd like but at the same time... almost all those packages don't even fetch, regardless of http or https :crying_cat_face: 
Well, at least adding those would make it somewhat easier to look for other errors, so we are not losing anything at least.

Can you rebase? I think the http to https is the only one remaining yea?

AndrewAmmerlaan:
> Great I merged that.
> The http -> https one is something that I'd like but at the same time... almost all those packages don't even fetch, regardless of http or https crying_cat_face
> Well, at least adding those would make it somewhat easier to look for other errors, so we are not losing anything at least.

My thoughts exactly, some of these packages are still broken, but now at the very least they are slightly less broken which makes it easier to find and fix other errors. Otherwise repoman spits out so many errors and warnings that we can no longer see the wood for the trees.

> Can you rebase? I think the http to https is the only one remaining yea?

Rebased :smile:

epsilon-0:
Thanks a lot !!

PR #1002, opened 2020-12-30
More fixes

AndrewAmmerlaan:
Both of these still need more work, MotionCor2 needs a version bump (but it is fetch restricted and I don't have access tot the software), fits needs a new build.xml file if we want to compile it from source (currently it installs a precompiled .jar file).

This at least fixes them to an acceptable level where the ebuild is usable and repoman is happy.

I'm starting on the dev-python/ ones next.


epsilon-0:
Can you put the fits one into a separate PR? I want to merge that in first, it looks good.

The MotionCor2 looks weird though...

AndrewAmmerlaan:
> Can you put the fits one into a separate PR? I want to merge that in first, it looks good.
> 
> The MotionCor2 looks weird though...

Separate PR is here: https://github.com/gentoo/sci/pull/1004

The MotionCor2 one is a bit difficult, because it is fetch restricted. Repoman is currently very unhappy about the ebuild because =dev-util/nvidia-cuda-toolkit-10.1*  does not exist, so I opted to drop the flags and have it depend on  =dev-util/nvidia-cuda-toolkit-9.2* unconditionally. It fixes the repoman errors and warnings, but it's an ugly workaround and what it really needs is a version bump.

epsilon-0:
So I managed to download the software, fwiw. It's not actually that bad to download it, just have to accept some tldr conditions about using it for not bad commercial things.
Unfortunately, I don't have easy access to CUDA right now...

AndrewAmmerlaan:
> Unfortunately, I don't have easy access to CUDA right now...

Me neither, but since it is a binary we could just `dobin stuff` and hope for the best. Otherwise we maybe should consider removing the package, because as things are now `emerge MotionCor2` simply fails because of missing dependencies.

epsilon-0:
> > Unfortunately, I don't have easy access to CUDA right now...
> 
> Me neither, but since it is a binary we could just `dobin stuff` and hope for the best. Otherwise we maybe should consider removing the package, because as things are now `emerge MotionCor2` simply fails because of missing dependencies.

TBH I'm in two minds about the package - on one hand, it's pretty simple to package, we only need to add the ldd dependencies for missing SONAMEs; on the other hand, it's literally just a binary, there's really not a lot of benefit from having it on a package manager level.

If you aren't too keen on doing this, I can do the update in a few hours.

AndrewAmmerlaan:
Awesome, Thanks!

PR #1004, opened 2020-12-30
dev-java/fits: version bump 1.15.1, EAPI bump

AndrewAmmerlaan:
can't get the PN-src to compile, the build.xml
probably needs to be updated. This now installs
the pre-compiled version instead, maybe later we
can switch back to the -src version


epsilon-0:
Thanks!!

PR #1005, opened 2020-12-30
Dev-python/snakemake: version bump

AndrewAmmerlaan:


dev-python/datrie/metadata.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="person">
```
epsilon-0, line 4: not sure we can do this...
They need to agree with moving/copying to ::science
Or we can drop them as maintainer and only keep science.

dev-python/snakemake/snakemake-5.31.1.ebuild
```diff
@@ -0,0 +1,56 @@
...
+# missing test dep google.cloud
+RESTRICT="test"
+
+BDEPEND="test? (
+	dev-python/requests-mock[${PYTHON_USEDEP}]
+	dev-python/google-api-python-client[${PYTHON_USEDEP}]
+	dev-python/jinja[${PYTHON_USEDEP}]
+	dev-python/networkx[${PYTHON_USEDEP}]
+	dev-python/pygments[${PYTHON_USEDEP}]
+	dev-python/pygraphviz[${PYTHON_USEDEP}]
```
epsilon-0, line 28: This feels very much like it is going to be doing network calls in the testing phase >.>

dev-python/toposort/metadata.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="person">
+		<email>mschu.dev@gmail.com</email>
+		<name>Michael Schubert</name>
+	</maintainer>
+	<maintainer type="project">
```
epsilon-0, line 8: Same, I don't think we should add python as maintainer.

dev-python/toposort/metadata.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="person">
```
epsilon-0, line 4: can't add them

dev-python/datrie/metadata.xml
```diff
@@ -0,0 +1,16 @@
...
+		<email>mschu.dev@gmail.com</email>
+		<name>Michael Schubert</name>
+	</maintainer>
+	<maintainer type="project">
```
epsilon-0, line 8: can't add python

epsilon-0 (changes requested):
starter comments.

dev-python/datrie/metadata.xml
AndrewAmmerlaan: Oops, I thought I fixed that. Seems I messed up somewhere when copying things.

dev-python/snakemake/snakemake-5.31.1.ebuild
AndrewAmmerlaan: Yes I think so too, which is why I'm not bothering with packaging the missing dependency, the testsuite will most likely fail anyway. I just added RESTRICT="test"

AndrewAmmerlaan:
Metadata.xml files should be fine now :smile:

epsilon-0:
Thanks a lot!

PR #1008, opened 2020-12-30
sci-biology/gffutils: drop py3.6, drop old

AndrewAmmerlaan:
py3.6 has been dropped from dep in ::gentoo
repoman is unhappy about it

### DIFF
--- a/sci-biology/gffutils/gffutils-0.10.1.ebuild
+++ b/sci-biology/gffutils/gffutils-0.10.1.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{6..9} )
+PYTHON_COMPAT=( python3_{7..9} )
 
 inherit distutils-r1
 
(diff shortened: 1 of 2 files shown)


TheChymera:
Thanks for the contribution. But why drop the older of the 2 versions? It's generally best to have 2 working versions so that if there are version-specific upstream error people can compare.

AndrewAmmerlaan:
> Thanks for the contribution. But why drop the older of the 2 versions? It's generally best to have 2 working versions so that if there are version-specific upstream error people can compare.

I figured the old version would not be compatible with the newer python versions, but I'll re-add it :smile: . If there are any problems it can always be deleted later.

epsilon-0:
I tested it locally, its working nicely.
Can you also remove empty IUSE ?

AndrewAmmerlaan:
> Can you also remove empty IUSE ?

Removed

epsilon-0:
Thanks !!

PR #1010, opened 2020-12-30
sci-mathematics/dsfmt: version bump to 2.2.4

AnAverageHuman:
This should be moved to ::gentoo once Julia 1.6 as the language maintainers have
upstreamed their patches (and thus we can compile using the system dSFMT).


epsilon-0:
Thanks a lot for the bump!

PR #1012, opened 2020-12-31
Sci-biology/pyfaidx: version bump and drop py3.6

AndrewAmmerlaan:
Now repoman is happy :smile:


epsilon-0:
thanks!

epsilon-0:
Tests are still failing for both the packages.

AndrewAmmerlaan:
> Tests are still failing for both the packages.

Hmm, strange, this worked when I tested it before, perhaps there is a missing dependency or a wrong version of a dependency?

PR #1013, opened 2020-12-31
Sci-biology/salmon: version bump, EAPI bump, cmake-utils --> cmake

AndrewAmmerlaan:
I need some help with this one, it fails to find boost:

```
-- Could NOT find Boost (missing: iostreams filesystem system timer chrono program_options) (found suitable version "1.75.0", minimum required is "1.59.0")
BOOST_INCLUDEDIR = 
BOOST_LIBRARYDIR = 
Boost_FOUND = FALSE
CMake Error at CMakeLists.txt:473 (message):
Salmon cannot be compiled without Boost.

It is recommended to visit http://www.boost.org/ and install Boost
according to those instructions.

This build system can also download and install a local version of boost
for you (this takes a lot of time).

To fetch and build boost locally, call cmake with -DFETCH_BOOST=TRUE


-- Configuring incomplete, errors occurred!
```

Either we are missing the correct version of boost, or we are missing an USE flag. My attempts at fixing it have been unsuccessful so far :cry:


sci-biology/salmon/salmon-1.4.0.ebuild
```diff
@@ -0,0 +1,27 @@
...
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+BDEPEND="net-misc/curl
+		app-arch/unzip
+		sys-libs/zlib
+		app-arch/bzip2
+		app-arch/xz-utils
+		dev-libs/boost:0[threads]
```
waebbl, line 22: boost in BDEPEND, is this true?
And the dep is IMO missing an `=` after the slot
Although I don't think, this will solve the issue

AndrewAmmerlaan: > boost in BDEPEND, is this true?

I think so, at least it was there before I started changing things, and it does seem to want it at build time (the configure phase fails with the error that it is missing, so it is definitely looking for boost)

> And the dep is IMO missing an = after the slot

Boost only has one slot (so I guess the `:0` is useless and should go?)

waebbl: > > boost in BDEPEND, is this true?
> 
> I think so, at least it was there before I started changing things, and it does seem to want it at build time (the configure phase fails with the error that it is missing, so it is definitely looking for boost)

But then it uses the dependencies in a wrong way. BDEPEND should only contain packages on the build machine (CBUILD), which need not be available on the machine the package is going to get installed (CHOST). Just using RDEPEND=${BDEPEND} is, in my opinion no clean way. I doubt you would get such a construction merged on ::gentoo.
Have a package available at build time is reached by using DEPEND=${RDEPEND}.
https://devmanual.gentoo.org/general-concepts/dependencies/index.html

> 
> > And the dep is IMO missing an = after the slot
> 
> Boost only has one slot (so I guess the `:0` is useless and should go?)

The 0 can go away, but it needs a `:=` slot operator, as it has subslots, to force a rebuild after boost get's updated.

AndrewAmmerlaan: Like this?

waebbl: Yes this looks better. But you have to check the dependencies carefully. zlib is most likely a runtime dependency. For tbb, it depends how it's used. If the library is linked, it's RDEPEND, if only the template headers are used and no linking is done, it can probably be BDEPEND. For curl, is the fetchPufferfish.sh script installed or only used during build time? For the first, curl needs to be in RDEPEND, for the latter BDEPEND would IMO be sufficient. Just check whether the dependencies are needed on CHOST (RDEPEND) or **only** on CBUILD (BDEPEND).

AndrewAmmerlaan: It probably doesn't have to depend on curl or unzip at all, because the patch removes those commands from the fetchPutterfish.sh script

AndrewAmmerlaan: > It probably doesn't have to depend on curl or unzip at all, because the patch removes those commands from the fetchPutterfish.sh script

Nope, I was wrong

waebbl:
Think I found the issue: `set(Boost_ADDITIONAL_VERSIONS "1.59.0" "1.60.0" "1.61.0" "1.62.0" "1.63.0" "1.64.0" "1.65.0" "1.66.0" "1.67.0" "1.68.0" "1.69.0" "1.70.0" "1.71.0")`
Looks like the package is not ready for anything >boost-1.71.0. If this is true, we're stuck here, but you can try to add 1.7{4,5}.0 to the list and see if that works.

AndrewAmmerlaan:
> set(Boost_ADDITIONAL_VERSIONS

I applied this patch:

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 1be02ab..7e04e42 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -428,7 +428,7 @@ endif()
 ##
 # Set the latest version and look for what we need
 ##
-set(Boost_ADDITIONAL_VERSIONS "1.59.0" "1.60.0" "1.61.0" "1.62.0" "1.63.0" "1.64.0" "1.65.0" "1.66.0" "1.67.0" "1.68.0" "1.69.0" "1.70.0" "1.71.0")
+set(Boost_ADDITIONAL_VERSIONS "1.59.0" "1.60.0" "1.61.0" "1.62.0" "1.63.0" "1.64.0" "1.65.0" "1.66.0" "1.67.0" "1.68.0" "1.69.0" "1.70.0" "1.71.0" "1.74.0" "1.75.0")
 if (NOT BOOST_RECONFIGURE)
 find_package(Boost 1.59.0 COMPONENTS iostreams filesystem system timer chrono program_options)
 message("BOOST_INCLUDEDIR = ${BOOST_INCLUDEDIR}")
```

And got the exact same error, re-tried with version 1.74.0 and got the same error again.

What should we do now? Remove the package?

waebbl:
Is the shared libs flag set? I noticed it wants static libs by default, and there's no src_configure in the ebuild, so probably cmake is searching for static boost libs?

AndrewAmmerlaan:
> Is the shared libs flag set? I noticed it wants static libs by default, and there's no src_configure in the ebuild, so probably cmake is searching for static boost libs?

That did the trick, Awesome.

Ran into another error, but this one looks easily fixable, will look into it.

```
ninja -v -j12 -l0
ninja: error: '/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/install/lib/libstaden-read.a', needed by 'src/unitTests', missing and no known rule to make it
```

AndrewAmmerlaan:
Well this package is being very annoying, every time I fix one thing, two other errors show up.

I got it to start compiling, but it fails :cry: 

```
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/src/PufferfishBinaryGFAReader.cpp:107:34: error: request for member ‘length’ in ‘str’, which is of non-class type ‘int’
107 |       for (size_t i = 0; i < str.length(); ++i) {
|                                  ^~~~~~
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/src/PufferfishBinaryGFAReader.cpp:108:42: error: invalid types ‘int[size_t {aka long unsigned int}]’for array subscript
108 |         auto c = kmers::codeForChar(str[i]);
|                                          ^
In file included from /var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/PufferfishIndex.hpp:11,
from /var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/src/PufferfishIndexer.cpp:22:
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp: At global scope:
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp:18:8: error: ‘string_view’ in namespace ‘stx’ does not name a type
18 |   stx::string_view s_;
|        ^~~~~~~~~~~
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp:81:15: error: ‘string_view’ in namespace ‘stx’ does not name a type
81 |   inline stx::string_view seq() { return s_; }
|               ^~~~~~~~~~~
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp: In constructor ‘pufferfish::CanonicalKmerIterator::CanonicalKmerIterator()’:
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp:33:7: error: class ‘pufferfish::CanonicalKmerIterator’ does not have any field named ‘s_’
33 |     : s_(), p_(), /*km_(), pos_(),*/ invalid_(true), lastinvalid_(-1),
|       ^~
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp: In constructor ‘pufferfish::CanonicalKmerIterator::CanonicalKmerIterator(const string&)’:
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp:36:7: error: class ‘pufferfish::CanonicalKmerIterator’ does not have any field named ‘s_’
36 |     : s_(s), p_(), /*km_(), pos_(),*/ invalid_(false), lastinvalid_(-1),
|       ^~
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp: In copy constructor ‘pufferfish::CanonicalKmerIterator::CanonicalKmerIterator(const pufferfish::CanonicalKmerIterator&)’:
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp:41:7: error: class ‘pufferfish::CanonicalKmerIterator’ does not have any field named ‘s_’
41 |     : s_(o.s_), p_(o.p_), /*km_(o.km_), pos_(o.pos_),*/ invalid_(o.invalid_),
|       ^~
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp:41:12: error: ‘const class pufferfish::CanonicalKmerIterator’ has no member named ‘s_’; did you mean ‘p_’?
41 |     : s_(o.s_), p_(o.p_), /*km_(o.km_), pos_(o.pos_),*/ invalid_(o.invalid_),
|            ^~
|            p_
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp: In member function ‘pufferfish::CanonicalKmerIterator& pufferfish::CanonicalKmerIterator::operator=(const pufferfish::CanonicalKmerIterator&)’:
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp:46:7: error: ‘s_’ was not declared in this scope; did you mean ‘p_’?
46 |       s_ = o.s_;
|       ^~
|       p_
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp:46:14: error: ‘const class pufferfish::CanonicalKmerIterator’ has no member named ‘s_’; did you mean ‘p_’?
46 |       s_ = o.s_;
|              ^~
|              p_
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp: In member function ‘void pufferfish::CanonicalKmerIterator::find_next(int, int)’:
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp:60:33: error: ‘s_’ was not declared in this scope; did you mean ‘p_’?
60 |     for (; j < static_cast<int>(s_.length()); ++j) {
|                                 ^~
|                                 p_
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp: In member function ‘pufferfish::CanonicalKmerIterator& pufferfish::CanonicalKmerIterator::operator++()’:
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/CanonicalKmerIterator.hpp:88:53: error: ‘s_’ was not declared in this scope; did you mean ‘p_’?
88 |     invalid_ = invalid_ || lpos >= static_cast<int>(s_.length());
|                                                     ^~
|                                                     p_
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/src/PufferfishIndexer.cpp: In function ‘int pufferfishIndex(pufferfish::IndexOptions&)’:
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/src/PufferfishIndexer.cpp:589:14: error: ‘string_view’ is not a member of ‘stx’; did you mean ‘std::experimental::fundamentals_v1::string_view’?
589 |         stx::string_view seqv(rp.seq);
|              ^~~~~~~~~~~
In file included from /var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/spdlog/fmt/bundled/format.h:69,
from /var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/spdlog/fmt/fmt.h:21,
from /var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/spdlog/common.h:47,
from /var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/spdlog/spdlog.h:13,
from /var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/Util.hpp:19,
from /var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/include/PufferfishBinaryGFAReader.hpp:5,
from /var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/src/PufferfishBinaryGFAReader.cpp:1,
from /var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/src/PufferfishIndexer.cpp:20:
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/include/g++-v9/experimental/string_view:565:9: note: ‘std::experimental::fundamentals_v1::string_view’ declared here
565 |   using string_view = basic_string_view<char>;
|         ^~~~~~~~~~~
/var/tmp/portage/sci-biology/salmon-1.4.0/work/salmon-1.4.0/external/pufferfish/src/PufferfishIndexer.cpp:593:38: error: ‘seqv’ was not declared in this scope
593 |         pf.encodeSeq(refseq, offset, seqv);
|                                      ^~~~
ninja: build stopped: subcommand failed.
```

AndrewAmmerlaan:
@waebbl @epsilon-0 :sob: another error:

```
/var/tmp/portage/sci-biology/pufferfish-1.4.0/work/pufferfish-salmon-v1.4.0/external/twopaco/graphconstructor/assemblyedgeconstructor.h: In constructor ‘TwoPaCo::AssemblyEdgeConstructor::AssemblyEdgeConstructor(const std::vector<std::__cxx11::basic_string<char> >&, const string&, const TwoPaCo::VertexEnumerator&)’:
/var/tmp/portage/sci-biology/pufferfish-1.4.0/work/pufferfish-salmon-v1.4.0/external/twopaco/graphconstructor/assemblyedgeconstructor.h:33:48: error: ‘bloomFilter’ was not declaredin this scope
33 |      assert(IsOutgoingEdgeInBloomFilter(hash, *bloomFilter, chr[i + edgeLength - 1]));
|                                                ^~~~~~~~~~~
ninja: build stopped: subcommand failed.
```

I give up on this one, the whole build system is a mess of things fetching other things, which also want to fetch even more things. And now it finally stopped complaining about not having internet access, and it still does not compile. If anyone feels like continuing where I left off, be my guest :)

waebbl:
I don't like such builds either. This looks like it's designed with a single user building this in $HOME, but not ready for a distribution build. Additionally, I'm currently quite busy with my own packages :/

PR #1015, opened 2020-12-31
sci-chemistry/chimera-bin: make repoman happy

AndrewAmmerlaan:

### DIFF
--- a/sci-chemistry/chimera-bin/chimera-bin-1.11.2.ebuild
+++ b/sci-chemistry/chimera-bin/chimera-bin-1.11.2.ebuild
@@ -8,10 +8,10 @@ inherit eutils
 DESCRIPTION="An extensible Molecular Modelling System"
 HOMEPAGE="http://www.cgl.ucsf.edu/chimera"
 SRC_URI="chimera-${PV}-linux_x86_64.bin"
+
 SLOT="0"
 LICENSE="chimera"
-KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
-IUSE=""
+KEYWORDS="~amd64 ~x86"
 
 DEPEND="prefix? ( dev-util/patchelf )"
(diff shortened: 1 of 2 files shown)


epsilon-0:
I'll merge this one, but I don't think you should work on dropping ~xxx-linux keywords.
They are used for prefix environments and rarely bother anyone.

AndrewAmmerlaan:
> I'll merge this one, but I don't think you should work on dropping ~xxx-linux keywords.
> They are used for prefix environments and rarely bother anyone.

Thanks 

I know the prefix keywords aren't really a problem, but the thing is that they do cause a lot of warnings in the repoman output and it makes it more difficult to distinguish between real problems and tiny problems like this, so I tend to fix it when I see it anyway. But I'll try to focus on more important things for now.

TheChymera:
@AndrewAmmerlaan can you post the errors you are getting from repoman? Is it equal visibility? If so, the solution might be to enable them on other builds, and not to start purging them.

AndrewAmmerlaan:
> @AndrewAmmerlaan can you post the errors you are getting from repoman? Is it equal visibility? If so, the solution might be to enable them on other builds, and not to start purging them.

Mostly errors like this:

```
   sci-biology/bx-python/bx-python-0.8.9.ebuild: BDEPEND: ~amd64-linux(default/linux/amd64/17.0/no-multilib/prefix/kernel-3.2+)
[     'dev-lang/python:3.6',
      'dev-lang/python:3.7',
      'dev-lang/python:3.8',
      'dev-lang/python:3.9']
```

because dev-lang/python does not have those keywords.

It's actually not that many packages that have problems with the ~amd64-linux keywords, all the packages that had these problems are already fixed or have an open PR to fix it. (snakemake, bx-python, ncbi-tools++, ncbi-blast+, salmon, and that's all)

PR #1016, opened 2020-12-31
sci-chemistry/modeller: make repoman happier

AndrewAmmerlaan:
The patch is still too big, but at least now it
doesn't complain about missing deps

RepoMan-Options: --force


epsilon-0:
thanks!

PR #1018, opened 2020-12-31
sci-libs/scikits_video: remove dep on libav, bump py targets

AndrewAmmerlaan:


epsilon-0:
@TheChymera 
as you were the one who added this in ::science, are you fine if I take this package and ask to move to ::gentoo?
There are a lot of scikit-* things in gentoo and it might be good to have this there so that they can be kept in sync.
I can open the PR to move it to ::gentoo, if you are fine with it.

PR #1020, opened 2020-12-31
app-doc/blacs-docs: EAPI bump, fix header

AndrewAmmerlaan:


epsilon-0:
ebuild fails at unpacking, could you try fixing that?

AndrewAmmerlaan:
I tried, but I'm confused with the error. It tells me to use the `-f` option, but there is no way to parse extra arguments to `unpack` (that I know), and the manual says not to call `tar` directly. The archive itself seems fine, I can open it with `ark`,

epsilon-0:
And it's telling me to use `-j` option:
```
tar: Archive is compressed. Use -j option
tar: Error is not recoverable: exiting now
```

AndrewAmmerlaan:
> And it's telling me to use `-j` option:
> 
> ```
> tar: Archive is compressed. Use -j option
> tar: Error is not recoverable: exiting now
> ```
> 
> oof, so weird.

Oops that's just a typo on my side

AndrewAmmerlaan:
I'm so confused :confused: , this should work but it doesn't. I can't find anything on google that suggest that we should treat .tar.xz any different than .tar.gz in ebuilds.

AndrewAmmerlaan:
Found a bug report for this: https://bugs.gentoo.org/756721

epsilon-0:
OK, so one nice thing, blacs is now integrated into scalapack and the standalone package no longer exists.
We can delete this package, can you update the PR for the deletion?

AndrewAmmerlaan:
> OK, so one nice thing, blacs is now integrated into scalapack and the standalone package no longer exists.
> We can delete this package, can you update the PR for the deletion?

Removed :smiley:

epsilon-0:
thanks!

PR #1023, opened 2020-12-31
Sci-chemistry/ctffind: fix and unmask

AndrewAmmerlaan:
This is the last one for today :smile:


epsilon-0:
Could you fix the name in the commit to ctffind

epsilon-0:
thanks!

AndrewAmmerlaan:
> Could you fix the name in the commit to ctffind

The name in the commit is correct, it is just the name of the branch and the title that is wrong (sorry about that). I don't think I can rename the branch without messing up this PR (and the name of the branch doesn't really matter, I think, right?)

PR #1024, opened 2021-01-01
Dev-java/jeuclid-core: fix patches, EAPI bump, add keyword

AndrewAmmerlaan:


dev-java/jeuclid-core/jeuclid-core-3.1.9-r1.ebuild
```diff
@@ -11,13 +11,13 @@ SRC_URI="mirror://sourceforge/jeuclid/jeuclid-parent-${PV}-src.zip"
...
-IUSE=""
+KEYWORDS="~amd64"
 
-COMMON_DEPEND="dev-java/ant-core
-	dev-java/batik:1.8=
-	dev-java/commons-logging
-	dev-java/jcip-annotations
+COMMON_DEPEND="
+	dev-java/ant-core:0
+	dev-java/batik:1.8
```
TheChymera, line 18: Does this not work with a newer batik? Wouldn't it be better to just bump this instead of bringing back an older batik?

AndrewAmmerlaan: > Does this not work with a newer batik? 

It did not when I tested it :(

> Wouldn't it be better to just bump this instead of bringing back an older batik?

That would be better, but this is the latest version, it hasn't been updated in ages. (it is however still a dependency of sci-mathmatics/scilab)

TheChymera: @AndrewAmmerlaan are you sure that this is not the same project? https://github.com/rototor/jeuclid/releases

AndrewAmmerlaan: > @AndrewAmmerlaan are you sure that this is not the same project? https://github.com/rototor/jeuclid/releases

It might very well be, thanks, I'll update the ebuild accordingly. Hopefully scilab will be compatible with this fork (I really wish scilab would document their dependencies better)

AndrewAmmerlaan: The fork says it is compatible with batik-1.9 and even 1.13. However when I emerge it I still get:

```
!!! ERROR: Package batik-1.8 was not found!
```

Or maybe I'm missing something obvious.

AndrewAmmerlaan: Never mind, I'm just being blind.

AndrewAmmerlaan: @TheChymera  Fixed, PR is updated, the package now compiles just fine with batik-1.9 :D

AndrewAmmerlaan: > Hopefully scilab will be compatible with this fork (I really wish scilab would document their dependencies better)

Sadly scilab still fails to compile with the same error with this newer version :cry:

TheChymera:
thaks!

PR #1030, opened 2021-01-02
net-misc/cernbox-client: Bump version to 2.7.4.2934

olifre:
This bumps the package to the latest version (matching `net-misc/owncloud-client`), pulls in the newest branding files, adapts to the new (simplified) way branding works, and drops the old version which failed to be fetched. 

Since CERNBox upstream sadly pulls the branding packages without prior announcement (they only ever keep one version), it would be best to have a way to mirror them so this can stay synchronized with `net-misc/owncloud-client` version bumps. Pinging @TheChymera , let me know if there is an approach I as contributor can use :smile: .


TheChymera:
@olifre do you have some server space somewhere? Since the authenticity is tracked via the Manifest, it's ok for the respective maintainer to just mirror the files. lease include this in the version bump, by adding a second line to SRC_URI with the address of the mirror. Otherwise please ping me again so that I can mirror it for you, but really, since you are the maintainer it would be best if this is in your hands. Eventually, if you are interested, you could also ask for push access (just open a bug on bugzilla).

olifre:
> @olifre do you have some server space somewhere?

@TheChymera Probably the most fitting webspace for this purpose I could use would be CERN-based web storage. While CERNBox itself is ownCloud-based and hence not well-suited, the backend storage (EOS) is also exportable via classic web services (including HTTPS). I can add it to my (not-yet-existing) personal site storage there, so I would be hosting the files on the very same storage the original files are hosted on, but with different retention policy ;-). 
I'll take care of that in the next days and update this PR accordingly. 


> Eventually, if you are interested, you could also ask for push access (just open a bug on bugzilla).

Thanks, I will consider it — most of the packages I maintain are now in Gentoo or Guru, but if ownCloud keeps updating at this rate, I'll probably indeed open a bug on bugzilla ;-).

olifre:
>  Please include this in the version bump, by adding a second line to SRC_URI with the address of the mirror.

@TheChymera Done in the force-pushed commits :-).

TheChymera:
@olifre sadly cernbox fails to build over here. Is it maybe due to some missing GCC version constraint?

[build.log](https://github.com/gentoo/sci/files/5762089/build.log)

olifre:
@TheChymera Thanks for the build log!
That was indeed an easy fix, my bad: Current `ownCloud` explicitly must depend on `kde-frameworks/extra-cmake-modules` even if not setting `USE=dolphin`. I've force-pushed a fix which sets the dependency correctly and also adapts the CMake flags accordingly.

PR #1021, opened 2020-12-31
sci-physics/genfit: EAPI bump, drop 9999

AndrewAmmerlaan:
This is the last one with `dependency.badindev`. It does not compile for me, I suspect because it is not compatible with new versions of sci-physics/root. But maybe there is an easy fix that I'm not seeing.

9999 is 404
cmake-utils --> cmake
use docs eclass to compile docs

Package-Manager: Portage-3.0.12, Repoman-3.0.2
Signed-off-by: Andrew Ammerlaan <andrewammerlaan@riseup.net>

```
/var/tmp/portage/sci-physics/genfit-02.00.00/work/genfit-02.00.00_build/utilitiesRootDict.cpp:174:1: error: ‘atomic_TClass_ptr’ does not name a type
174 | atomic_TClass_ptr HelixTrackModel::fgIsA(0);  // static to hold class pointer
| ^~~~~~~~~~~~~~~~~
/var/tmp/portage/sci-physics/genfit-02.00.00/work/genfit-02.00.00_build/utilitiesRootDict.cpp: In static member function ‘static TClass* genfit::HelixTrackModel::Dictionary()’:
/var/tmp/portage/sci-physics/genfit-02.00.00/work/genfit-02.00.00_build/utilitiesRootDict.cpp:197:4: error: ‘fgIsA’ was not declared in this scope; did you mean ‘IsA’?
197 |    fgIsA = ::ROOT::GenerateInitInstanceLocal((const ::genfit::HelixTrackModel*)0x0)->GetClass();
|    ^~~~~
|    IsA
/var/tmp/portage/sci-physics/genfit-02.00.00/work/genfit-02.00.00_build/utilitiesRootDict.cpp: In static member function ‘static TClass* genfit::HelixTrackModel::Class()’:
/var/tmp/portage/sci-physics/genfit-02.00.00/work/genfit-02.00.00_build/utilitiesRootDict.cpp:204:9: error: ‘fgIsA’ was not declared in this scope; did you mean ‘IsA’?
204 |    if (!fgIsA.load()) { R__LOCKGUARD(gInterpreterMutex); fgIsA = ::ROOT::GenerateInitInstanceLocal((const ::genfit::HelixTrackModel*)0x0)->GetClass(); }
|         ^~~~~
|         IsA
[...]

PR #1031, opened 2021-01-02
sci-physics/genfit: Add functional live ebuild.

olifre:
This also fixes documentation generation
and installation.

Signed-off-by: Oliver Freyermuth <o.freyermuth@googlemail.com>


olifre:
As mentioned in #1021 , documentation generation can not be done with the `docs` eclass, since CMake is used by the package to replace some macros in the `Doxyfile.in`, so I changed back to the `cmake`-based documentation generation. 

Upstream is sadly not doing releases:
https://github.com/GenFit/GenFit/issues/42
and the last tagged `02.00.00` package which is currently in this overlay does not work with any unmasked ROOT version in Gentoo portage anymore. 

I am not sure what is the best way to proceed — we could package a date-based tag to have a kind of release? 
Dropping the old `02.00.00` version which is not of much use anymore could be done afterwards.

AndrewAmmerlaan:
> I am not sure what is the best way to proceed — we could package a date-based tag to have a kind of release?
> Dropping the old `02.00.00` version which is not of much use anymore could be done afterwards.

This seems like a good idea to me, we could just add a 02.00.00_p20201231.ebuild and remove the old 02.00.00 version :+1:

olifre:
> This seems like a good idea to me, we could just add a 02.00.00_p20201231.ebuild and remove the old 02.00.00 version :+1:

Thanks! 
I've done exactly that in the linked commits. One note:

I also force-pushed the `9999` version again, removing all references in there to the old, pre-GitHub tag (so we get rid of the HTTP warnings by `repoman`). I've used `EGIT_COMMIT=""` in the unused code branch there, so when making a new tagged version, one can copy the `9999` ebuild over and fill in the `EGIT_COMMIT`. 

I have not added myself as maintaner (not sure if I should). I have a look at `genfit` from time to time, but am not actively using it anymore, so I can for sure help out from time to time, but it will not be my highest priority. If I should still add myself, just let me know.

TheChymera:
> 02.00.00_p20201231
Yes, that's generally the way to go if upstream does not version.

> I have not added myself as maintaner (not sure if I should).
If the package does not already have another maintainer, it means whatever attention level you're willing to prioritize it at is higher than what anybody else is willing to give it :) It also helps other maintainers to reach out to you if they have issues understanding the ebuild.

Please add yourself as a maintainer, and merge that into the version bump commit. Just for the sake of consistency and grep-ability, a commit text like `version bump 02.00.00_p20201231 snapshot` would be better :)

olifre:
> Please add yourself as a maintainer, and merge that into the version bump commit. 

Done, I merged that in the first commit adding the live `ebuild`. 

> Just for the sake of consistency and grep-ability, a commit text like `version bump 02.00.00_p20201231 snapshot` would be better :)

Indeed, that seems much more useful, I completely missed that since I did not consider it a "bump", but of course it is :wink:.

TheChymera:
closed by https://gitweb.gentoo.org/proj/sci.git/commit/?id=1f55f85dcb465ea7267a5362ac4ab3b5fc319c63 and preceding.

PR #1033, opened 2021-01-03
App-text/zotero: Your personal research assistant

AndrewAmmerlaan:
This is a very useful application I just found.

It helps find and manage citations, and exports to many formats (including bibtex), there is a LibreOffice and Firefox plug-in. It is written in java and the build system is not supported by portage so this is a `-bin` package. I also added the python interface while I was at it.

Loosely based on the existing ebuilds in the gerislay and xdch47 overlays: version bumped, EAPI bumped and style improvements.


TheChymera:
Good contribution, as always, but for anybody's personal use I would recommend considering that actually you only really need vim+BibTeX. There's also this nice resource for no-nonsens DOI to bibtex conversion: https://www.doi2bib.org/

I know it sounds bare-bones and I initially avoided it for years, but ultimately it has a certain elegance in simplicity and I would never go back to <insert-framework-here> :)

AndrewAmmerlaan:
> There's also this nice resource for no-nonsens DOI to bibtex conversion: https://www.doi2bib.org/

Wow, this is awesome, definitely bookmarking this one, Thanks!

PR #1040, opened 2021-01-13
Bump dev-libs/intel-common and dev-lang/icc to v19.1.3.304

### LINKED ISSUE #787: New ebuild for dev-lang/icc-17.0.x
Would it be possible to upgrade the icc ebuild to support "Parallel Studio XE 2017"? 

http://registrationcenter-download.intel.com/akdlm/irc_nas/tec/11537/parallel_studio_xe_2017_update4.tgz

abusse:
This should also resolves #787. With the introduction of Intel oneAPI they seem to have made changes to the structure of the legacy package. This made slight changes to the sdp eclass necessary. Furthermore, there seems no longer to be the Fortran compiler included or MPI libraries. I was only able to make a quick functionality check. Even though everything should be fine, further testing would be appreciated.


epsilon-0:
Intel MPI - https://github.com/gentoo/gentoo/pull/18452

AndrewAmmerlaan:
Could you please sign-off your commits? Can't merge without it, see:
https://wiki.gentoo.org/wiki/Talk:Gentoo_git_workflow#Sign-Off
https://www.gentoo.org/glep/glep-0076.html

abusse:
Sorry for taking so long. I totally overlooked that you asked me to sign off.

AndrewAmmerlaan:
So I bumped into this again while looking into some other intel/oneAPI related ebuilds. And it looks like for most of those things (e.g. sci-libs/mkl) we need the DPC++ compiler from here: https://github.com/intel/llvm which looks like it is the modern equivalent of dev-lang/icc. So I am wondering if we should remove dev-lang/icc and dev-libs/intel-common and package this DPC++ compiler instead so we can properly bump sci-libs/mkl.

AndrewAmmerlaan:
It took a lot of effort, but I finally packaged the DPC++ compiler. I removed the old icc/ifc ebuilds because they no longer seem to work and they are not very user-friendly to begin with since they require manual fetching. Therefore I am also closing the related Pull Requests. Thanks for your work here though, and sorry that this never got merged.

PR #1049, opened 2021-02-02
Commits for branch fsl-cstd-fixes

AbstractMethod:


TheChymera:
sorry for the delay, looking into this right now

TheChymera:
@AbstractMethod looks ok, can you please break this down into a PR for fslpy and another one for FSL? We prefer to do package-specific PRs if at all feasible, and it is def very feasible here :)

TheChymera:
@AbstractMethod actually, not entirely, ok, we have `thin-manifests = true` in `metadata/layout.conf` so the Manifest files should not include checksums.

PR #1050, opened 2021-02-03
added mmseqs2

joelb123:


joelb123:
This is my first submit to the science overlay by PR, so my apologies if things are a bit odd.  I thought my PR message was more detailed than it turned out to be.

This PR add sci-biology/mmseqs2, a fast and sensitive sequence search and clustering program:

- the metadata.xml contains my email as maintainer, upstream tags, and a long description
- repoman gives me no error messages
- build and install are clean, with no QA messages.  Flags are respected.   I recommend CFLAGS of "-march=native -O3"
- the binary contains a semantic versioning number consistent with upstream's rough usage

This is the first PR of several from my private repository, pybio [https://github.com/soedinglab/mmseqs2](url). If this process goes smoothly I have quite a few others.

epsilon-0:
Could you update the commit message to be complaint with gentoo git GLEP - https://www.gentoo.org/glep/glep-0066.html
What it means is that the commit message should be of form:
```
sci-biology/mmseqs2: fast and sensitive sequence analyzer

fast capabilities for sensitive sequence searching and clustering
Signed-off-by: Your Name <your-email@domain.tld>
```
Will add more comments.

sci-biology/mmseqs2/mmseqs2-12.0.ebuild
```diff
@@ -0,0 +1,46 @@
...
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE=""
+
+RDEPEND="
+	app-shells/bash-completion
+	sys-libs/zlib
+	app-arch/bzip2
+"
+DEPEND="${RDEPEND}"
```
epsilon-0, line 28: Almost surely, you don't need `app-shells/bash-completion` in `DEPEND` - https://devmanual.gentoo.org/general-concepts/dependencies/
Set 
```
DEPEND="
	sys-libs/zlib
	app-arch/bzip2
"
RDEPEND="${DEPEND}
	app-shells/bash-completion
"
```

sci-biology/mmseqs2/mmseqs2-12.0.ebuild
```diff
@@ -0,0 +1,46 @@
...
+
+S="${WORKDIR}/${MY_PN}-${FULL_PV}"
+
+src_configure() {
+	local mycmakeargs=(
+		-DBUILD_SHARED_LIBS=OFF
```
epsilon-0, line 34: Why are shared libs off?
The default gentoo pilicy is to always enable shared libs, instead of static-libs.

sci-biology/mmseqs2/mmseqs2-12.0.ebuild
```diff
@@ -0,0 +1,46 @@
...
+		-DVERSION_OVERRIDE=${PV}
+	)
+	cmake_src_configure
+}
+
+src_install(){
+	cmake_src_install
+	insinto /usr/share/bash-completion/completions/
```
epsilon-0, line 42: use the `bash-completions-r1` eclass - https://devmanual.gentoo.org/eclass-reference/bash-completion-r1.eclass/index.html

sci-biology/mmseqs2/mmseqs2-12.0.ebuild
```diff
@@ -0,0 +1,46 @@
...
+	mv util/bash-completion.sh util/mmseqs
+	doins util/mmseqs
+	rm -r "$D"/usr/util/
```
epsilon-0, line 45: What is this folder? Is this the original folder where it installed the bash-completion?
In that case, let it be there, its fine to have duplicates, this can serve as reference. 
Don't remove the folder.

epsilon-0 (changes requested):
here are some preliminary comments.

epsilon-0:
Also, if the repository is private, how are people going to be able to download this?

joelb123:
Sorry that I didn't get back to this earlier, and thanks for the edited build, @AndrewAmmerlaan.  In the 9 months since I wrote it the following things have happened:

1.  Upstream changed their release naming policy in response to my bitching.  The first 5 characters of the commit are now included in the release, which is still a bit of an ugh because those characters happen to look like digits on the v13 release but are actually hexits which will cause the normal versioning scheme to blow up for the next release.
2. The dependencies all need [static-libs] use flags.
3. The need for a live version has become apparent, as lots of important things have gotten committed.

I have ebuilds that address these.  Not sure on the best way to send them to you...post a git diff here or submit a whole new PR?

AndrewAmmerlaan:
PR is probably easiest

PR #1054, opened 2021-02-21
net-misc/cernbox-client: Bump to 2.7.6.3261. 

olifre:
Follows new version of net-misc/owncloud-client.


AndrewAmmerlaan:
Thanks!

PR #1056, opened 2021-02-22
Commits for fsl-6.0.4-cstd-fixes.

AbstractMethod:


AbstractMethod:
@TheChymera OK, I split the PR into 2 separate ones and closed/delete the other PR.

sci-biology/fsl/fsl-6.0.2-r1.ebuild
```diff
@@ -16,7 +16,8 @@ IUSE="cuda"
 
 DEPEND="
 	dev-libs/boost
-	dev-python/fslpy
+	<dev-python/fslpy-3
+	!>=dev-python/fslpy-3
```
TheChymera, line 20: these lines are equivalent, I'd just keep the first one, in both ebuilds.

TheChymera (changes requested):

PR #1057, opened 2021-02-23
app-offic/visidata: use distutils_enable_tests

ezzieyguywuf:


app-office/visidata/visidata-2.1.1.ebuild
```diff
@@ -53,6 +53,7 @@ python_prepare_all() {
 python_test() {
 	git init || die "Git init failed."
 	git add tests/golden/ || die "Git add failed."
+	# this test script eventually calls pytest under the hood
 	dev/test.sh || die "Tests failed."
 	rm .git -rf || die "Could not clean up git test directory."
```
epsilon-0, line 58: While here, could you also remove the `-f` please, and only make it `rm -r`. The `-f` makes it never fail.

epsilon-0 (changes requested):

app-office/visidata/visidata-2.1.1.ebuild
epsilon-0: whoops too slow on commenting.
Will change it myself.

PR #1063, opened 2021-03-11
HoloViz: the full Holoviews + Panel + Param + Colorcet stack

leycec:
This PR finalizes Gentoo support for [most of the HoloViz data visualization stack](https://holoviz.org), including:

* **[HoloViews](https://holoviews.org),** the *really* succinct visualization API that makes you never want to use Matplotlib ever again.
* **[Panel](https://panel.pyviz.org),** the pure-Python [WebSocket](https://en.wikipedia.org/wiki/WebSocket)-driven [Tornado](https://www.tornadoweb.org)-backed [Bokeh](https://bokeh.org)-based framework for stateful reactive data visualization dashboard web apps. Yeah.
* **[Param](https://param.holoviz.org),** the declarative decorator API that transforms Panel apps from procedural to object-oriented. It's good.
* **[Colorcet](https://colorcet.holoviz.org),** perceptually accurate colormaps leveraged by all of the above.

Thankfully, this overlay already provides the [core Bokeh server](https://docs.bokeh.org/en/latest/docs/user_guide/server.html) required by those packages. That makes life easy for everyone.

Tests are omitted. Some of those packages still test with `nose`, [which is dead and soon to be last-rited](https://fedoraproject.org/wiki/Changes/DeprecateNose). More importantly, supporting tests would've been so time-prohibitive I just never would've done it. The tests for HoloViews and Panel are... *really* involved. If no one objects, I'd like to defer that to later. :pray: 

Thanks for a decade of scientific packaging, everyone – especially @AndrewAmmerlaan for his constant dedication to the cause.


dev-python/colorcet/colorcet-2.0.6.ebuild
```diff
@@ -0,0 +1,26 @@
...
+	https://pypi.org/project/colorcet
+	https://github.com/holoviz/colorcet"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
+
+BDEPEND=">=dev-python/setuptools-30.3.0[${PYTHON_USEDEP}]"
```
thesamesam, line 21: We can avoid this, it's set by [default](https://dev.gentoo.org/~mgorny/python-guide/distutils.html#distutils-use-setuptools)

dev-python/colorcet/colorcet-2.0.6.ebuild
```diff
@@ -0,0 +1,26 @@
...
+DEPEND="
```
thesamesam, line 22: Are these needed at build time? (Sometimes the answer is yes, but it's often no if it's pure Python)

dev-python/colorcet/colorcet-2.0.6.ebuild
```diff
@@ -0,0 +1,26 @@
...
+DESCRIPTION="Collection of perceptually uniform colormaps"
+HOMEPAGE="
+	https://pypi.org/project/colorcet
+	https://github.com/holoviz/colorcet"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
```
thesamesam, line 19: We can drop [empty](https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-620007.3) IUSE (this is a nit, I just like giving people the link to PMS so they're aware of it - it's ok to keep it as-is).

dev-python/colorcet/metadata.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="project">
+		<email>sci@gentoo.org</email>
+		<name>Gentoo Science Project</name>
+	</maintainer>
+</pkgmetadata>
```
thesamesam, line 8: Could you [add](https://devmanual.gentoo.org/keywording/index.html#simultaneous-stabilization-on-all-architectures) ```<stabilize-allarches/>``` if it's pure Python (no compilation and so on)?

(Applies to all packages where appropriate, as with most of these comments - I'l try make clear if it's just one.)

dev-python/holoviews/holoviews-1.14.2.ebuild
```diff
@@ -0,0 +1,36 @@
...
+	https://pypi.org/project/holoviews
+	https://github.com/holoviz/holoviews"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
+
+BDEPEND=">=dev-python/setuptools-30.3.0[${PYTHON_USEDEP}]"
```
thesamesam, line 25: We can drop this and IUSE as ```DISTUTILS_USE_SETUPTOOLS=rdepend``` [includes](https://dev.gentoo.org/~mgorny/python-guide/distutils.html#distutils-use-setuptools) BDEPEND.

(This isn't true for the RDEPEND variable in general, it's just what the eclass does.)

dev-python/holoviews/holoviews-1.14.2.ebuild
```diff
@@ -0,0 +1,36 @@
...
+DEPEND="
```
thesamesam, line 26: Built-against or similar?

dev-python/holoviews/metadata.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="project">
+		<email>sci@gentoo.org</email>
+		<name>Gentoo Science Project</name>
+	</maintainer>
```
thesamesam, line 7: ALLARCHES?

dev-python/panel/metadata.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="project">
+		<email>sci@gentoo.org</email>
+		<name>Gentoo Science Project</name>
+	</maintainer>
+</pkgmetadata>
```
thesamesam, line 8: ALLARCHES? BTW, are you interested in maintaining any of these (too?) - if so, add yourself above sci

dev-python/panel/panel-0.11.0.ebuild
```diff
@@ -0,0 +1,43 @@
...
+	https://github.com/holoviz/panel"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
+
+BDEPEND="
+	>=dev-python/setuptools-30.3.0[${PYTHON_USEDEP}]
```
thesamesam, line 26: We can drop

dev-python/panel/panel-0.11.0.ebuild
```diff
@@ -0,0 +1,43 @@
...
+HOMEPAGE="
+	https://panel.holoviz.org
+	https://pypi.org/project/panel
+	https://github.com/holoviz/panel"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
```
thesamesam, line 23: Drop

dev-python/panel/panel-0.11.0.ebuild
```diff
@@ -0,0 +1,43 @@
...
+RDEPEND="${DEPEND}
+	dev-python/markdown[${PYTHON_USEDEP}]
+	dev-python/requests[${PYTHON_USEDEP}]
+	dev-python/tqdm[${PYTHON_USEDEP}]
+	>=dev-python/pyviz_comms-0.7.4[${PYTHON_USEDEP}]
+"
+
+# Permit Panel to violate network sandboxing, as "setup.py" invokes "npm" to
+# fetch and bundle Node.js assets with Panel.
+RESTRICT=network-sandbox
```
thesamesam, line 43: We're going to add this to [EAPI 8 ](https://wiki.gentoo.org/wiki/Future_EAPI/EAPI_8_tentative_features)but for now, we need just RESTRICT="test" to be [PMS-compliant](https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-680007.3.6).

dev-python/param/metadata.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="project">
+		<email>sci@gentoo.org</email>
+		<name>Gentoo Science Project</name>
+	</maintainer>
+</pkgmetadata>
```
thesamesam, line 8: allarches? interested in mailing yourself?

dev-python/param/param-1.10.1.ebuild
```diff
@@ -0,0 +1,20 @@
...
+HOMEPAGE="
+	http://param.holoviz.org
+	https://pypi.org/project/param
+	https://github.com/holoviz/param"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="BSD"
+KEYWORDS="~amd64 ~x86"
+SLOT="0"
+IUSE=""
```
thesamesam, line 20: drop

dev-python/param/param-1.10.1.ebuild
```diff
@@ -0,0 +1,20 @@
...
+
+PYTHON_COMPAT=( python3_{7..9} pypy3 )
+
+inherit distutils-r1
+
+DESCRIPTION="Make your Python code clearer and more reliable by declaring Parameters"
+HOMEPAGE="
+	http://param.holoviz.org
+	https://pypi.org/project/param
+	https://github.com/holoviz/param"
```
thesamesam, line 14: Oh, that reminds me. We can add pypi and GitHub as remote-ids in [metadata.xml](https://devmanual.gentoo.org/ebuild-writing/misc-files/metadata/) instead.

dev-python/pyct/pyct-0.4.8.ebuild
```diff
@@ -0,0 +1,22 @@
...
+	https://pypi.org/project/pyct
+	https://github.com/pyviz-dev/pyct"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
+
+DEPEND=">=dev-python/param-1.7.0[${PYTHON_USEDEP}]"
```
thesamesam, line 21: Definitely needed at build time?

dev-python/pyct/pyct-0.4.8.ebuild
```diff
@@ -0,0 +1,22 @@
...
+DESCRIPTION="Python packaging Common Tasks"
+HOMEPAGE="
+	https://pypi.org/project/pyct
+	https://github.com/pyviz-dev/pyct"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
```
thesamesam, line 19: Drop

dev-python/pyct/metadata.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="project">
+		<email>sci@gentoo.org</email>
+		<name>Gentoo Science Project</name>
+	</maintainer>
+</pkgmetadata>
```
thesamesam, line 8: ALLARCHES if applicable, remote-ids, maintain yourself if interested?

dev-python/pyviz_comms/pyviz_comms-2.0.1.ebuild
```diff
@@ -0,0 +1,21 @@
...
+DESCRIPTION="Bidirectional communication for the HoloViz ecosystem"
+HOMEPAGE="
+	https://pypi.org/project/pyviz_comms
+	https://github.com/holoviz/pyviz_comms"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
```
thesamesam, line 19: drop

dev-python/pyviz_comms/metadata.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="project">
+		<email>sci@gentoo.org</email>
+		<name>Gentoo Science Project</name>
+	</maintainer>
+</pkgmetadata>
```
thesamesam, line 8: allarches if appropriate, remote-ids, maintain yourself if interested?

thesamesam (commented):
This looks like a great chunk of work, thanks!

I'm not familiar with HoloViz so I'll let someone else approve/merge but I've left some general Python packaging tips. They're mostly nit-picks but I thought I'd say them so you know -- a lot of it was a good excuse to link to helpful docs which we need to better advertise though.

Let me know if you have any questions but otherwise I'll leave it to the experts. Cheers!

leycec:
Thanks for the incisive nitpicks, @thesamesam! I have now levelled up my Python packaging and am grateful for it.

> EDIT: If you can for any easy ones, please add tests with `distutils_enable_tests`.

...you're right. I should do that. Parsing the 100-line `[tests]` extra in `setup.py` for Panel and HoloViews made my eyes bleed. I gave up shortly thereafter. But something is better than nothing; as you say, enabling tests for all of the *other* packages should be trivial.

Let's give that a go. :+1:

dev-python/colorcet/colorcet-2.0.6.ebuild
leycec: Gotcha. <sup>I *thought* that might be the case, but wasn't quite sure.</sup>

dev-python/colorcet/colorcet-2.0.6.ebuild
leycec: Sadly, yes. `setup.py` for HoloViz packages is more convoluted than expected (or needed, frankly). [The Colorcet `setup.py` contains this snippet](https://github.com/holoviz/colorcet/blob/ae9deb9e27cb73cc274ce15fa1a9ec8aa8b557dd/setup.py#L16), for example:

```
import pyct.build

def get_setup_version(reponame):
    ...
    try:
        from param import version
    except:
        version = None
    if version is not None:
        return version.Version.setup_version(basepath, reponame, archive_commit="$Format:%h$")
    else:
        print("WARNING: param>=1.6.0 unavailable. If you are installing a package, this warning can safely be ignored. If you are creating a package or otherwise operating in a git repository, you should install param>=1.6.0.")
```

We could probably patch out the `get_setup_version()` call and thus the install-time `param` import, but... the `pyct.build` system might dynamically hook into that import somehow.

Their build workflow looks suspiciously fragile to me, so it might be better to just leave this as is.

dev-python/colorcet/colorcet-2.0.6.ebuild
leycec: Nice! Nice nitpick, too. I'm all for dropping extraneous lines.

dev-python/colorcet/metadata.xml
leycec: **Nice!** I'll do so. Shamefully, this is the first I've heard of `<stabilize-allarches/>`. I will now use that everywhere.

There's an interesting parallel here with `conda-forge`, the automated packaging system for third-party Anaconda packages. `conda-forge` recipes support a [similar `noarch: python` directive](https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-builds):

> Noarch packages are packages that are not architecture specific and therefore only have to be built once. Declaring these packages as `noarch` in the `build` section of the `meta.yaml` reduces shared CI resources. Therefore all packages that qualify to be `noarch packages`, should be declared as such.

dev-python/holoviews/holoviews-1.14.2.ebuild
leycec: You got it. :+1:

dev-python/holoviews/holoviews-1.14.2.ebuild
leycec: Right. I'll drop the `BDEPEND`, but the `DEPEND` probably needs to stay for consistency with their circuitous `setup.py` workflow. It is tragic.

dev-python/holoviews/metadata.xml
leycec: It shall be done, Master.

dev-python/panel/metadata.xml
leycec: That's... a good question. My volunteer time is fairly scarce, but that's true for everyone here. Rather than officially signing up as proxy maintainer, I'd kinda prefer to occasionally push PRs bumping ebuilds at my leisure.

With apologies, would you mind if we left this as is for the moment?

dev-python/panel/panel-0.11.0.ebuild
leycec: By your command, it shall be done.

dev-python/panel/panel-0.11.0.ebuild
leycec: It's dead.

dev-python/panel/panel-0.11.0.ebuild
leycec: I'll patch that right up. Thanks for those topical links, too. [I'm with Górny](https://bugs.gentoo.org/553696#c5):

> Let's extend `RESTRICT` syntax to make it `token[:reason]`, e.g. `test:network`, `test:docker`. The idea behind that is that the first token is well-defined and always work, while the second one can have implementation-defined meaning. We can then add new reasons without any changes to the PMS, and old PMs will work just fine (assuming `RESTRICT=test` is effective).

dev-python/param/metadata.xml
leycec: Yes to question one. "Can I get a rain check?" to question two.

dev-python/param/param-1.10.1.ebuild
leycec: It's so dropped.

dev-python/param/param-1.10.1.ebuild
leycec: Oh, I forgot about `remote-id`. I'll make that happen everywhere.

dev-python/pyct/pyct-0.4.8.ebuild
leycec: Definitely. I solemnly swear it on my fat Maine Coon cat.

dev-python/pyct/pyct-0.4.8.ebuild
leycec: So long, `IUSE`. We'll miss you.

dev-python/pyct/metadata.xml
leycec: Yes, yes, I'd rather stab myself with a plastic spork.

dev-python/pyviz_comms/pyviz_comms-2.0.1.ebuild
leycec: You're dead meat, `IUSE`!

dev-python/pyviz_comms/metadata.xml
leycec: Yes, yes, spork-stabbing.

AndrewAmmerlaan:
This looks good @leycec :+1: 

While you're changing things anyway, could you please add a `Signed-off-by:` line to the commit message, otherwise the PR-merging tool will complain: https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin (Using `repoman ci` instead of `git commit` will append this line automatically if `SIGNED_OFF_BY=""` is set in `make.conf`, another way to do it automatically is to use `git commit` with the `--signoff` or `-s` argument)

AndrewAmmerlaan:
Repoman is telling me that:   
```
RESTRICT.invalid              1
   dev-python/panel/panel-0.11.0.ebuild: network-sandbox
```
Are you sure this works as intended, because the ebuild manual does not list 'network-sandbox' as a valid entry for the RESTRICT variable? Usually if the tests require network access, RESTRICT="test" is added to just disable the tests.

```
  dependency.badindev           1
   dev-python/panel/panel-0.11.0.ebuild: BDEPEND: ~amd64(default/linux/amd64/17.0/x32)
['>=net-libs/nodejs-15.11.0']
```
We should mask `dev-python/panel` on this profile, since the nodejs dependency is unavailable. The GURU overlay contains a working example on how to mask packages on a specific profile in an overlay, it is a bit of a strange and complex procedure so let me know if you need any help with it: https://github.com/gentoo/guru/tree/master/profiles

leycec:
Done *and* done. Commit 5e52673 resolves all the bad things, including:

* Added `Signed-off-by:` line to commit message.
* Removed `IUSE=""` everywhere.
* Removed `>=dev-python/setuptools-30.3.0[${PYTHON_USEDEP}]` everywhere.
* Replaced extraneous `${HOMEPAGE}` URLs with `<remote-id...>` tags in all `metadata.xml`.
* Added `<stabilize-allarches/>` to all `metadata.xml`.
* Added `distutils_enable_tests pytest` to `dev-python/{colorcet,pyct}`. For the other packages:
  * `dev-python/param` still expects `nose`. It is shameful.
  * `dev-python/pyviz_comms` has no tests. It is more shameful.
  * `dev-python/{holoviews,panel}` have eye-bleeding tests.
* Masked `dev-python/panel` on the `default/linux/amd64/17.0/x32` profile.
* Replaced `RESTRICT=network-sandbox` in `dev-python/panel` with:
  * In `src_prepare()`, a `sed` patch preventing `setup.py` from invoking `npm` at installation time.
  * In `pkg_postinst()`, one `elog` message for each enabled Python implementation detailing how to manually rebuild Node.js packages after installation.

That should make all the bad things go away. Thanks for the prompt, detailed, and overly generous review, @thesamesam and @AndrewAmmerlaan. I appreciate it. Really.

Also, [GURU](https://github.com/gentoo/guru) looks hot – filling a void left by the long-dead Sunrise overlay. Thanks for maintaining that on behalf of the community, @AndrewAmmerlaan. :heart:

## The Horror of "RESTRICT=network-sandbox"

For those curious about `RESTRICT=network-sandbox`, here's a writeup that doesn't matter because we've solved everything but maybe somebody cares.

`RESTRICT=network-sandbox` is *technically* a valid `RESTRICT` feature (as described by `man 5 ebuild`). I say *technically*, because that feature is rightly prohibited in the Portage tree. That's probably what `repoman` was complaining about, except `RESTRICT=network-sandbox` isn't invalid; it's just bad.

Here's why we needed that before but don't now. For future posterity (and my future self when I bump into this again), here's what Panel is doing:

* Panel bundles Node.js packages in its tarball. This is good.
* Panel's unpatched `setup.py`:
  1. Thinks these bundled Node.js packages are desynchronized in some way with the currently installed version of Bokeh. This is bad.
  1. Invokes `npm`, the Node.js package manager. This is bad.
  1. `npm` tries to redownload *all* of these packages. This is bad.
* `src_prepare()` for Panel now elides this harmful behaviour by patching `setup.py` to avoid performing that logic. This is good.
* `pkg_postinst()` for Panel now details how to manually perform that logic with a simple CLI one-liner after installation, just in case we're wrong and somebody actually needs to do this. This is good.
 
I've verified this works by serving a toy Panel app. It works *without* having to run that one-liner, which means that Panel is wrong when it thinks its bundled Node.js packages are desynchronized in some way with the currently installed version of Bokeh.

We suspect Panel developers were trying to minimize issue churn on their bug tracker by forcing Node.js rebuilds on each Panel installation, which had the unintended side effect of preventing Panel from being packaged by any system package manager. *Woops.*

We should file an upstream issue. Sadly, I am lazy.

leycec:
Done *and* done. Commit 5e52673 resolves all the bad things, including:

* Added `Signed-off-by:` line to commit message.
* Removed `IUSE=""` everywhere.
* Removed `>=dev-python/setuptools-30.3.0[${PYTHON_USEDEP}]` everywhere.
* Replaced extraneous `${HOMEPAGE}` URLs with `<remote-id...>` tags in all `metadata.xml`.
* Added `<stabilize-allarches/>` to all `metadata.xml`.
* Added `distutils_enable_tests pytest` to `dev-python/{colorcet,pyct}`. For the other packages:
  * `dev-python/param` still expects `nose`. It is shameful.
  * `dev-python/pyviz_comms` has no tests. It is more shameful.
  * `dev-python/{holoviews,panel}` have eye-bleeding tests.
* Masked `dev-python/panel` on the `default/linux/amd64/17.0/x32` profile.
* Replaced `RESTRICT=network-sandbox` in `dev-python/panel` with:
  * In `src_prepare()`, a `sed` patch preventing `setup.py` from invoking `npm` at installation time.
  * In `pkg_postinst()`, one `elog` message for each enabled Python implementation detailing how to manually rebuild Node.js packages after installation.

That should make all the bad things go away. Thanks for the prompt, detailed, and overly generous review, @thesamesam and @AndrewAmmerlaan. I appreciate it. Really.

Also, [GURU](https://github.com/gentoo/guru) looks hot – filling a void left by the long-dead Sunrise overlay. Thanks for maintaining that on behalf of the community, @AndrewAmmerlaan. :heart:

## The Horror of "RESTRICT=network-sandbox"

For those curious about `RESTRICT=network-sandbox`, here's a writeup that doesn't matter because we've solved everything but maybe somebody cares.

`RESTRICT=network-sandbox` is *technically* a valid `RESTRICT` feature (as described by `man 5 ebuild`). I say *technically*, because that feature is rightly prohibited in the Portage tree. That's probably what `repoman` was complaining about, except `RESTRICT=network-sandbox` isn't invalid; it's just bad.

Here's why we needed that before but don't now. For future posterity (and my future self when I bump into this again), here's what Panel is doing:

* Panel bundles Node.js packages in its tarball. This is good.
* Panel's unpatched `setup.py`:
  1. Thinks these bundled Node.js packages are desynchronized in some way with the currently installed version of Bokeh. This is bad.
  1. Invokes `npm`, the Node.js package manager. This is bad.
  1. `npm` tries to redownload *all* of these packages. This is bad.
* `src_prepare()` for Panel now elides this harmful behaviour by patching `setup.py` to avoid performing that logic. This is good.
* `pkg_postinst()` for Panel now details how to manually perform that logic with a simple CLI one-liner after installation, just in case we're wrong and somebody actually needs to do this. This is good.
 
I've verified this works by serving a toy Panel app. It works *without* having to run that one-liner, which means that Panel is wrong when it thinks its bundled Node.js packages are desynchronized in some way with the currently installed version of Bokeh.

We suspect Panel developers were trying to minimize issue churn on their bug tracker by forcing Node.js rebuilds on each Panel installation, which had the unintended side effect of preventing Panel from being packaged by any system package manager. *Woops.*

We should file an upstream issue. Sadly, I am lazy.

AndrewAmmerlaan:
Thanks @leycec

leycec:
Thanks! Just ping me (@leycec) when these need bumping. :+1:

leycec:
@philippjfr: Gentoo Linux now quasi-officially packages most of HoloViz. Just thought you'd like to know. :wink:

philippjfr:
Thanks, really appreciate your efforts here. I'm not following all the details here but let me know if you need me to explain any of the reasoning behind some of the packaging decisions in Panel in particular.

leycec:
Sure thing! I'd intended to open a Panels issue with a few feature requests to simplify packaging by system package managers (e.g., `apt` under Debian, `pacman` under Arch). But this is easier, since we're all here and willing. :smile: 

The big "pain point" for us are `setup.py` scripts that open network connections, which we prohibit for both security and stability reasons. Panel's `setup.py` invokes `npm`, which opens network connections. Cue pain.

Ideally, we'd like Panel's `setup.py` to avoid downloading Node.js packages. After all, aren't those packages already bundled with Panel? If that's infeasible, we'd like Panel's `setup.py` to permit callers to conditionally disable that functionality: e.g.,

* Via an environment variable like `NO_NPM=1`.
* Via a command-line option like `--no-npm`.

Do you think either of those two things might be feasible? If not, no worries. We currently patch Panel's `setup.py` with a `sed` one-liner removing all calls to the `_build_paneljs()` function. That's a bit fragile, so official support at your leisure would be fantastic! :heart_eyes: 

Lastly, it'd also be great if Panel's `setup.py` allowed callers to specify a system prefix for `data_files`. Currently, the `data_files` in your `setup.py` are relative rather than absolute:

```
    data_files=[
        # like `jupyter serverextension enable --sys-prefix`
        (
            "etc/jupyter/jupyter_notebook_config.d",
```

That installs Panel's Jupyter configuration into `/usr/etc/jupyter/` rather than `/etc/jupyter/`. Naturally, Gentoo expects the former. We'd like Panel's `setup.py` to permit callers to conditionally specify a system prefix: e.g.,

* Via an environment variable like `SYS_PREFIX="/"`.
* Via a command-line option like `--sys-prefix="/"`.

The above `data_files` declaration would then become:

```
    data_files=[
        # like `jupyter serverextension enable --sys-prefix`
        (
            f"{sys_prefix}etc/jupyter/jupyter_notebook_config.d",
```

...*or something.* We leave this in your capable hands. Thanks for all the data visualization, Phil. HoloViz rocks the science sphere.

PR #1065, opened 2021-03-14
dev-cpp/blitz: 1.0.3_pre fix boost, doxygen dependencies

band-a-prend:
* The `blitz-1.0.2_p20200524` provides `libblitz.so.1.0.3` therefore
the new ebuild is renamed to `blitz-1.0.3_pre20200524` instead of revision bump while both use the same git-commit-sha-tarball.
* Blitz++ optionally depends on `dev-libs/boost[static-libs]`
to allow build with `Boost::serialization` support
(note: cmake option is realy with typo).
`USE="boost"` is used to handle this option.

* Add `USE="doc"` `app-doc/doxygen[dot]` build dependency to build dev docs.

* Add python build time dependency.

* The `testsuite`, `benchmark`, `examples` targets now is only conditionally build
in `src_compile()` and tests are ran in implicit `src_test()` phase.

* Update live ebuild.


band-a-prend:
It's seems that using of boost serialization is optional, so it's need to define with use flag.

dev-cpp/blitz/blitz-1.0.2_p20200524-r1.ebuild
```diff
@@ -0,0 +1,66 @@
...
+BDEPEND="${PYTHON_DEPS}"
+DEPEND="
+	dev-libs/boost:=
+	doc? ( app-doc/doxygen[dot] )
+"
+
+src_prepare() {
+	default
+	# provide "libblitz-1.0.2.so" instead of "libblitz-1.0.3.so"
+	sed -i -e "s/blitz_PATCH 3/blitz_PATCH 2/" CMakeLists.txt || die
```
epsilon-0, line 38: lets not do this, there's no specific reason and does not provide any benefit,

dev-cpp/blitz/blitz-1.0.2_p20200524-r1.ebuild
```diff
@@ -0,0 +1,66 @@
...
+
+LICENSE="BSD LGPL-3+"
+SLOT="0"
+IUSE="doc test"
+RESTRICT="!test? ( test )"
+
+BDEPEND="${PYTHON_DEPS}"
+DEPEND="
+	dev-libs/boost:=
+	doc? ( app-doc/doxygen[dot] )
```
epsilon-0, line 32: `doxygen` needs to be in `BDEPEND`.

dev-cpp/blitz/blitz-1.0.2_p20200524-r1.ebuild
```diff
@@ -0,0 +1,66 @@
...
+	local mycmakeargs=(
+		-DBUILD_TESTING=$(usex test)
+		-DBUILD_DOC=$(usex doc)
+	)
+	use doc && mycmakeargs+=( -DDISABLE_REFMAN_PDF=ON )
+	cmake_src_configure
+}
+
+src_compile() {
+	cmake_build
```
epsilon-0, line 53: `cmake_src_compile`

dev-cpp/blitz/blitz-1.0.2_p20200524-r1.ebuild
```diff
@@ -0,0 +1,66 @@
...
+	use doc && cmake_build blitz-doc
+}
+
+src_test() {
+	cmake_build check-testsuite check-benchmarks check-examples
```
epsilon-0, line 58: does it not need `cmake_src_test`?

epsilon-0 (commented):
Thanks a lot.
Have left a few comments.

Squash both commits into one.

Drop the previous version of the dated ebuild, in a separate commit.

dev-cpp/blitz/blitz-1.0.2_p20200524-r1.ebuild
band-a-prend: It slightly confuse me that package is 1.0.2 but installed library is 1.0.3.
Maybe then to rename ebuild to `blitz-1.0.3_pre20200524`, isn't?
And then to drop old dated ebuild.

epsilon-0: That sounds good :+1:

band-a-prend: Done

dev-cpp/blitz/blitz-1.0.2_p20200524-r1.ebuild
band-a-prend: Done

dev-cpp/blitz/blitz-1.0.2_p20200524-r1.ebuild
band-a-prend: Done. With addition of conditionally compiled `doc` and `test` targets.

dev-cpp/blitz/blitz-1.0.2_p20200524-r1.ebuild
band-a-prend: `cmake_src_test` tries to run precompiled testsuite, benchmark and examples tests. So the build of this targets is placed in `src_compile()` under appropriate condition.

band-a-prend:
Boost::serialization support requires `dev-libs/boost[static-libs]` as it is checked in `src/CMakeLists.txt` if it's static library is presented in system.

AndrewAmmerlaan:
> Boost::serialization support requires `dev-libs/boost[static-libs]` as it is checked in `src/CMakeLists.txt` if it's static library is presented in system.

Sometimes you can just `sed s/.a/.so/g' in those cases (and sometimes that will break everything)

band-a-prend:
> Sometimes you can just `sed s/.a/.so/g` in those cases (and sometimes that will break everything)

This option is used and affected this
```
set(Boost_USE_STATIC_LIBS        ON) # only find static libs
```
The blitz  unconditionally provides static library build that also build in boost::serialization in case under review.

In some cases static linking is usefull in scientific application then it build on platform with differ runtime and use just to run on other system. I'm not a fan of this approach but I encountered such case several times.

dev-cpp/blitz/blitz-1.0.3_pre20200524.ebuild
```diff
@@ -0,0 +1,53 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+CMAKE_MAKEFILE_GENERATOR=emake
```
thesamesam, line 6: We really need this? :(

dev-cpp/blitz/blitz-1.0.3_pre20200524.ebuild
```diff
@@ -0,0 +1,53 @@
...
+PYTHON_COMPAT=( python3_{7..9} )
+
+inherit cmake python-any-r1
+
+DESCRIPTION="Multi-dimensional array library for C++"
+HOMEPAGE="https://github.com/blitzpp/blitz"
+
+COMMIT=39f885951a9b8b11f931f917935a16066a945056
```
thesamesam, line 14: Standard structure has DESCRIPTION/HOMEPAGE/SRC_URI/S together, then a blank newline, then LICENSE/SLOT/KEYWORDS/IUSE/RESTRICT.

thesamesam: I'd also quote COMMIT and put it above DESCRIPTION.

dev-cpp/blitz/blitz-1.0.3_pre20200524.ebuild
band-a-prend: Usually it requires if Fortran code is compiled. I.e. Sundials had compilation problems with ninja.

In blitz++ the Fortran compiler is required only optionally to build fortran benckmark that is used to compare speed of blitz++ with pure fortran array operations. 

In this ebuild this benchmark isn't built so maybe blitz++ could use ninja. I need to recheck it.

dev-cpp/blitz/blitz-1.0.3_pre20200524.ebuild
```diff
@@ -0,0 +1,53 @@
...
+DEPEND="
+	boost? ( dev-libs/boost:=[static-libs] )
+"
+
+src_configure() {
+	local mycmakeargs=(
+		-DBUILD_TESTING=$(usex test)
+		-DBUILD_DOC=$(usex doc)
+	)
+	use boost && mycmakeargs+=( -DENABLE_SERIALISATION=ON )
```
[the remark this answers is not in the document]
band-a-prend: Ready

epsilon-0:
I have to fix my gentoo machine and will merge this when I get it fixed :heart:

band-a-prend:
@epsilon-0 
Ping!

epsilon-0:
thanks! merged :D

PR #1081, opened 2021-05-25
dev-perl/BTLib: add blocker

lucasmitrak:
* add blocker against app-misc/sphinx.
* added for both versions 0.19 and 0.20.

Currently, the package dev-perl/BTLib has file collisions with
app-misc/sphinx; therefore, they cannot be merged at on the same
system.
This commit adds blockers to all versions of dev-perl/BTLib
against app-misc/sphinx.
With this commit, they will not be able to be installed on the same
system, which solves the file collision problem.
Without this commit, a file collision detection will occur during
emerge.
This bug is similar to and was solved in the same method to
Bug: https://bugs.gentoo.org/756865.
This commit was tested locally within a docker w/ ebuildtester.
This commit was tested, written, and submitted by Lucas Mitrak.

Closes: https://bugs.gentoo.org/756796

### DIFF
--- a/dev-perl/BTLib/BTLib-0.19.ebuild
+++ b/dev-perl/BTLib/BTLib-0.19.ebuild
@@ -13,4 +13,7 @@ LICENSE="|| ( Artistic GPL-1 GPL-2 GPL-3 )"
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
 
+RDEPEND="
+	!app-misc/sphinx"
+
 SRC_TEST="do"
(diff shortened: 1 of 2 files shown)


AndrewAmmerlaan:
Thanks,

One small thing, could you add a `sign-off` line to your commit message please? This is required in order to adhere to Gentoo's copyright policy, and it also prevents the merging tool from complaining that it is missing. See here for details: https://www.gentoo.org/glep/glep-0076.html

AndrewAmmerlaan:
> You're welcome. Please feel free to give me any additional feedback as well. I added a `Signed-off-by` with a `git commit --amend` and then `git push origin BTLib -f`. Let me know if I did it correctly, and what steps to take if not.

Perfect, Thank You! :+1:

PR #1082, opened 2021-05-27
Fix all the Numba badness

### LINKED ISSUE #1080: Numba 0.53.1 installation has horrible race conditions that non-deterministically blow up
`dev-python/numba-0.53.1.ebuild` currently has an unversioned TBB requirement:

```
RDEPEND="
	...
	threads? ( dev-cpp/tbb )
"
```

Unfortunately, that's not quite right. Numba requires a minimum TBB version. Installation fails when the current TBB version is insufficient with compile-time errors like:

```
numba/np/ufunc/tbbpool.cpp:31:2: error: #error "TBB version is too old, 2019 update 5, i.e. TBB_INTERFACE_VERSION >= 11005 required"
   31 | #error "TBB version is too old, 2019 update 5, i.e. TBB_INTERFACE_VERSION >= 11005 required"
      |  ^~~~~
```

From that message, we can optimistically deduce this ebuild should instead require at least TBB >= 2019.5:

```
RDEPEND="
	...
	threads? ( >=dev-cpp/tbb-2019.5 )
"
```

Thanks as always for the extraordinary volunteerism, @AndrewAmmerlaan and crew! :+1:

leycec:
This PR resolves #1080 for Numba by:

* Disabling parallel build completely, which is also:
  * [What we do for SciPy](https://github.com/gentoo/gentoo/commit/809afd4a24533311ced5ecfb2f022b539a3b6dd2).
  * What we should probably do for Pandas, too.
* Add minimum dependency versions.
* Documents where exactly those versions come from. The source of the minimum TBB version is *super* nonsense, so that definitely needs a name-drop. It was also surprising to me that Numba unconditionally requires `pip` at runtime... but then I ran `numba -s` and remembered why. `</sigh>`

Shout-outs to @kiwifb and @epsilon-0 for their long suffering all these years.

### DIFF
--- a/dev-python/numba/numba-0.53.1.ebuild
+++ b/dev-python/numba/numba-0.53.1.ebuild
@@ -18,13 +18,20 @@ SLOT="0"
 KEYWORDS="~amd64 ~arm ~arm64 ~x86"
 IUSE="openmp threads"
 
+# Minimum dependency versions derive from:
+# * For llvmlite, the "min_llvmlite_version" and "max_llvmlite_version" globals
+#   in "setup.py".
+# * For NumPy, the "min_numpy_run_version" global in "setup.py".
+# * For TBB, #error pragmas in "numba/np/ufunc/tbbpool.cpp" like:
+#   #error "TBB version is too old, 2019 update 5...
+# Lastly, note the "numba -s" subcommand requires "pip" at runtime.
 RDEPEND="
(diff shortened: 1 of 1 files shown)


AndrewAmmerlaan:
Thanks :+1:

leycec:
Thank *you*, @AndrewAmmerlaan. You continue to amaze and stun us all with your relentless – almost robotic, dare I say – scientific volunteerism. :astonished:

PR #1083, opened 2021-05-29
 sci-libs/fox: add inherit fortran-2

lucasmitrak:
* inherit fortran-2 to sci-libs/fox

Currently, sci-libs/fox does not require a fortran compiler
to be installed. If sci-libs/fox is emerged without the fortran
compiler, the configure phase will check for a fortran compiler, and the
configure phase will fail. For example, the hardened profile disables
the fortran USE flag in sys-devel/gcc, so sci-libs/fox will fail.
With this commit, sci-libs/fox will require a fortran compiler
to be installed.
Without this commit, sci-libs/fox will fail on any system which does
not have the fortran compiler, which by default is all hardened profiles.

Closes: https://bugs.gentoo.org/757771
Signed-off-by: Lucas Mitrak lucas@lucasmitrak.com

### DIFF
--- a/sci-libs/fox/fox-4.1.2-r2.ebuild
+++ b/sci-libs/fox/fox-4.1.2-r2.ebuild
@@ -28,6 +28,8 @@ S="${WORKDIR}/${MY_P}"
 
 FORTRAN_STANDARD=90
 
+DEPEND="sys-devel/gcc[fortran]"
+
 PATCHES=(
 	"${FILESDIR}"/4.1.2-r2-install-customizations.patch
 )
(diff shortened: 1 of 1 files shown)


lucasmitrak:
I should add that the description of sci-libs/fox is "A library designed to allow the easy use of XML from Fortran", so it makes sense why a fortran compiler is needed.

waebbl:
I think the package should `inherit fortran-2`. You won't need the dependency on `gcc[fortran]` then.

lucasmitrak:
> 
> 
> I think the package should `inherit fortran-2`. You won't need the dependency on `gcc[fortran]` then.

Thank you for your comment. I added another commit and updated my comments to reflect this change.

PR #1088, opened 2021-06-04
media-libs/libgfx: add test, dependencies, patches

lucasmitrak:
* Add test, dependencies, and new patches to ebuild
* Add cstring to needed files due to newer version of gcc
* Change png jmpbuf to use the newer safer method due to GLSA 200408-03

Currently, media-libs/libgfx will not compile due to a vulnerability in
libpng which has since been patched [1]. Therefore, the patch
libPNG-1.2.5 updates the code to libPNG's newer and safer method. This
also fixes bug https://bugs.gentoo.org/756061.
However, even with this patch the package will
not compile due to outdated C++, so the patch gcc-4.3 updates two files
to include cstring, which is neccessary for memcpy. These patches will
allow the package to compile. Without these patches, the package will
not compile.
In addition, the ebuild was updated to include these patches, as well as
add the dependencies listed on upstream's homepage [2]. Without these
dependencies, the package will not compile. Finally, a
 src_test() function was added. Due to the package using
outdated FLTK code, some of the tests had to be disabled using a sed
command.

[1] https://security.gentoo.org/glsa/200408-03
[2] http://mgarland.org/software/libgfx.html
Closes: https://bugs.gentoo.org/756061


media-libs/libgfx/libgfx-1.1.0.ebuild
```diff
@@ -10,9 +10,19 @@ SRC_URI="http://mgarland.org/dist/${P}.tar.gz"
 LICENSE="LGPL-2.1"
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
-IUSE="static-libs"
+IUSE="static-libs test"
```
AndrewAmmerlaan, line 13:
```suggestion
IUSE="static-libs test"
RESTRICT="!test? ( test )"
```
To enforce that USE="test" if FEATURES="test"

media-libs/libgfx/libgfx-1.1.0.ebuild
```diff
@@ -10,9 +10,19 @@ SRC_URI="http://mgarland.org/dist/${P}.tar.gz"
...
+DEPEND="
+	virtual/opengl
+	x11-libs/fltk
+"
+
+RDEPEND="${DEPEND}"
+
+PATCHES=(
+	"${FILESDIR}"/${PV}-gcc-4.3.patch
+	"${FILESDIR}"/${PV}-libPNG-1.0.6.patch
```
AndrewAmmerlaan, line 24: This does not match the filename of the patch

AndrewAmmerlaan (changes requested):

media-libs/libgfx/libgfx-1.1.0.ebuild
AndrewAmmerlaan: Actually, since there aren't any test dependencies or other things that are conditional on the test USE flag, you actually don't need to add the flag at all.

media-libs/libgfx/libgfx-1.1.0.ebuild
lucasmitrak: Thank you for catching that, what a silly mistake. One of my commits fixes that, and I tested it to make sure both patches are applied.

media-libs/libgfx/libgfx-1.1.0.ebuild
lucasmitrak: > 
> 
> Actually, since there aren't any test dependencies or other things that are conditional on the test USE flag, you actually don't need to add the flag at all.

I do not understand this comment. I removed test from IUSE in the second commit; however, now the tests are not performed. Is this desirable behavior? If so, that is what my second commit does.

AndrewAmmerlaan: There is USE="test" and FEATURES="test", and these do different things (the identical naming is slightly confusing). FEATURES="test" enables running the test phase when emerging an ebuild, so if I do `FEATURES="test" emerge foobar` emerge will execute the test phase of foobar after the compile phase and before the install phase.
USE="test" is a helper to pull in extra test dependencies or enable the building of tests if required. So my ebuild foobar might have:
```
DEPEND="test? ( some/test-dep )"

src_configure() {
       econf $(use_enable test)
}
```
if foobar happens to have a build system which optionally accepts `--enable-test` to build the test suite. In this case I have used the flag "test", so I must define `test` in IUSE. If I need a test flag, I should also add `RESTRICT="!test? ( test )"` (and this is where the identical naming gets confusing, the first "test" is the USE flag, the second is the FEATURES), this will tell emerge to only run the test phase if both USE="test" _and_ FEATURES="test".

Now in this case, there don't seem to be any test dependencies or anything else which is conditional on `test? ()`, so we don't need IUSE="test" or RESTRICT="!test? ( test )". Emerging the ebuild with FEATURES="test" _should_ enable running the test phase without the need to set any flag. The castxml can probably also lose the test flag, since it doesn't seem to be doing anything there.

AndrewAmmerlaan:
Perfect, Thanks :+1:

media-libs/libgfx/libgfx-1.1.0.ebuild
lucasmitrak: Thank you for this explanation, I did not see it until now.

PR #1090, opened 2021-06-08
 sci-mathematics/gappa: Add multiprocessing.eclass for makeopts_jobs

lucasmitrak:
* Add inherit multiprocessing.eclass for makeopts_jobs

Currently, sci-mathematics/gappa will fail to compile if either -l or
--load-average is set in ${MAKEOPTS} because remake does not accept
them. This commit adds the implements the makeopts_jobs within the
multiprocessing eclass to ensure only --jobs are passed to remake.
This will allow the package to compile and install
correctly even with --load-average in MAKEOPTS.

Closes: https://bugs.gentoo.org/568368

### DIFF
--- a/sci-mathematics/gappa/gappa-1.3.5.ebuild
+++ b/sci-mathematics/gappa/gappa-1.3.5.ebuild
@@ -3,6 +3,8 @@
 
 EAPI=7
 
+inherit multiprocessing
+
 DESCRIPTION="Verifying and proving properties on floating-point or fixed-point arithmetic"
 HOMEPAGE="http://gappa.gforge.inria.fr/"
 SRC_URI="https://gforge.inria.fr/frs/download.php/file/38044/${P}.tar.gz"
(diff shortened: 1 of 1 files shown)


waebbl:
Thanks for picking up this bug, I didn't have it on my focus. Have you looked at `multiprocessing.eclass`? IMO this eclass wasn't available back in 2015 when I opened the bug, or at least didn't have the functionality it has now. Think it could be easier using the `makeopts_jobs` function from this eclass nowadays than taking your approach.

lucasmitrak:
> 
> 
> Thanks for picking up this bug, I didn't have it on my focus anymore. Have you looked at `multiprocessing.eclass`? IMO this eclass wasn't available back in 2015 when I opened the bug, or at least didn't have the functionality it has now. Think it could be easier using the `makeopts_jobs` function from this eclass nowadays than taking your approach.

Thank you for your review and suggestion. I implemented the `multiprocessing.eclass` in my last commit.

AndrewAmmerlaan:
Thanks, :+1: 

I ran into a compilation issue with gcc-11, so I bumped to the latest version (1.4.0) in a follow-up commit that fixed that issue.

PR #1091, opened 2021-06-10
sci-biology/dcm2niix: add patch

lucasmitrak:
* Disable finding git in cmake
* Add patch to ebuild

Currently, sci-biology/dcm2niix will not compile if git is not found.
The patch dcm2niix-disable_find_git.patch comments out the lines which
attempt to find git and require it to be found. With this patch, the
package will compile correctly. Without this patch, cmake will throw
an error if git is not found.

Closes: https://bugs.gentoo.org/755746


AndrewAmmerlaan:
Thanks :+1:

PR #1092, opened 2021-06-11
sci-biology/dcm2niix: add USE flags, dependencies, src_configure

lucasmitrak:
* Add static system-jpeg +jpeg-ls jpeg2k use flags and implement them

Currently, sci-biology/dcm2niix does not take advantage of many
of the package's potential features. These potential features include
a static runtime, using the system-wide media-libs/libjpeg-turbo
instead of bundled compact NanoJPEG decoder, enabling JPEG-LS using
the bundled CharLS library, and enabling JPEG2000 support. These are
important, especially since the static runtime is enabled by default
and JPEG-LS can be enabled without any downside. This commit adds
these USE flags and their implementation in the dependencies and in
src_configure. The naming of the use flags follow the same naming
convention found elsewhere. For example, system-jpeg is the same
as thunderbird, firefox, and seamonkey. In addition, jpeg-ls is the
only possible name for the JPEG-LS transfer syntaxes. These two
package-specific USE flag names were added to metadata.xml.
With this commit, the use flags will be added to the package.
Without this commit, the package will not have these potential
features.

### DIFF
--- a/sci-biology/dcm2niix/dcm2niix-1.0.20201102.ebuild
+++ b/sci-biology/dcm2niix/dcm2niix-1.0.20201102.ebuild
@@ -12,13 +12,29 @@ SRC_URI="https://github.com/rordenlab/dcm2niix/archive/v${PV}.tar.gz -> ${P}.tar
 SLOT="0"
 LICENSE="BSD"
 KEYWORDS="~amd64 ~x86"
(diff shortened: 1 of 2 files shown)


AndrewAmmerlaan:
Excellent :+1:

Issue #1097, opened 2021-07-16
[Feature Request] CadQuery, the great CAD GUI destroyer


leycec:
This feature request is actually just me tracking the multiple PRs I'll be submitting shortly for [CadQuery 2.1](https://cadquery.readthedocs.io), the most recent stable release of a fairly revolutionary open-source Python framework.

CadQuery *kinda* single-handedly obsoletes proprietary CAD GUIs that literally cost millions in licensing fees like AutoCAD and SolidWorks, similar to how NumPy *kinda* single-handedly obsoleted proprietary math GUIs with similar costs like MATLAB. The killer idea here is that CadQuery programmatically generates Computer-Aided Design (CAD) files in obscure industry formats like STEP, DXF, STL, and OBG from Python one-liners like this:

```
import cadquery as cq
cq.Workplane("XY").box(length,height,thickness).faces(">Z").workplane().hole(bearing_diam)
        .faces(">Z").workplane()
        .rect(length-padding,height-padding,forConstruction=True)
        .vertices().cboreHole(2.4, 4.4, 2.1)
```

...which generates [this parametric bearing pillow block](https://cadquery.readthedocs.io/en/latest/quickstart.html):

![parametric bearing pillow block, yo](https://cadquery.readthedocs.io/en/latest/_images/000.png)

*Yeah.* That's hot.

Sadly, CadQuery's build system isn't. Thus this issue. For those familiar with PySide2 (i.e., the official Python bindings for Qt 5), CadQuery is fiendishly similar. And I should know; I authored the PySide2 family of ebuilds currently in Portage. <sup>*good times*</sup>

Just like PySide2, CadQuery implements its own C++-to-Python bindings generator called [pywrap](https://github.com/CadQuery/pywrap) for automagically generating Python bindings to humongously large C++ libraries – in this case, [Open Cascade, a C++ CAD library that's **1.5GB large**](https://en.wikipedia.org/wiki/Open_Cascade_Technology). Just like PySide2, CadQuery is split up into multiple components, each with multiple dependencies collectively taking over half-a-day to build on my ailing 2000-era AMD Athlon II X2 240. <sup>*good times*</sup>

So, here we go. These are the three main CadQuery pull requests I'll be submitting shortly (in order):

1. **[pywrap](https://github.com/CadQuery/pywrap),** CadQuery's aforementoined CLang-based binding generators. It's... pretty intense.
[...]

PR #1099, opened 2021-07-20
CadQuery phase 2: OCP

leycec:
This PR continues the ancient viking saga of CadQuery first transcribed in #1097. We boldly package the next major dependency: [**OCP,** Python bindings to the entirety of Open Cascade auto-generated by pywrap](https://github.com/CadQuery/OCP). This PR is *much* smaller and saner than phase 1, mostly because the only things OCP needs are **(A)** Open Cascade and **(B)** pywrap, which are all packaged. We now give thanks.

Since Gentoo slots Open Cascade into the prior 7.4.0 line and ongoing 7.5.x line, this PR follows suite by providing two suspiciously similar OCP ebuilds:

* `cadquery-ocp-7.4.0.ebuild`, the most recent stable release of OCP targetting Open Cascade 7.4.0. Shockingly, it works. I know, right? :open_mouth: 
* `cadquery-ocp-7.5.2_rc20210706.ebuild`, the most recent Git commit of OCP targetting Open Cascade 7.5.2. Shockingly, it also works. What is even going on here?

These two ebuilds are *mostly* duplicates of one other. Minor differences between the two all stem from minor differences between Open Cascade 7.4.0 and 7.5.2. Basically, Open Cascade 7.5.2 supports JSON and is packaged by Gentoo into saner system directories (e.g., `/usr/include/opencascade-7.5.2/` rather than `/usr/lib64/opencascade-7.4.0/ros/include/`). That's it. We now give thanks. 

Everything mostly makes sense on the packaging front... subject to these familiar caveats:

* *No tests.* Everyone saw that one coming. We couldn't do anything about this with pywrap, but we *can* do something about this with OCP. All OCP does is provide a single Python C extension! So, we just test that we can import a few well-known Open Cascade attributes from that extension. This increases the likelihood that OCP actually did what it said, which is better than the usual NothingBurger™ we've come to expect from CadQuery packaging. I'll take it.
* *Hard-coded paths.* Again, `sed` solves all these problems and more. Thanks, Stallman, for you continue to grace us with modern solutions to age-old problems.
* *Manual `cmake` invocation.* The `cmake` eclass doesn't play nicely with the `distutils-r1` eclass, so I ditched it. Instead, we invoke `cmake` manually. Unfortunately, <sup>...*sigh*</sup> the build process by which OCP generates bindings isn't quite as automated as one would hope – by which I mean it's totally unautomated. I had to manually reverse engineer this process from their Anaconda-specific `build-bindings-job.yml` file. To compound matters, this process needs header include directories for Clang, VTK, and Open Cascade – but there isn't an automated way to query for those directories. Neither Clang, VTK, nor Open Cascade install `.pc` files for inspection by `pkg-config`. Although Clang, VTK, and Open Cascade *do* all install `.cmake` files for inspection by `cmake`, OCP ignores those files and instead ships its own broken `FindOpenCascade.cmake` file **that completely fails at its only job.** So, we **(A)** manually find header include directories and **(B)** patch those directories into `FindOpenCascade.cmake`. It works, but it sucks. I'll be filing an upstream mega-issue detailing all of this suckage once we finish packaging CadQuery in full.

Upstream will pay for their crimes! Until they do, OCP is actually really amazing – ignoring the packaging crisis detailed above. OCP not only provides one-to-one Python bindings for Open Cascade but, when combined with CadQuery, a *much* higher-level and easier-to-use API for converting curmudgeonly CAD formats birthed in the seventh circle of Hell like STEP, IGES, and DXF into reasonable VTK `PolyData` objects.

Honestly, I consider STEP, IGES, and DXF unusable without OCP. All hail CadQuery for it makes the insane things sane again.


dev-python/cadquery-ocp/cadquery-ocp-7.5.2_rc20210706.ebuild
```diff
@@ -0,0 +1,160 @@
...
+		cmake \
+			-Wno-dev \
+			-S OCP/ \
+			-B OCP.binary_tree/ \
+			-G Ninja \
+			-D CMAKE_BUILD_TYPE=Gentoo \
+			-D PYTHON_EXECUTABLE="${PYTHON}" \
+			|| die
+	}
+	python_foreach_impl cadquery-ocp_src_configure
```
AndrewAmmerlaan, line 133:
```suggestion
	python_foreach_impl CMAKE_USE_DIR="${BUILD_DIR}/OCP" cmake_src_configure
```
(and maybe something with `my_cmake_args`)

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_rc20210706.ebuild
```diff
@@ -0,0 +1,160 @@
...
+}
+
+src_compile() {
+	cadquery-ocp_src_compile() {
+		cd "${BUILD_DIR}" || die
+		cmake --build OCP.binary_tree/ -- -j $(get_nproc) || die
+	}
+	python_foreach_impl cadquery-ocp_src_compile
```
AndrewAmmerlaan, line 141:
```suggestion
	python_foreach_impl CMAKE_USE_DIR="${BUILD_DIR}/OCP" cmake_src_compile
```

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_rc20210706.ebuild
```diff
@@ -0,0 +1,160 @@
...
+
+	# "dev-python/clang-python" atom targeting this Clang version.
+	local _CLANG_PYTHON_ATOM="dev-python/clang-python-${_CLANG_VERSION}"
+
+	# Ensure "dev-python/clang-python" targets this Clang version.
+	has_version -r "=${_CLANG_PYTHON_ATOM}" ||
+		die "${_CLANG_PYTHON_ATOM} not installed."
+
+	# Remove all vendored subdirectories.
+	rm -rf conda opencascade pywrap || die
```
AndrewAmmerlaan, line 82: AFAIK `-f` makes `rm` exit successfully even if the dir does not exist, so now the `|| die` is useless. I think we should just get rid of the `-f`

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_rc20210706.ebuild
```diff
@@ -0,0 +1,160 @@
...
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{8..9} )
+
+# The "cmake" eclass is intentionally omitted here despite cmake being run.
+# That eclass expects a "${S}/CMakeLists.txt" file during src_prepare(),
+# whereas this package uniquely generates one "${BUILD_DIR}/OCP/CMakeLists.txt"
+# file for each active Python version.
+inherit llvm multiprocessing python-r1
```
AndrewAmmerlaan, line 12:
```suggestion
inherit cmake llvm multiprocessing python-r1
```

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_rc20210706.ebuild
```diff
@@ -0,0 +1,160 @@
...
+	python_foreach_impl cadquery-ocp_src_test
+}
+
+src_install() {
+	python_moduleinto .
+	cadquery-ocp_src_install() {
+		cd "${BUILD_DIR}" || die
+		python_domodule OCP.binary_tree/OCP*.so
+	}
+	python_foreach_impl cadquery-ocp_src_install
```
AndrewAmmerlaan, line 159:
```suggestion
	python_foreach_impl CMAKE_USE_DIR="${BUILD_DIR}/OCP" cmake_src_install
```

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_rc20210706.ebuild
```diff
@@ -0,0 +1,160 @@
...
+
+# Ensure the path returned by get_llvm_prefix() contains clang.
+llvm_check_deps() {
+	has_version -r "sys-devel/clang:${LLVM_SLOT}"
+}
+
+# OCP currently requires manual configuration, compilation, and installation
+# inspired by the conda-specific "build-bindings-job.yml" file.
+src_prepare() {
+	default
```
AndrewAmmerlaan, line 43:
```suggestion
```
(Otherwise we apply the patches twice, cause `cmake_src_prepare` also does this)

AndrewAmmerlaan:
>     * _Manual `cmake` invocation._ The `cmake` eclass doesn't play nicely with the `distutils-r1` eclass, so I ditched it. Instead, we invoke `cmake` manually. Unfortunately, ..._sigh_ the build process by which OCP generates bindings isn't quite as automated as one would hope – by which I mean it's totally unautomated. I had to manually reverse engineer this process from their Anaconda-specific `build-bindings-job.yml` file. To compound matters, this process needs header include directories for Clang, VTK, and Open Cascade – but there isn't an automated way to query for those directories. Neither Clang, VTK, nor Open Cascade install `.pc` files for inspection by `pkg-config`. Although Clang, VTK, and Open Cascade _do_ all install `.cmake` files for inspection by `cmake`, OCP ignores those files and instead ships its own broken `FindOpenCascade.cmake` file **that completely fails at its only job.** So, we **(A)** manually find header include directories and **(B)** patch those directories into `FindOpenCascade.cmake`. It works, but it sucks. I'll be filing an upstream mega-issue detailing all of this suckage once we finish packaging CadQuery in full.

I think we might be able to workaround the cmake mess with `BUILD_DIR/CMAKE_IN_SOURCE_BUILD/CMAKE_USE_DIR`, if it doesn't work then we'll just do it like it is now.

waebbl:
> Since Gentoo slots Open Cascade into the prior 7.4.0 line and ongoing 7.5.x line, this PR follows suite by providing two suspiciously similar OCP ebuilds:
> 
>     * `cadquery-ocp-7.4.0.ebuild`, the most recent stable release of OCP targetting Open Cascade 7.4.0. Shockingly, it works. I know, right? open_mouth
> 
>     * `cadquery-ocp-7.5.2_rc20210706.ebuild`, the most recent Git commit of OCP targetting Open Cascade 7.5.2. Shockingly, it also works. What is even going on here?
> 
As maintainer of the opencascade package, I'm planning on removing the 7.4.0 line not very far into the future. AFAIK pkgcheck and repoman can't handle cross-repo dependencies, so a removal would silently break the `cadquery-ocp-7.4.0.ebuild`. I'd suggest, we stick to the v7.5.2 release and just ignore the v7.4.0 release altogether.

>     * _Manual `cmake` invocation._ The `cmake` eclass doesn't play nicely with the `distutils-r1` eclass, so I ditched it. Instead, we invoke `cmake` manually. Unfortunately, ..._sigh_ the build process by which OCP generates bindings isn't quite as automated as one would hope – by which I mean it's totally unautomated. I had to manually reverse engineer this process from their Anaconda-specific `build-bindings-job.yml` file. To compound matters, this process needs header include directories for Clang, VTK, and Open Cascade – but there isn't an automated way to query for those directories. Neither Clang, VTK, nor Open Cascade install `.pc` files for inspection by `pkg-config`. Although Clang, VTK, and Open Cascade _do_ all install `.cmake` files for inspection by `cmake`, OCP ignores those files and instead ships its own broken `FindOpenCascade.cmake` file **that completely fails at its only job.** So, we **(A)** manually find header include directories and **(B)** patch those directories into `FindOpenCascade.cmake`. It works, but it sucks. I'll be filing an upstream mega-issue detailing all of this suckage once we finish packaging CadQuery in full.

I haven't looked at your code yet, but all three packages, clang, vtk and occt ship with the new config style `.cmake` files. From my POV there's no need to manually find header include dirs. We can instead use the `get_target_property` cmake function to ask cmake for where the header files are located. Imagine a construct like
```
find_package(vtk 9 CONFIG REQUIRED)
if(TARGET VTK::VTK)
  get_target_property(VTK_INCLUDE_DIRS VTK::VTK INTERFACE_INCLUDE_DIRECTORIES)
endif()
```
Replace `vtk` with the targets we actually need. It could (and in theory should) be possible to just delete (or ignore) the old style `FindOpenCascade.cmake` module file.

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_rc20210706.ebuild
```diff
@@ -0,0 +1,160 @@
...
+			-D PYTHON_EXECUTABLE="${PYTHON}" \
+			|| die
+	}
+	python_foreach_impl cadquery-ocp_src_configure
+}
+
+src_compile() {
+	cadquery-ocp_src_compile() {
+		cd "${BUILD_DIR}" || die
+		cmake --build OCP.binary_tree/ -- -j $(get_nproc) || die
```
waebbl, line 139: Shouldn't we honour the users choice by using `makeopts_jobs` instead of `get_nproc`?

AndrewAmmerlaan: Somewhat related to this, I have a feeling that major parts of the emerge are not parallelized at all:
```
* dev-python/cadquery-ocp-7.5.2_rc20210706

current merge time: 5 hours, 7 minutes and 58 seconds.
ETA: any time now.
```
And it's still busy.....

The System Monitor shows me that it is clearly using only one core at the moment (I haven't been monitoring it all the time, so maybe it was multithreaded before)

If we can't speed this up by parallelizing it (more), then maybe we should use `python-single-r1` instead of `python-r1`. Because if this really takes this long than no one is going to want to install this for multiple PYTHON_TARGETS since that doubles or triples the emerge time. (This also has the additional benefit of simplifying the cmake problem, since we don't have to deal with multiple BUILDDIRs)

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_rc20210706.ebuild
```diff
@@ -0,0 +1,160 @@
...
+# whereas this package uniquely generates one "${BUILD_DIR}/OCP/CMakeLists.txt"
+# file for each active Python version.
+inherit llvm multiprocessing python-r1
+
+MY_COMMIT="9a6210e3c490e741719ccdada900c70e58b09c91"
+OCCT_PV=$(ver_cut 1-3)
+
+DESCRIPTION="Python wrapper for OCCT generated using pywrap"
+HOMEPAGE="https://github.com/CadQuery/OCP"
+SRC_URI="https://github.com/CadQuery/OCP/archive/${MY_COMMIT}.tar.gz -> ${P}.tar.gz"
```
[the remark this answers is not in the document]
leycec: Oh, great catch! Somehow, I failed to notice that. Somehow, you never fail to notice everything. You're like the PR Eye of Sauron. :eye:

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_rc20210706.ebuild
leycec: Why are you always right, Andrew? Why? WHY!?!?

Embarrassingly, that was muscle memory. I always pass `-rf` together, so that's what the fingers do by themselves now. I never even knew that you *could* pass `-r` without `-f`. Of course, you can. And here we are, washed up on the lonely shore of another misbegotten code fragment.

leycec:
> As maintainer of the opencascade package...

Thank you for the blessed work you have done.

> I'm planning on removing the 7.4.0 line not very far into the future.

Excellent! This simplifies my life, for which I am also grateful. I'll drop the 7.4.0 ebuild exactly as you say. :+1: 

> I haven't looked at your code yet...

Let us just safely assume that it's ugly, fragile, and dangerous.

> We can instead use the `get_target_property` cmake function to ask cmake for where the header files are located. Imagine a construct like...

Yes, I can see it now – a working `CMakeLists.txt` file, the true Shangri-La of every Gentooer who packages risky bleeding-edge software. Your possibly working code snippet is *brilliant,* because my `cmake` skills are on par with my COBOL skills. Now I just need to wait several hours for `bindgen` to regenerate its non-working `CMakeLists.txt` file, because I foolishly failed to save a copy of the build tree after `src_prepare()`. :drooling_face:

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_rc20210706.ebuild
leycec: > Shouldn't we honour the users choice by using `makeopts_jobs` instead of `get_nproc`?

...yes. Yes, we should. I actually didn't know that `makeopts_jobs()` even existed, because I never actually read the eclass documentation. Let's pretend I didn't just admit that.

leycec: > I have a feeling that major parts of the emerge are not parallelized at all...

That's... not good. I thought it was just me. I'm relieved to learn that my geriatric CPU isn't entirely too blame, but disappointed to learn that OCP may have no idea what it's doing. In theory:

* The `src_prepare()` phase *should* be parallelized, because we explicitly pass `--njobs $(makeopts_jobs)`. This assumes Cadquery pywrap works as expected, which may be too much to ask.
* The `src_compile()` phase *should* be parallelized too, both because we explicitly pass `-j $(makeopts_jobs)` to `ninja` *and* because `ninja` defaults to parallelized compilation dependent on core count + 1.
* The `src_configure()` phase doesn't appear to be parallelized but probably should be.

Let's see if I can track down exactly what isn't happening and why. `<cracks_arthritic_knuckles>`

leycec:
@AndrewAmmerlaan: Your proposed `cmake` rework blows my feeble mind. I'd *much* rather hand all of this off to the `cmake` eclass, because I don't trust myself enough to get `cmake` interactions right.

Under optimistic utopian conditions, this should take a day or two to rework. And then there's reality.

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_rc20210706.ebuild
leycec: I can confirm that `src_prepare()` is parallelized. So, that's nice.

I suspect we're just going to discover that generating Python bindings to a 1.5GB C++ API unavoidably consumes a *crazy* amount of wallclock time. I'm kinda okay with that. This is the nature of the beast. Of course, half-a-day build times makes this more a sloth than a beast.

Still, I'd kinda be nice if we could preserve multiple `PYTHON_TARGETS`. Gentoo is choice, even if that means a bad choice. More relevantly, CadQuery will probably end up becoming a mandatory runtime dependency of various other downstream Python projects in a few years. Dropping multiple targets for CadQuery means dropping multiple targets for those future packages, too.

Users on painfully outdated systems that wheeze uncontrollably (like my own) can always disable multiple targets with USE flags like:

```
dev-python/cadquery-ocp -python_targets_python3_8
```

Maybe we could emit a warning message in `pkg_setup()` when multiple targets are enabled?

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_rc20210706.ebuild
AndrewAmmerlaan: I only noticed this because I made the exact same mistake about a month ago :smile: . On full-auto I did `-rf` and someone complained to me that `--force` is wrong in ebuilds (which in hind-sight makes sense, it is just that the `-f` is basically glued to the `-r` in my head :P )

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_rc20210706.ebuild
AndrewAmmerlaan: > Maybe we could emit a warning message in `pkg_setup()` when multiple targets are enabled?

That would work too, it might even be better because then the user is warned about what they are getting into when compiling this.

I haven't checked the memory/disk usage, but it is probably large too. So we might want to set `CHECKREQS_DISK_BUILD` from `check-reqs.eclass`. And we should probably set it conditionally to how many `PYTHON_TARGETS` are enabled, something like:
```
pre_build_checks() {
     local M=0
     for x in ${PYTHON_COMPAT}; do
         M=$(( <some value for each target in Mb> + ${M} ))
    done
    CHECKREQS_DISK_BUILD=${M}M check-reqs_pkg_${EBUILD_PHASE}
}

pkg_pretend() {
	pre_build_checks
}

pkg_setup() {
	pre_build_checks
}
```
(untested, and stolen from rust.ebuild)

leycec: I approve of ~~stealing~~ being inspired by everything not nailed down from Rust.

My crude and probably wrong disk usage heuristic is showing ~400MB disk consumption and negligible memory consumption (...maybe because parsing is much less memory intensive than linking?) per Python target. Your hunch was spot on; the disk consumption is severe enough that we really do want to enforce pre-build constraints. Let's do this!

waebbl:
> **EDIT:** Ugh! Nothing I wrote below applies. I conveniently forgot to mention that OCP doesn't actually have a top-level `CMakeLists.txt` file. Moreover, the sub-level `OCP/CMakeLists.txt` file it _does_ have is only dynamically generated by manually running this command in our ebuild:
> 
> ```
> 		${EPYTHON} -m bindgen \
> 			--verbose \
> 			--njobs $(makeopts_jobs) \
> 			--libclang "${_CLANG_LIB_FILE}" \
> 			--include "${_CLANG_INCLUDE_DIR}" \
> 			--include "${_VTK_INCLUDE_DIR}" \
> 			all ocp.toml || die
> ```
> 
> Since that command runs both outside of _and_ before any CMake context, there's no relevant `CMakeLists.txt` file to patch your clever `get_target_property` CMake calls into. Upstream really needs to solve that by adding a top-level `CMakeLists.txt` file governing their full build process (including the initial `bindgen` command above). I'm confident they'll botch it, but that's okay – because at least then there'd be something for us to fix.

I never had to work with toml files before, so excuse me, if my suggestions are not usable.
- I still think, the `FindOpenCASCADE.cmake` module might not be necessary. They seem to restrict the list of modules to a selection of the available modules, which, at least in theory, if they want to create bindings for the complete package is possibly only a safeguard mechanism until everything works. IMO it won't hurt, if all available modules are scanned and included by cmake, just some additional, not yet needed definitions. So we can eventually strip the file from the `additional_files` entry in ocp.toml and influence the creation of the CMakeLists.txt file.
- Is the file `templates/CMakeLists.j2` used when creating the CMakeLists.txt files? If so, we could patch this file to control the creation of the cmake file(s)
```
find_package(OpenCASCADE CONFIG REQUIRED)
find_package(VTK 9.0 CONFIG REQUIRED COMPONENTS ...)
```
Note the removal of the COMPONENTS option for occt. There's no component named OPENCASCADE defined in the config file, which might give an error if it's defined. If no components are specified, the occt config file looks for all modules which have been installed. We might also need to check whether the components passed for vtk are correct, but IIRC they are.

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_rc20210706.ebuild
```diff
@@ -0,0 +1,160 @@
...
+
+LICENSE="Apache-2.0"
+KEYWORDS="~amd64 ~x86"
+SLOT="0"
+
+# This version requirement derives from "${BUILD_DIR}/OCP/CMakeLists.txt".
+BDEPEND=">=dev-util/cmake-3.16"
+RDEPEND="~sci-libs/opencascade-${OCCT_PV}[json,tbb,vtk]"
```
waebbl, line 27: We need to take care here, vtk is being installed with python support enabled as well. The occt ebuild only requires vtk build with `USE=rendering` support enabled. AFAICS the build system requires the `WrappingPythonCore` vtk component being available.
This **will** restrict us to using python-single eclass, as vtk only supports a single python target. I think, it's non-trivial to add multi-python support to vtk, if it's possible at all.

AndrewAmmerlaan:
> **EDIT:** Ugh! Nothing I wrote below applies. I conveniently forgot to mention that OCP doesn't actually have a top-level `CMakeLists.txt` file. Moreover, the sub-level `OCP/CMakeLists.txt` file it _does_ have is only dynamically generated by manually running this command in our ebuild:
> 
> ```
> 		${EPYTHON} -m bindgen \
> 			--verbose \
> 			--njobs $(makeopts_jobs) \
> 			--libclang "${_CLANG_LIB_FILE}" \
> 			--include "${_CLANG_INCLUDE_DIR}" \
> 			--include "${_VTK_INCLUDE_DIR}" \
> 			all ocp.toml || die
> ```
> 
> Since that command runs both outside of _and_ before any CMake context, there's no relevant `CMakeLists.txt` file to patch your clever `get_target_property` CMake calls into. Upstream really needs to solve that by adding a top-level `CMakeLists.txt` file governing their full build process (including the initial `bindgen` command above). I'm confident they'll botch it, but that's okay – because at least then there'd be something for us to fix.

If you call `CMAKE_USE_DIR="${BUILD_DIR}/OCP" cmake_src_prepare` *after* the `bindgen` stuff then the patches are only applied after bindgen has generated the `CMakeLists.txt` file, or am I missing something? (since `cmake_src_prepare` calls`eapply_user` and friends)

leycec:
That feeling when you double facepalm yourself in the privacy of your own man-cave.

`sci-libs/vtk` is `PYTHON_SINGLE_TARGET`. Both OCP and CadQuery require `sci-libs/vtk` at runtime, so OCP and CadQuery *must* also be `PYTHON_SINGLE_TARGET`. At least it's summer in Canada. We still have this much to cling to.

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_rc20210706.ebuild
AndrewAmmerlaan: I just discovered this little helper function that might come in handy here:
```
ebegin

Like einfo, we output a helpful message and then hint that the following operation may take some time to complete. Once the task is finished, you need to call eend. 
```

leycec:
> If you call `CMAKE_USE_DIR="${BUILD_DIR}/OCP" cmake_src_prepare` after the `bindgen` stuff then the patches are only applied after `bindgen` has generated the `CMakeLists.txt` file, or am I missing something?

**Yupper!** You are right, because you are @AndrewAmmerlaan.

Sadly, patching `CMakeLists.txt` with @waebbl's clever cmake logic still wouldn't help. We still have to manually discover these paths before running `bindgen` (which generates the `CMakeLists.txt` file to be patched), because we have to pass these paths *to* `bindgen` before doing anything else:

```
	${EPYTHON} -m bindgen \
		--verbose \
		--njobs $(makeopts_jobs) \
		--libclang "${_CLANG_LIB_FILE}" \
		--include "${_CLANG_INCLUDE_DIR}" \
		--include "${_VTK_INCLUDE_DIR}" \
		all ocp.toml || die
```

This is why OCP needs a top-level `CMakeLists.txt` file governing the full build process, but that's a bit out-of-scope for distro-specific packaging. The cost of bleeding-edge software is high indeed.

leycec:
**Okay!** We're now cooking with electrons.

Andrew's exhaustive code review has yielded total success. This is no surprise. The only surprise is that I did what Andrew said this weekend rather than just binge-watch Ted Lasso. In short, I've now:

* **Dropped the 7.4.0 line** in accordance with @waebbl's helpful reminder that Open Cascade 7.4.0 is on life support and to be ~~axe-murdered~~ retired shortly.
* **Inherited the 7.5.2 line from the `cmake` eclass** in lieu of manually invoking `cmake`. Sprinkling a liberal dose of `CMAKE_USE_DIR="${S}/OCP" cmake_src_{phase}` black magic everywhere did, indeed, work like a charm. This has the extreme advantage of enabling Gentoo's cmake integration, including full `CFLAGS`, `CXXFLAGS`, and `LDFLAGS` support as well as ANSI colors. I am stunned.
* **Dropped multiple Python target support** in lieu of single Python target support, streamlining everything across the ebuild. OCP now requires the same single Python target that VTK requires. Sanity has been restored.
* **Increased disk usage requirements to 1.3GB,** after extensive profiling before, during, and after every ebuild phase. Specifically:
  * CLang-based preparation consumes an initial 400MB.
  * Ninja-based compilation adds an eye-opening 900MB.

We're almost done! But we're not *quite* there. We still need to regenerate Open Cascade symbols with [LIEF (Library for Instrumenting Executable Formats)](https://github.com/lief-project/LIEF/blob/master/cmake/LIEFOptions.cmake) in the `src_prepare()` phase *before* running `bindgen` against those symbols. I have locally working ebuilds for LIEF and its dependencies, but haven't quite finished integrating that work with OCP.

In any case, let's push this before a freak thunderstorm demagnetizes the last good sectors on the ten year-old mechanical HDD hosting my most sensitive data. :cloud_with_lightning_and_rain: :floppy_disk:

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_rc20210706.ebuild
leycec: **Yup, yup, and more yup.** I just independently discovered this yesterday after adding an explicit dependency on VTK >= 9.0.0. Of course, everything then broke.

OCP now requires the same single Python target that VTK requires, which dramatically simplifies my life, which means this is the right thing to do.

> I think, it's non-trivial to add multi-python support to vtk, if it's possible at all.

Yeah... I kinda briefly contemplated that, then realized I needed to package [LIEF (Library for Instrumenting Executable Formats)](https://github.com/lief-project/LIEF) for this, then realized I was too tired to do anything more. A tired man is tired.

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_beta.ebuild
```diff
@@ -0,0 +1,172 @@
...
+}
+
+# OCP currently ships no test suite, so we synthesize a crude import unit test.
+src_test() {
+	PYTHONPATH="${S}/OCP.build" ${EPYTHON} -c \
+		'from OCP.gp import gp_Vec, gp_Ax1, gp_Ax3, gp_Pnt, gp_Dir, gp_Trsf, gp_GTrsf, gp, gp_XYZ'
+}
+
+src_install() {
+	die
```
AndrewAmmerlaan, line 169: I removed this `die` while merging

AndrewAmmerlaan:
Awesome, thanks! 

There was one left over `die` hanging around in `src_install` which I removed while merging.

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_beta.ebuild
```diff
@@ -0,0 +1,172 @@
...
+SLOT="0"
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
+
+# CMake and VTK requirements derive from the "OCP/CMakeLists.txt" file
+# generated by the src_prepare() phase.
+BDEPEND=">=dev-util/cmake-3.16"
+RDEPEND="
+	${PYTHON_DEPS}
+	~sci-libs/opencascade-${OCCT_PV}[json,tbb,vtk]
+	>=sci-libs/vtk-9.0.0[${PYTHON_SINGLE_USEDEP}]
```
waebbl, line 31: Sorry for being late to the party. Didn't had the time yesterday to review this further. We need to add `[python]` USE flag here as well, i.e. `vtk-9.0.0[python,${PYTHON_SINGLE_USEDEP}]`.

AndrewAmmerlaan: Added, Thanks :+1:

leycec:
**Bam.** Thanks so much for the insightful code reviews and cleanup! I *really* appreciate how helpful, constructive, and fun everyone is here. Another CadQuery PR bites the `/dev/null`.

Up next: "CadQuery phase 2.5: OCP + LIEF," where we improve the OCP ebuild to dump local OCP symbols with LIEF before generating OCP bindings for safety. *The tension is palpable.*

PR #1100, opened 2021-07-26
media-libs/libgfx: version bump to 1.1.1 which is hosted on github

lucasmitrak:
* Version bump to 1.1.1 which is hosted on github on its own branch

Currently, the science repo only has media-libs/libgfx-1.1.0. Upstream
has made available version 1.1.1; however, it is not located in 1.1.0's
SRC_URI [1]. Instead, he made version 1.1.1 available on github in its
own branch [2]. This new version implements both patches stored in
the ebuild's files directory[3]. These patches were made available
upstream by me, Lucas Mitrak, via email.

[1] http://mgarland.org/files/dist/
[2] https://github.com/mjgarland/libgfx/tree/branch-1.1.1
[3] https://github.com/gentoo/sci/tree/master/media-libs/libgfx/files


media-libs/libgfx/libgfx-1.1.1.ebuild
```diff
@@ -0,0 +1,45 @@
...
+
+EAPI=7
+
+DESCRIPTION="Simplify the creation of computer graphics software"
+HOMEPAGE="http://mgarland.org/software/libgfx.html"
+
+if [[ ${PV} == "1.1.1" ]] ; then
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/mjgarland/libgfx"
+	EGIT_BRANCH="branch-1.1.1"
```
AndrewAmmerlaan, line 12: The difficulty with `git-r3.eclass` is that it skips the regular fetching logic, and therefore skips the Manifest file. Because it skips the Manifest check we can't have KEYWORDS in it for QA reasons (and pkgcheck will complain about it).

However, we can use GitHub's ability of creating tarballs for any commit as a workaround:
```
COMMIT="f28ca5f0df3044f6399d6214faf66daa36712b91"
SRC_URI="https://github.com/mjgarland/libgfx/archive/${COMMIT}.tar.gz -> ${P}.tar.gz"
S="${WORKDIR}/${PN}-${COMMIT}"
```
Because now we don't use `git-r3.eclass` and have the tarball in plain old `SRC_URI` instead we can have KEYWORDS.

media-libs/libgfx/libgfx-1.1.1.ebuild
```diff
@@ -0,0 +1,45 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
AndrewAmmerlaan, line 4: Probably we can already use `EAPI=8` on this ebuild

media-libs/libgfx/libgfx-1.1.1.ebuild
lucasmitrak: Thank you for your review. Should I drop version 1.1.0? When I run `pkgcheck scan .` I get a QA message:
`media-libs/libgfx`
`RedundantVersion: version 1.1.0: slot(0) keywords are overshadowed by version: 1.1.1`

AndrewAmmerlaan: >  Should I drop version 1.1.0? When I run `pkgcheck scan .` I get a QA message:
> `media-libs/libgfx`
> `RedundantVersion: version 1.1.0: slot(0) keywords are overshadowed by version: 1.1.1`

Nah the RedundantVersion check is not really important, we skip it in the GitHub actions because it gets triggered every time there is more then one version with identical keywords.

media-libs/libgfx/libgfx-1.1.1.ebuild
lucasmitrak: I changed the EAPI to 8 in the commit https://github.com/gentoo/sci/pull/1100/commits/d4b4ded1e25e5090e9a612b9db307e16a432b1b9

media-libs/libgfx/libgfx-1.1.1.ebuild
lucasmitrak: I changed the SRC_URI to what you suggested in the commit https://github.com/gentoo/sci/pull/1100/commits/53850a631f8c0ea277b97c6f492c3f913fad68ef

lucasmitrak:
I had to rebase my commits and then force push onto this branch. If I did it incorrectly, let me know.

media-libs/libgfx/libgfx-1.1.0.ebuild
```diff
@@ -1,10 +1,11 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=7
+EAPI=8
```
AndrewAmmerlaan, line 4: If we bump the EAPI of an ebuild we should revision bump it as well to trigger a rebuild for users who have this package installed.

AndrewAmmerlaan:
> I had to rebase my commits and then force push onto this branch. If I did it incorrectly, let me know.

Perfect! 

I merged the added 1.1.1 ebuild, but left out the changes for the 1.1.0 ebuild because if we bump the EAPI then we also have the revbump the ebuild. In any case it is also fine to just leave the 1.1.0 ebuild as is, we will probably remove it some time in the future anyway.

Next time please squash your commits into a single commit, we try to always have a single commit for every unique package. This makes reverting things easier and makes the git history a bit cleaner. The procedure to do this (an interactive git rebase) is not quite trivial if you are unfamiliar with it, but there is a good tutorial here: https://thoughtbot.com/blog/git-interactive-rebase-squash-amend-rewriting-history

PR #1103, opened 2021-08-03
Cadquery phase 2.8: OCP + LIEF

leycec:
This PR in the timeless saga that is CadQuery improves our already peerless OCP ebuild to:

* **Fix CadQuery/OCP#36,** an upstream issue that we can circumvent because Gentoo is hot stuff. For some reason, upstream expects Open Cascade (OCCT) to be compiled with `DUSE_GLES2=OFF`. We trivially guarantee that on our side by disabling the `gles2` USE flag on `sci-libs/opencascade`. On other platforms, that is *not* trivial. There was much griping from FreeBSD-ers, for example. Thankfully, we're not them. Everything is trivial on Gentoo if you just hit it hard enough.
* **Relax OCCT version requirements.** Previously, OCP required an exact OCCT version synchronized with itself – which gets pretty rough to maintain. Whenever you bump OCCT, you have to bump OCP in lockstep. What a pain! After merging this PR, OCP will loosely require any OCCT 7.5.x version. *All the pain suddenly goes away.*

## Symbol Dumps: It Feels Good

You might now be wondering, "How do we accomplish this mysterious miracle, @leycec?" It is simple, I assure you. <sup>*shifty eyes*</sup>

We just:

1. Add a new build-time OCP dependency on [LIEF, the (L)ibrary to (I)nstrument (E)xecutable (F)ormats](https://github.com/lief-project/LIEF).
1. Add a new `dev-libs/lief` package packaging LIEF.
1. Run the `dump_symbols.py` script bundled with OCP before doing anything else. That dumps the symbol tables for OCCT's shared libraries with the help of LIEF's Python bindings, thus synchronizing OCP with the current version of OCCT. This works well but churns for about a minute without printing anything. So, we preface that with a terse `einfo` that *does* print something. y u so quiet, `dump_symbols.py`?

## LIEF: It Feels Bad

What could be simpler?

...okay, I'll be honest here. That "Add a new `dev-libs/lief` package packaging LIEF" step that I glibly pretended was easy up above? Yeah. That was *really* hard. Like, hernia-inducing hard. Like, "I now need a qualified chiropractor to crack my spine in twenty-three places hopefully paid for by the Canadian government" hard.

In the spirit of Microsoft's patented 3E <sup>(E)mbrace, (E)xtend, (E)xtingish</sup> approach to life, I first tried embracing an [existing LIEF ebuild](https://data.gpo.zugaina.org/pentoo/dev-util/lief) authored by Pentoo co-developer @blshkv (Anton Bolshakov). It didn't go well. The ebuild:

* Erroneously depends on Python packages like [xtract](https://pypi.org/project/xtract) that LIEF doesn't actually require.
* Defines this weird `wrap_python()` shim around `distutils-r1` eclass phase functions that doesn't actually seem to do anything useful.
* Worst of all, calls `setup.py`. That's usually a good thing... but this is LIEF, so that's a bad thing. LIEF's `setup.py` is this Bizarro World thin wrapper around `cmake` that fails to do anything useful while doing many things that are *not* useful, like *not* supporting most CMake options required by Gentoo's `cmake` eclass (e.g., `-C`, `-DCMAKE_TOOLCHAIN_FILE=`, `-DCMAKE_BUILD_TYPE=`).

In short, Pentoo's LIEF ebuild is an admirable first try! It's just not quite up to Portage par – and we always aim for [birdies](https://www.scottishgolfhistory.org/origin-of-golf-terms/bogey/#birdie) here. So, I wrote an entirely new LIEF ebuild from the ground up. Rather than dangerously trying to split the work between CMake and `setuptools`, this ebuild entirely ignores LIEF's `setup.py` by instead defering to `CMakeLists.txt` for everything. It works great!

...except that upstream, a project with ~2,500 GitHub stars, **never tested their build process.** [Most of their CMake options break when you disable them](https://github.com/lief-project/LIEF/issues/599), so we don't do that. Moreover, [they entirely forgot to implement the `-DLIEF_INSTALL_PYTHON` CMake option](https://github.com/lief-project/LIEF/issues/599#issuecomment-889624251), so we do that for them.

[I notified upstream of everything](https://github.com/lief-project/LIEF/issues/599). I pray they will eventually spare future generations the ~~burning personal Hell~~ mild unpleasantness I experienced while packaging LIEF.

## All's Well That Ends

Phew! CadQuery phase 2.8 was more of a purgatorial ride than I wanted, but at least it's all over. We march with grim faces onward to **CadQuery phase 3: Yes, it's finally CadQuery.**


dev-libs/lief/lief-0.11.5.ebuild
```diff
@@ -0,0 +1,125 @@
...
+# LIEF tests are non-trivial (if not infeasible) to run in the general case.
+# For example, "tests/CMakeLists.txt" implies all USE flags must be enabled:
+#     if (NOT LIEF_ELF OR NOT LIEF_PE OR NOT LIEF_MACHO)
+#       message(FATAL_ERROR "Tests require all LIEF's modules activated" )
+#     endif()
+RESTRICT="test"
+
+S="${WORKDIR}/LIEF-${PV}"
+
+DISTUTILS_OPTIONAL=1
```
AndrewAmmerlaan, line 47: Since we are not inheriting `distutils-r1` this is not doing anything, right?

dev-libs/lief/lief-0.11.5.ebuild
```diff
@@ -0,0 +1,125 @@
...
+		-DLIEF_MACHO=ON
+		-DLIEF_ART=ON
+		-DLIEF_DEX=ON
+		-DLIEF_OAT=ON
+		-DLIEF_VDEX=ON
+
+		-DBUILD_SHARED_LIBS="$(usex static-libs OFF ON)"
+		-DLIEF_C_API="$(usex c ON OFF)"
+		-DLIEF_EXAMPLES="$(usex examples ON OFF)"
+		-DLIEF_FORCE32="$(usex x86 ON OFF)"
```
AndrewAmmerlaan, line 104: Is this a x86 thing, or is it a multilib thing? Because I suspect the latter, in which case the `cmake-multilib` eclass might come in handy. I haven't ever used it myself, but the description says "provides a glue between cmake.eclass(5) and multilib-minimal.eclass(5)" so it does sound like it might do what we need it to do.

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_beta.ebuild
```diff
@@ -100,13 +110,27 @@ src_prepare() {
...
+	mkdir -p "${_OCCT_DUMP_SYMBOLS_DIR}" || die
+	ln -s "${_OCCT_LIB_DIR}" "${_OCCT_DUMP_SYMBOLS_DIR}"/. || die
+
+	# Update all hardcoded OCCT shared library versions in "dump_symbols.py".
+	sed -i -e 's~\(\.so\.\)[0-9]\+.[0-9]\+.[0-9]\+~\1'${OCCT_PV}'~' \
+		dump_symbols.py || die
+
+	# Dump (i.e., generate) symbol tables for this OCCT version.
+	einfo 'Dumping OCCT symbol tables...'
+	${EPYTHON} dump_symbols.py "${_OCCT_DUMP_SYMBOLS_ROOT_DIR}"
```
AndrewAmmerlaan, line 130: `|| die "Failed dumping symbols` or something similar

dev-libs/lief/lief-0.11.5.ebuild
```diff
@@ -0,0 +1,125 @@
...
+inherit python-single-r1 cmake
+
+DESCRIPTION="Library to instrument executable formats"
+HOMEPAGE="https://lief.quarkslab.com"
+SRC_URI="https://github.com/lief-project/LIEF/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="c examples +python static-libs"
```
AndrewAmmerlaan, line 22: If GitHub's code-type detection is to be believed, the code is almost entirely c++ instead of plain c. In which case we should name this flag `cxx`, which also happens to be a global flag so in that case we can drop the flag description in `metadata.xml`: https://github.com/lief-project/LIEF

dev-libs/lief/lief-0.11.5.ebuild
leycec: `</sigh>`

That vestigial holdover from the bad old days (...so, last week) in which I still believed in LIEF's `setup.py` is, indeed, vestigial. Andrew's All-Seeing Eye clarifies all. I'll cauterize the rot away.

dev-python/cadquery-ocp/cadquery-ocp-7.5.2_beta.ebuild
leycec: y u always right about everything

dev-libs/lief/lief-0.11.5.ebuild
leycec: > If GitHub's code-type detection is to be believed, the code is almost entirely c++ instead of plain c.

You remain correct in all things. This baffles me, but I've come to just accept it.

> In which case we should name this flag `cxx`...

I'd love to! I'd like nothing more than to simplify my life. But... doesn't the `cxx` USE flag govern C++ support? The `c` USE flag here actually governs LIEF's C support via `-DLIEF_C_API="$(usex c ON OFF)"`, as odd as a framework with dual C and C++ support seems to even me.

I couldn't actually find a comparable USE flag governing C support in `/usr/portage/profiles/use.desc`. So, I just went with `c`. Is that awful? I bicycled 30km today and am thus on the verge of physical and spiritual collapse, so I accept that I may currently have no idea what's going on here.

dev-libs/lief/lief-0.11.5.ebuild
leycec: **Oh.** I didn't even know `cmake-multilib` was a thing! Upstream's CMake support is non-standard and seemingly untested. So, I must confess I actually have no idea what (if anything) `-DLIEF_FORCE32=ON` is doing.

Let's see how far down the rabbit hole I get before throwing up.

leycec: **Fascinating.** `-DLIEF_FORCE32=ON` just internally passes `-m32` to the linker. There's probably a *much* saner way for us to do that, so let's see what `cmake-multilib` has to say about this.

leycec:
**Stunning success!** Thanks to Science Sauron's gutting review, all of the bad things have been forced to see the error of their ways:

* **Removed vestigial `distutils-r1` references.** Suck it, `DISTUTILS_OPTIONAL=1`.
* **Packaged LIEF with `cmake-multilib`.** Thanks to the magic of Michał Górny, <sup>...seriously, if this guy ever dies we're boned</sup> it was *really* trivial to refactor our LIEF ebuild from `cmake` to `cmake-multilib`. I just renamed `src_configure()` to `multilib_src_configure()`, forced `-DLIEF_FORCE32=OFF`, and let the `*multilib*` family of eclasses handle everything. I still can't believe that worked. It's *never* that easy – but somehow it was.
* **Protected `dump_symbols.py`.** The call to `dump_symbols.py` in our OCP ebuild has now been suffixed by `|| die`, narrowly avoiding silent disaster. :man_facepalming: 

I will now sleep for twelve hours next to a snoring Maine Coon cat.

dev-libs/lief/lief-0.11.5.ebuild
AndrewAmmerlaan: > I'd love to! I'd like nothing more than to simplify my life. But... doesn't the cxx USE flag govern C++ support? The c USE flag here actually governs LIEF's C support via -DLIEF_C_API="$(usex c ON OFF)", as odd as a framework with dual C and C++ support seems to even me.

Right, I thought maybe the `-DLIEF_C_API` option toggled both the c++ part and the c part, but if it toggles only the c part, then I think `c` would indeed be the more appropriate name for the USE flag.

AndrewAmmerlaan:
Nice, this looks very good.

When emerging, I'm getting an error in the configure phase:
```
>>> Configuring source in /var/tmp/portage/dev-libs/lief-0.11.5/work/LIEF-0.11.5 ...
* abi_x86_32.x86: running multilib-minimal_abi_src_configure
* Working in BUILD_DIR: "/var/tmp/portage/dev-libs/lief-0.11.5/work/lief-0.11.5_build-abi_x86_32.x86"
cmake -C /var/tmp/portage/dev-libs/lief-0.11.5/work/lief-0.11.5_build-abi_x86_32.x86/gentoo_common_config.cmake -G Ninja -DCMAKE_INSTALL_PREFIX=/usr -DLIEF_COVERAGE=OFF -DLIEF_DISABLE_FROZEN=OFF -DLIEF_EXTRA_WARNINGS=OFF-DLIEF_FORCE32=OFF -DLIEF_PROFILING=OFF -DLIEF_SUPPORT_CXX14=ON -DLIEF_USE_CCACHE=OFF -DLIEF_ELF=ON -DLIEF_PE=ON -DLIEF_MACHO=ON -DLIEF_ART=ON -DLIEF_DEX=ON -DLIEF_OAT=ON -DLIEF_VDEX=ON -DBUILD_SHARED_LIBS=ON -DLIEF_C_API=ON -DLIEF_EXAMPLES=OFF -DLIEF_FORCE_API_EXPORTS=ON -DLIEF_PYTHON_API=ON -DLIEF_INSTALL_PYTHON=ON -DLIEF_ENABLE_JSON=OFF -DLIEF_DOC=OFF -DLIEF_FUZZING=OFF -DLIEF_INSTALL_COMPILED_EXAMPLES=OFF -DLIEF_LOGGING=OFF -DLIEF_LOGGING_DEBUG=OFF -DLIEF_TESTS=OFF -LIEF_ASAN=OFF -LIEF_LSAN=OFF -LIEF_TSAN=OFF -LIEF_USAN=OFF -DPYTHON_EXECUTABLE=/usr/bin/python3.8 -DCMAKE_BUILD_TYPE=Gentoo -DCMAKE_TOOLCHAIN_FILE=/var/tmp/portage/dev-libs/lief-0.11.5/work/lief-0.11.5_build-abi_x86_32.x86/gentoo_toolchain.cmake  /var/tmp/portage/dev-libs/lief-0.11.5/work/LIEF-0.11.5
CMake Error: Unknown argument -LIEF_USAN=OFF
CMake Error: Run 'cmake --help' for all supported options.
* ERROR: dev-libs/lief-0.11.5::localrepo failed (configure phase):
```
Perhaps these should be `-DLIEF_USAN=OFF` instead?

leycec:
> Perhaps these should be `-DLIEF_USAN=OFF` instead?

*the wet meaty sound of a forehead striking a keyboard*

I see that you are correct again. LIEF happily built for me with all of these broken CMake options, which makes no sense. But this is LIEF. I've learned to question nothing.

This has now been fixed, because this was just embarrassing. `LIEF_[ALTU]SAN` options are now passed sanely. Thus ends the tale of two build systems. CMake wins! `setuptools`, go hang your head in shame.

We march on CadQuery itself later this week. Will @leycec exhaustedly crawl into a ditch before reaching his destination? Will Andrew have second thoughts about his increasingly taxing leadership role? Will @waebbl choose to maintain all of this in @leycec's noble absence? Tune in this Friday to read the stunning conclusion to... *The CQ Tales.* :drum:

blshkv:
And at the end of these woodoo, you will know why i was not willing to write a perfect ebuild 🤣

AndrewAmmerlaan:
Perfect, Thanks :+1: 

I discovered a small problem if `USE="python abi_x86_32"`:
```
-- Found PythonInterp: /usr/bin/python3.8 (found suitable version "3.8.11", minimum required is "3.5") 
CMake Error at cmake/FindPythonLibsNew.cmake:125 (message):
Python config failure: Python is 64-bit, chosen compiler is 32-bit
Call Stack (most recent call first):
api/python/CMakeLists.txt:21 (find_package)
```
At the moment the Python ebuild does not appear to support multilib, which means that cmake can't find a correct python version when attempting to build LIEF multilib. I have added a REQUIRED_USE blocker in a follow-up commit to explicitly prevent this combination of flags from being used.

leycec:
> And at the end of these woodoo, you will know why i was not willing to write a perfect ebuild :rofl:

You walked the road of sanity. If only I had followed you.

Thanks so much, by the way, for all of your gruelling work co-maintaining Pentoo! You give the Gentoo world a good name, Anton. :+1: :+1: :+1: 

> I discovered a small problem if `USE="python abi_x86_32":`

**Yikes.** I... may have been negligent there. I forgot to double-check the `cmake-multilib` refactoring on `x86`. Much apologies for foisting that grim work onto you. Much more thanks for resolving this on your end. Because you saved me a great hardship, CadQuery phase 3 is on schedule and still due to land tomorrow. Let's see if I lie as badly as I test.

PR #1105, opened 2021-08-05
sci-biology/SEECER: EAPI bump, patch, remove files, fix build system

lucasmitrak:
* EAPI bump
* Remove "rename_jellyfish_binary.patch" from ebuild which does nothing
* Add a patch that increases the max sequence length
* Add a patch that removes CXXFLAGS, CFLAGS, LDFLAGS
* Add tc-check-openmp so the ebuild fails without OpenMP support
* Remove copying a new replace_ids.cc file which is now a patch
* Delete all object files
* Delete packaged seqan
* Add eautoreconf to recreate the build system
* Add system seqan include path and CXXFLAGS to econf

Currently, sci-biology/SEECER will not compile due to object files
remaining in the package. This was fixed by deleting all object files
and so forcing them to be rebuilt. The packaged seqan was deleted as
well so the system's seqan is used. In order to achieve this, the
system's seqan include path was passed into econf as
"/usr/include/seqan". The user's CXXFLAGS were passed into econf also.
The patch SEECER-remove-flags.patch removes any other CXXFLAGS, CFLAGS,
and LDFLAGS present in configure.ac and Makefile.am.
Since configure.ac and Makefile.am were updated, the build system was
recreated using eautoreconf -i to account for this. The -i was needed
because the "compile" file is missing, so it had to be added. The patch
SEECER-increase-max-sequence-length.patch increases the max sequence
length from 256 to 10000. This patch was made from an updated
replace_ids.cc downloaded from an archived upstream mail list
that is no longer available online [1]. The previous revision ebuild
included this same replace_ids.cc file; however, despite it appearing
as a text file, it was actually a gzipped file with an incorrect
extension. In addition, this gzipped file was copied into the src/
directory when the replace_ids.cc file is in the pipeline/ directory.
The patch file "rename_jellyfish_binary.patch" also did nothing
but modify a comment so it was removed as well. In addition,
a tc-check-openmp was added so the ebuild fails without OpenMP support.
OpenMP cannot be added as a use flag or a configure enable flag because
gomp is always enabled and the source code assumes the system has OpenMP
installed. Finally, the EAPI was bumped.

[1] URL Split into three parts for the sake of the commit message:
https://web.archive.org/web/20150911132231/
http://seecer-rna-read-error-correction-mailing-list.21961.x6.nabble.com
/Segmentation-fault-in-step-4-td41.html

Closes: https://bugs.gentoo.org/755776

### DIFF
--- a/sci-biology/SEECER/SEECER-0.1.3-r3.ebuild
+++ b/sci-biology/SEECER/SEECER-0.1.3-r3.ebuild
@@ -1,0 +1,66 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit autotools toolchain-funcs
+
+DESCRIPTION="SEquence Error Corrector for RNA-Seq reads"
+HOMEPAGE="http://sb.cs.cmu.edu/seecer/"
+SRC_URI="
+	http://sb.cs.cmu.edu/seecer/downloads/"${P}".tar.gz
+	http://sb.cs.cmu.edu/seecer/downloads/manual.pdf -> "${PN}"-manual.pdf"
+
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64"
+
+# although has bundled jellyfish-1.1.11 copy it just calls the executable during runtime
+# seems jellyfish-2 does not accept same commandline arguments
+DEPEND="
+	sci-libs/gsl:0=
+	sci-biology/seqan:0="
+RDEPEND="${DEPEND}
+	sci-biology/jellyfish:1"
+
+S="${WORKDIR}/${P}/SEECER"
+
+PATCHES=(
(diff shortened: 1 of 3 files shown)


AndrewAmmerlaan:
Thanks :+1:

sci-biology/SEECER/SEECER-0.1.3-r3.ebuild
```diff
@@ -0,0 +1,66 @@
...
+}
+
+pkg_setup(){
+	[[ ${MERGE_TYPE} != "binary" ]] && tc-check-openmp
+}
+
+src_prepare(){
+	rm -f bin/.run*.swp || die
+	rm -f src/*.o pipeline/*.o || die
+	rm -rf seqan-cxx || die
```
AndrewAmmerlaan, line 48: `-f` makes `rm` exit successfully even if the files do not exist, as such it makes the `|| die` useless. Therefore, I have removed the `-f` before merging.

PR #1107, opened 2021-08-05
sci-mathematics/flocq: EAPI bump, add multiprocessing to remake

lucasmitrak:
* EAPI bump from 7 to 8
* Add multiprocessing to remake using the $(makeopts_jobs) variable

Currently, sci-mathematics/flocq does not have multiprocessing even
though the homepage [1] states this is available. This is done by
adding --jobs=$(makeopts_jobs) to remake. the makeopts_jobs
variable comes from the multiprocessing eclass.

[1] http://flocq.gforge.inria.fr/

### DIFF
--- a/sci-mathematics/flocq/flocq-3.4.0.ebuild
+++ b/sci-mathematics/flocq/flocq-3.4.0.ebuild
@@ -1,7 +1,9 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=7
+EAPI=8
+
+inherit multiprocessing
 
 DESCRIPTION="A floating-point formalization for the Coq system"
 HOMEPAGE="http://flocq.gforge.inria.fr/"
@@ -26,7 +28,7 @@ src_configure() {
 }
 
 src_compile() {
-	./remake || die "emake failed"
+	./remake --jobs=$(makeopts_jobs) || die "emake failed"
 }
 
 src_install() {


lucasmitrak:
I was investigating this package for a bug report which is now closed and saw I could add multiprocessing so I quickly made this PR. However, I forgot, am I supposed to bump the revision on the ebuild if I make a change like this? If so, maybe its not worth it?

AndrewAmmerlaan:
> I was investigating this package for a bug report which is now closed and saw I could add multiprocessing so I quickly made this PR. However, I forgot, am I supposed to bump the revision on the ebuild if I make a change like this? If so, maybe its not worth it?

Adding the multiprocessing does not require a revision bump because this only effects build time, not the installed result. The EAPI bump on the other hand does require a revision bump, since a newer EAPI can change any number of things which may effect the installed result.

AndrewAmmerlaan:
> > I was investigating this package for a bug report which is now closed and saw I could add multiprocessing so I quickly made this PR. However, I forgot, am I supposed to bump the revision on the ebuild if I make a change like this? If so, maybe its not worth it?
> 
> Adding the multiprocessing does not require a revision bump because this only effects build time, not the installed result. The EAPI bump on the other hand does require a revision bump, since a newer EAPI can change any number of things which may effect the installed result.

In any case, the compile time is not very long so I think the revision bump is not that big of an issue. I have merged this and amended a revision bump.

Thanks :+1:

lucasmitrak:
> 
> 
> > I was investigating this package for a bug report which is now closed and saw I could add multiprocessing so I quickly made this PR. However, I forgot, am I supposed to bump the revision on the ebuild if I make a change like this? If so, maybe its not worth it?
> 
> Adding the multiprocessing does not require a revision bump because this only effects build time, not the installed result. The EAPI bump on the other hand does require a revision bump, since a newer EAPI can change any number of things which may effect the installed result.

Ok we so add a revision bump if if the installed result might change, and an EAPI could change the installed result. Thank you for your explanation because I was unclear on that matter.

PR #1109, opened 2021-08-09
sci-mathematics/acl2: fix TAGS install, add emacs/doc flags, etc

lucasmitrak:
* EAPI bump
* Remove eutils because it is no longer necessary and deprecated
* inherit elisp-common for compiling/installing emacs files
* Add emacs use flag for emacs files and TAGS file
* Add app-editor/emacs to BDEPEND if its needed during compilation
* Add doc use flag for compiling and installing html docs
* Add dev-lang/perl as a DEPEND for doc
* Change SRC_URI to upstream and updated github
* Update metadata.xml to this new SRC_URI
* Add patch which changes `make` to $(MAKE) in GNUmakefile for Q/A
* Add src_prepare which deletes *.bak, *.orig, and sparc binary acl2link
* Remove `emake certify-books` because certify-books no longer exists
* Add `emake basic` because it is the one which is the default [1]
* emake basic is much faster so the einfos are removed
* Add `emake DOC` to compilation phase if DOC use flag is enabled
* Add `elisp-compile` to compilation phase if emacs use flag is enabled
* Install README.md always and html docs if doc use flag is enabled
* Install emacs files and TAGS file is emacs use flag is enabled
* Add `use emacs && elisp-site-regen` to postinst() and postrm() phases

Currently, sci-mathematics/acl2 will not always install successfully.
This is because the TAGS file is only created if the etags program is
installed. The etags program is installed when emacs is
installed. Therefore, an emacs use flag is added so the TAGS file is
installed only when it is enabled. Since app-editors/emacs is required
for the CBUILD, the dependency is added to the BDEPEND. In order to
compile/install emacs files, the elisp-common eclass is inherited.
This compilation and installation only takes place if the emacs use
flag is enabled. The `elisp-site-regen` is added to pkg_postinst() and
pkg_postrm() with the requirement of the emacs use flag being enabled.
If the doc flag is enabled, then `emake DOC` is executed and the html
docs are installed; however, the README.md is always installed. Since
dev-lang/perl is needed for this compilation, it is added as a
conditional requirement. If the books use flag is enabled, then the
books will be compiled using `emake basic`. The reason `emake basic`
was chosen instead of `emake regression` is because the latter is
stated as being "usually unnecessary" [1] and that "most users will not
want to use this target" [2] while the former is the default [1]
and stated as "a convenient starting place" [3]. Every book is made
before for major releases, so it is unecessary to to build all for an
install. Since the compilation of the books is much faster, the einfo
messages are removed. In addition, the SRC_URI for the previous ebuild
is behind in commits [4] and is simply a clone. Therefore, the main
github repo replaced it. The metadata is updated to this change also.
A src_prepare phase was added which copies debian's Changelog of this
package [5]. The entry on 2015-10-21 of this Changelog shows that
the *.bak and *.orig files were removed to clean the target. Also,
the inadvertent sparc binary acl2link was removed. This sparc binary
lacks source code and therefore violates the GNU GPL. Debian filed a
bug about this binary and the solution was to delete it [6].
A patch is added so a Q/A warning is not given: "make[1]: warning:
jobserver unavailable: using -j1.  Add '+' to parent make rule".
The patch changes a make command to use the variable $(MAKE). This
patch will be made available upstream via a github pull request.
Finally, eutils is removed due to being depreciated and the EAPI is
bumped.

[1] https://www.cs.utexas.edu/users/moore/acl2/v8-3/combined-manual/
index.html?topic=ACL2____BOOKS-CERTIFICATION
[2] https://github.com/acl2/acl2/blob/8.3/books/GNUmakefile#L61
[3] https://github.com/acl2/acl2/blob/8.3/books/GNUmakefile#L48
[4] https://github.com/acl2-devel/acl2-devel
[5] https://debian.pkgs.org/9/debian-main-arm64/
acl2_7.2dfsg-3_arm64.deb.html
[6] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=787368

Closes: https://bugs.gentoo.org/755368

### DIFF
--- a/sci-mathematics/acl2/acl2-8.3-r1.ebuild
+++ b/sci-mathematics/acl2/acl2-8.3-r1.ebuild
@@ -1,0 +1,86 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit elisp-common
+
+DESCRIPTION="Industrial strength theorem prover"
+HOMEPAGE="https://www.cs.utexas.edu/users/moore/acl2/"
+SRC_URI="https://github.com/acl2/acl2/archive/${PV}/${P}.tar.gz"
+
+SLOT="0"
+LICENSE="BSD"
+KEYWORDS="~amd64 ~x86"
+IUSE="books doc emacs"
+
+BDEPEND="
+	dev-lisp/sbcl
+	emacs? ( >=app-editors/emacs-23.1:* )
+"
+DEPEND="
+	dev-lisp/sbcl:=
+	books? ( dev-lang/perl )
+	doc? ( dev-lang/perl )
+"
+RDEPEND="${DEPEND}"
+
+PATCHES=( "${FILESDIR}"/${PN}-use_make_variable.patch )
+
+src_prepare() {
+	find . -type f -name "*.bak" -delete
+	find . -type f -name "*.orig" -delete
+	# Remove sparc binary inadvertently included in upstream
+	rm books/workshops/2003/schmaltz-al-sammane-et-al/support/acl2link || die
+	default
+}
+
+src_compile() {
+	emake LISP="sbcl --noinform --noprint \
+		--no-sysinit --no-userinit --disable-debugger"
+
+	if use books; then
+		emake "ACL2=${S}/saved_acl2" basic
+	fi
+
+	if use doc; then
+		emake "ACL2=${S}/saved_acl2" DOC
+	fi
+
+	if use emacs; then
+		elisp-compile emacs/*.el
+	fi
+}
+
+src_install() {
+	local SAVED_NAME=saved_acl2
+	sed -e "s:${S}:/usr/share/acl2:g" -i ${SAVED_NAME} || die
+	dobin ${SAVED_NAME}
+
+	insinto /usr/share/acl2
+	doins ${SAVED_NAME}.core
+	if use books; then
+		sed -e "/5/a export ACL2_SYSTEM_BOOKS=/usr/share/acl2/books/" \
+			-i ${SAVED_NAME} || die
+		doins -r books
+	fi
+
+	DOCS=( books/README.md )
+	if use doc; then
+		HTML_DOCS=( doc/HTML/. )
+	fi
+	einstalldocs
+
+	if use emacs; then
+		elisp-install ${PN} emacs/*{.el,elc}
+		doins TAGS
+	fi
+}
+
+pkg_postinst() {
+	use emacs && elisp-site-regen
+}
+
+pkg_postrm() {
+	use emacs && elisp-site-regen
+}
--- a/sci-mathematics/acl2/files/acl2-use_make_variable.patch
+++ b/sci-mathematics/acl2/files/acl2-use_make_variable.patch
@@ -1,0 +1,13 @@
+use make variable to avoid QA issue: "make[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule"
+
+--- a/GNUmakefile
++++ b/GNUmakefile
+@@ -576,7 +576,7 @@ doc/home-page.html: doc/home-page.lisp
+ # xdoc::save that populates doc/manual/ (not under books/).
+ acl2-manual: check-books
+ 	rm -rf doc/manual books/system/doc/acl2-manual.cert
+-	cd books ; make USE_QUICKLISP=1 system/doc/acl2-manual.cert
++	cd books ; $(MAKE) USE_QUICKLISP=1 system/doc/acl2-manual.cert
+ 	rm -rf doc/manual/download/*
+ 
+ # WARNING: The dependency list just below isn't complete, since it
--- a/sci-mathematics/acl2/metadata.xml
+++ b/sci-mathematics/acl2/metadata.xml
@@ -20,6 +20,6 @@
 		<flag name="books">build community books, the canonical collection of open-source libraries</flag>
 	</use>
 	<upstream>
-		<remote-id type="github">acl2-devel/acl2-devel</remote-id>
+		<remote-id type="github">acl2/acl2</remote-id>
 	</upstream>
 </pkgmetadata>


lucasmitrak:
Should the commit message be shorter for the sake of `git log`?

sci-mathematics/acl2/acl2-8.3-r1.ebuild
```diff
@@ -0,0 +1,78 @@
...
+	doins ${SAVED_NAME}.core
+	if use books; then
+		sed -e "/5/a export ACL2_SYSTEM_BOOKS=/usr/share/acl2/books/" \
+			-i ${SAVED_NAME} || die
+		doins -r books
+	fi
+
+	DOCS=( books/README.md )
+	if use doc; then
+		HTML_DOCS=( doc/HTML/. )
```
AndrewAmmerlaan, line 62: Technically these should be global vars, but since there is no way to make this conditional in global, I guess we'll have to do it like this.

sci-mathematics/acl2/acl2-8.3-r1.ebuild
```diff
@@ -0,0 +1,78 @@
...
+		emake "ACL2=${S}/saved_acl2" DOC
+	fi
+
+	if use emacs; then
+		elisp-compile emacs/*.el
+	fi
+}
+
+src_install() {
+	SAVED_NAME=saved_acl2
```
AndrewAmmerlaan, line 48:
```suggestion
	local SAVED_NAME=saved_acl2
```

AndrewAmmerlaan:
> Should the commit message be shorter for the sake of `git log`?

AFAIK, there is no maximum length rule for the commit message. However, you can make it shorter if you want.

I'm getting some QA notices:
```
* QA Notice: Unresolved soname dependencies:
* 
*      /usr/share/acl2/books/workshops/2003/schmaltz-al-sammane-et-al/support/acl2link: libc.so.1 libm.so.1 libnsl.so.1 libsocket.so.1
*
* QA Notice: The following files contain writable and executable sections
*  Files with such sections will not work properly (or at all!) on some
*  architectures/operating systems.  A bug should be filed at
*  https://bugs.gentoo.org/ to make sure the issue is fixed.
*  For more information, see:
*
*    https://wiki.gentoo.org/wiki/Hardened/GNU_stack_quickstart
*
*  Please include the following list of files in your report:
*  Note: Bugs should be filed for the respective maintainers
*  of the package in question and not hardened@gentoo.org.
* --- --- RWX usr/share/acl2/books/workshops/2003/schmaltz-al-sammane-et-al/support/acl2link
* QA Notice: Files built without respecting CFLAGS have been detected
*  Please include the following list of files in your report:
* /usr/share/acl2/books/workshops/2003/schmaltz-al-sammane-et-al/support/acl2link
* QA Notice: Files built without respecting LDFLAGS have been detected
*  Please include the following list of files in your report:
* /usr/share/acl2/books/workshops/2003/schmaltz-al-sammane-et-al/support/acl2link
```
The later two can be reproduced by having a testing env with something like this:
```
CFLAGS="-march=native -O2 -pipe -frecord-gcc-switches"
CXXFLAGS="${CFLAGS}"

# Defining all four *FLAGS variables enables a portage QA check reporting when these variables are not respected
FFLAGS="${CFLAGS}"
FCFLAGS="${FFLAGS}"

# Enables a Portage QA check to report when LDFLAGS is not respected
LDFLAGS="${LDFLAGS} -Wl,--defsym=__gentoo_check_ldflags__=0"
```
The unresolved soname notice indicates that there is probably a missing runtime dependency. It looks like the file in question might be pre-built, in which case we should either fix the build system to properly build it (respecting CFLAGS and LDFLAGS) or add this file to the `QA_PREBUILT` variable to tell portage that this file is pre-built and can be excluded from the QA checks.

sci-mathematics/acl2/acl2-8.3-r1.ebuild
lucasmitrak: This change was implemented with a squashed commit.

lucasmitrak:
> 
> 
> > Should the commit message be shorter for the sake of `git log`?
> 
> AFAIK, there is no maximum length rule for the commit message. However, you can make it shorter if you want.
> 
> I'm getting some QA notices:
> 
> ```
> * QA Notice: Unresolved soname dependencies:
> * 
> *      /usr/share/acl2/books/workshops/2003/schmaltz-al-sammane-et-al/support/acl2link: libc.so.1 libm.so.1 libnsl.so.1 libsocket.so.1
> *
> * QA Notice: The following files contain writable and executable sections
> *  Files with such sections will not work properly (or at all!) on some
> *  architectures/operating systems.  A bug should be filed at
> *  https://bugs.gentoo.org/ to make sure the issue is fixed.
> *  For more information, see:
> *
> *    https://wiki.gentoo.org/wiki/Hardened/GNU_stack_quickstart
> *
> *  Please include the following list of files in your report:
> *  Note: Bugs should be filed for the respective maintainers
> *  of the package in question and not hardened@gentoo.org.
> * --- --- RWX usr/share/acl2/books/workshops/2003/schmaltz-al-sammane-et-al/support/acl2link
> * QA Notice: Files built without respecting CFLAGS have been detected
> *  Please include the following list of files in your report:
> * /usr/share/acl2/books/workshops/2003/schmaltz-al-sammane-et-al/support/acl2link
> * QA Notice: Files built without respecting LDFLAGS have been detected
> *  Please include the following list of files in your report:
> * /usr/share/acl2/books/workshops/2003/schmaltz-al-sammane-et-al/support/acl2link
> ```
> 
> The later two can be reproduced by having a testing env with something like this:
> 
> ```
> CFLAGS="-march=native -O2 -pipe -frecord-gcc-switches"
> CXXFLAGS="${CFLAGS}"
> 
> # Defining all four *FLAGS variables enables a portage QA check reporting when these variables are not respected
> FFLAGS="${CFLAGS}"
> FCFLAGS="${FFLAGS}"
> 
> # Enables a Portage QA check to report when LDFLAGS is not respected
> LDFLAGS="${LDFLAGS} -Wl,--defsym=__gentoo_check_ldflags__=0"
> ```
> 
> The unresolved soname notice indicates that there is probably a missing runtime dependency. It looks like the file in question might be pre-built, in which case we should either fix the build system to properly build it (respecting CFLAGS and LDFLAGS) or add this file to the `QA_PREBUILT` variable to tell portage that this file is pre-built and can be excluded from the QA checks.



> 
> 
> > Should the commit message be shorter for the sake of `git log`?
> 
> AFAIK, there is no maximum length rule for the commit message. However, you can make it shorter if you want.
> 
> I'm getting some QA notices:
> 
> ```
> * QA Notice: Unresolved soname dependencies:
> * 
> *      /usr/share/acl2/books/workshops/2003/schmaltz-al-sammane-et-al/support/acl2link: libc.so.1 libm.so.1 libnsl.so.1 libsocket.so.1
> *
> * QA Notice: The following files contain writable and executable sections
> *  Files with such sections will not work properly (or at all!) on some
> *  architectures/operating systems.  A bug should be filed at
> *  https://bugs.gentoo.org/ to make sure the issue is fixed.
> *  For more information, see:
> *
> *    https://wiki.gentoo.org/wiki/Hardened/GNU_stack_quickstart
> *
> *  Please include the following list of files in your report:
> *  Note: Bugs should be filed for the respective maintainers
> *  of the package in question and not hardened@gentoo.org.
> * --- --- RWX usr/share/acl2/books/workshops/2003/schmaltz-al-sammane-et-al/support/acl2link
> * QA Notice: Files built without respecting CFLAGS have been detected
> *  Please include the following list of files in your report:
> * /usr/share/acl2/books/workshops/2003/schmaltz-al-sammane-et-al/support/acl2link
> * QA Notice: Files built without respecting LDFLAGS have been detected
> *  Please include the following list of files in your report:
> * /usr/share/acl2/books/workshops/2003/schmaltz-al-sammane-et-al/support/acl2link
> ```
> 
> The later two can be reproduced by having a testing env with something like this:
> 
> ```
> CFLAGS="-march=native -O2 -pipe -frecord-gcc-switches"
> CXXFLAGS="${CFLAGS}"
> 
> # Defining all four *FLAGS variables enables a portage QA check reporting when these variables are not respected
> FFLAGS="${CFLAGS}"
> FCFLAGS="${FFLAGS}"
> 
> # Enables a Portage QA check to report when LDFLAGS is not respected
> LDFLAGS="${LDFLAGS} -Wl,--defsym=__gentoo_check_ldflags__=0"
> ```
> 
> The unresolved soname notice indicates that there is probably a missing runtime dependency. It looks like the file in question might be pre-built, in which case we should either fix the build system to properly build it (respecting CFLAGS and LDFLAGS) or add this file to the `QA_PREBUILT` variable to tell portage that this file is pre-built and can be excluded from the QA checks.

In case anyone else reads this, adding those five lines to /etc/portage/env (with the accompanying assignment in /etc/portage/package.env) will let portage display a QA message but it is easy to miss because a text can pass by, so I recommend also adding `PORTAGE_ELOG_CLASSES="log warn error qa"`.
The inadvertent sparc binary acl2link was removed. This sparc binary lacks source code and therefore violates the GNU GPL. Debian filed a bug about this binary and the solution was to delete it.
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=787368

AndrewAmmerlaan:
Very Nice, Thanks :+1: 

I bumped the ebuild to version 8.4 while merging

PR #1111, opened 2021-09-05
sci-biology/open-ephys-gui: version bump

malfatti:
This will bump `sci-biology/open-ephys-gui` to version `0.5.5`. This allows for dropping the `<gcc-9` dependency.


sci-biology/open-ephys-gui/open-ephys-gui-0.5.5.ebuild
```diff
@@ -0,0 +1,79 @@
...
+	media-libs/freeglut
+	media-libs/freetype
+	x11-libs/libXrandr
+	x11-libs/libXcursor
+	x11-libs/libXinerama
+	jack? ( || ( media-sound/jack-audio-connection-kit media-sound/jack2 ) )
+"
+RDEPEND="${DEPEND}"
+
+BUILD_DIR="$S/Build"
```
AndrewAmmerlaan, line 45:
```suggestion
BUILD_DIR="${S}/Build"
```

Btw, why do we override this? Can't we use the default?

sci-biology/open-ephys-gui/open-ephys-gui-0.5.5.ebuild
```diff
@@ -0,0 +1,79 @@
...
+
+src_configure() {
+	local mycmakeargs=( -DCMAKE_SKIP_RPATH=ON )
+	cmake_src_configure
+}
+
+src_install() {
+	dodir opt/open-ephys-"$Suffix"/ lib/udev/rules.d/
+	cp -R "${BUILD_DIR}"/Gentoo/* "${ED}"/opt/open-ephys-"$Suffix"/
+	cp -R "${WORKDIR}"/"${SubDir}"/Resources/Scripts/40-open-ephys.rules "${ED}"/lib/udev/rules.d/
```
AndrewAmmerlaan, line 70:
```suggestion
	udev_dorules "${SubDir}"/Resources/Scripts/40-open-ephys.rules
```

From `udev.eclass`

sci-biology/open-ephys-gui/open-ephys-gui-0.5.5.ebuild
```diff
@@ -0,0 +1,79 @@
...
+	dosym ../../opt/open-ephys-"$Suffix"/open-ephys usr/bin/open-ephys-"$Suffix"
+}
+
+pkg_postinst() {
+	ewarn " "
+	ewarn "You must restart the udev service in order to allow your computer to"
+	ewarn "communicate with the Open Ephys acquisition board."
+	ewarn " "
```
AndrewAmmerlaan, line 78:
```suggestion
	ewarn " "
	udev_reload
```

sci-biology/open-ephys-gui/open-ephys-gui-0.5.5.ebuild
```diff
@@ -0,0 +1,79 @@
...
+}
```
AndrewAmmerlaan, line 79:
```suggestion
}

pkg_postrm() {
     udev_reload
}

```

sci-biology/open-ephys-gui/open-ephys-gui-0.5.5.ebuild
```diff
@@ -0,0 +1,79 @@
+# Copyright 2019-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
AndrewAmmerlaan, line 4: While you're at it, could you please bump the EAPI of this ebuild? I don't expect any issues with EAPI 8 for this ebuild so it should be trivial in this case.

sci-biology/open-ephys-gui/open-ephys-gui-0.5.5.ebuild
```diff
@@ -0,0 +1,79 @@
...
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/open-ephys/plugin-GUI"
+	EGIT_BRANCH="master"
+	Suffix=$EGIT_BRANCH
+	SubDir=${P}
+elif [[ ${PV} == "99999999" ]] ; then
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/open-ephys/plugin-GUI"
+	EGIT_BRANCH="development"
+	Suffix=$EGIT_BRANCH
```
AndrewAmmerlaan, line 21:
```suggestion
	Suffix=${EGIT_BRANCH}
```

sci-biology/open-ephys-gui/open-ephys-gui-0.5.5.ebuild
```diff
@@ -0,0 +1,79 @@
...
+
+DESCRIPTION="Processing, recording, and visualizing multichannel ephys data"
+HOMEPAGE="http://www.open-ephys.org/gui/"
+LICENSE="GPL-3"
+
+if [[ ${PV} == "9999" ]] ; then
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/open-ephys/plugin-GUI"
+	EGIT_BRANCH="master"
+	Suffix=$EGIT_BRANCH
```
AndrewAmmerlaan, line 15:
```suggestion
	Suffix=${EGIT_BRANCH}
```

sci-biology/open-ephys-gui/open-ephys-gui-0.5.5.ebuild
```diff
@@ -0,0 +1,79 @@
...
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/open-ephys/plugin-GUI"
+	EGIT_BRANCH="development"
+	Suffix=$EGIT_BRANCH
+	SubDir=${P}
+else
+	SRC_URI="https://github.com/open-ephys/plugin-GUI/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	Suffix=${PV}
+	SubDir="plugin-GUI-${PV}"
+	S="${WORKDIR}/$SubDir"
```
AndrewAmmerlaan, line 27:
```suggestion
	S="${WORKDIR}/${SubDir}"
```

sci-biology/open-ephys-gui/open-ephys-gui-0.5.5.ebuild
```diff
@@ -0,0 +1,79 @@
...
+	fi
+}
+
+src_configure() {
+	local mycmakeargs=( -DCMAKE_SKIP_RPATH=ON )
+	cmake_src_configure
+}
+
+src_install() {
+	dodir opt/open-ephys-"$Suffix"/ lib/udev/rules.d/
```
AndrewAmmerlaan, line 68:
```suggestion
	dodir opt/open-ephys-"${Suffix}"/ lib/udev/rules.d/
```

sci-biology/open-ephys-gui/open-ephys-gui-0.5.5.ebuild
```diff
@@ -0,0 +1,79 @@
...
+	cp -R "${BUILD_DIR}"/Gentoo/* "${ED}"/opt/open-ephys-"$Suffix"/
```
AndrewAmmerlaan, line 69:
```suggestion
	cp -R "${BUILD_DIR}"/Gentoo/* "${ED}"/opt/open-ephys-"${Suffix}"/
```

sci-biology/open-ephys-gui/open-ephys-gui-0.5.5.ebuild
```diff
@@ -0,0 +1,79 @@
...
+	cp -R "${WORKDIR}"/"${SubDir}"/Resources/Scripts/40-open-ephys.rules "${ED}"/lib/udev/rules.d/
+	dosym ../../opt/open-ephys-"$Suffix"/open-ephys usr/bin/open-ephys-"$Suffix"
```
AndrewAmmerlaan, line 71:
```suggestion
	dosym ../../opt/open-ephys-"${Suffix}"/open-ephys usr/bin/open-ephys-"${Suffix}"
```

sci-biology/open-ephys-gui/open-ephys-gui-0.5.5.ebuild
malfatti: Unfortunately not. If we don't, I got this:
```bash
CMake Error at CMakeLists.txt:19 (message):
  Please run cmake inside the Build folder to build the binaries in the
  predefined paths.

  Set the OE_DONT_CHECK_BUILD_PATH cmake variable to TRUE to disable this
  error
```

So I guess it's easier to build it where it expects to be built :)

sci-biology/open-ephys-gui/open-ephys-gui-0.5.5.ebuild
malfatti: I tried, but then for reasons I do not understand, with `EAPI=8` cmake will try to build with an invalid build type. I could not get it to pass the configure phase with `EAPI=8`.

malfatti:
Hey @AndrewAmmerlaan,

Thanks a lot for your reviews. As I commented above, I could not drop the build dir override nor bump the EAPI version. The software expect the build dir to be under `Build/`; and with `EAPI=8` I cannot pass the configure phase (getting `invalid build type` error), even though the patch to fix this is applied correctly. CMake is somehow setting the build type to `RelWithDebInfo` instead of `Gentoo`. It's weird because `RelWithDebInfo` is actually a valid build type. I kept `EAPI=7`, since I could not solve this issue.

Cheers :)

AndrewAmmerlaan:
> Those 2 issues aside, all other suggestions were applied.

Thanks :+1: 

We should still also do `udev_reload` in the `pkg_postrm` phase, because after removing the package we should reload udev to account for the removed rules file. I'm not sure if udev misbehaves if we don't reload it after removing, but other ebuilds are doing this so it is probably the right thing to do.

> EDIT2: I got a file collision with the udev rules, as soon as I fix it I'll update all ebuilds.

You can probably just rename the rules file with `udev_newrules` instead of `udev_dorules`

malfatti:
> We should still also do `udev_reload` in the `pkg_postrm` phase, because after removing the package we should reload udev to account for the removed rules file. I'm not sure if udev misbehaves if we don't reload it after removing, but other ebuilds are doing this so it is probably the right thing to do.

Sorry I scrolled through this, now it is added.

> You can probably just rename the rules file with `udev_newrules` instead of `udev_dorules`

Just did, now all versions are merging and the software works. Thanks a lot :)

AndrewAmmerlaan:
Perfect, Thanks :+1:

Issue #1113, opened 2021-09-18
sci-mathematics/acl2-8.4 fails to compile with `books` or `doc` USE flag enabled


lucasmitrak:
sci-mathematics/acl2 was version bumped during a bug fix PR [1]. However, it seems version 8.3 and 8.4 are too different for the ebuild to be version bumped so easily. Even without these flags, the package does not compile correctly. Therefore, I suggest the original PR is implemented with the original version 8.3. If a version bump is desired, I suggest creating a bug report. Logs are attached.

[1] https://github.com/gentoo/sci/pull/1109
[logs.zip](https://github.com/gentoo/sci/files/7189400/logs.zip)

AndrewAmmerlaan:
Sorry, I can't seem to find the full build log in this zip file, could you attach it please?

Version 8.4 compiled successfully for me before, so I doubt the problem is the version bump. Maybe some incompatibility with a dependency was introduced by some version bump somewhere else? 
[acl2-8.4:20210912-073000.log](https://github.com/gentoo/sci/files/7189494/acl2-8.4.20210912-073000.log)

I'll test again and see if I can reproduce this.

lucasmitrak:
[logs.zip](https://github.com/gentoo/sci/files/7190522/logs.zip)
This zip includes the build log. This was done inside an isolated docker image using ebuildtester. I cannot think of why we are getting a different result. I think I will try installing outside of the docker image to see if it makes a difference; I will look into this later.

lucasmitrak:
@AndrewAmmerlaan  You are running sbcl version 2.1.8 according to the log file you uploaded which states "This is SBCL 2.1.8". acl2-8.4 requires a sbcl version greater then 1.5.2. The unmasked version is 1.4.9 and the next version after that is 2.1.5. So I will create a PR which adds requests a sbcl version greater then or equal to 1.5.2 and put all my evidence of this requirement there.

PR #1114, opened 2021-09-19
sci-mathematics/acl2: Require >=dev-lisp/sbcl-1.5.2 in BDEPEND 

lucasmitrak:
* Require >=dev-lisp/sbcl-1.5.2 due to addition of "--tls-limit 16384"

Currently, sci-mathematics/acl2-8.4 will not compile with the current
stable dev-lisp/sbcl-1.4.9 due to the addition of "--tls-limit 16384"
to sbcl within the created file saved_acl2. This addition is documented
in the acl-8.4 release notes [1] as "Increased the number of special
variables that can be created". Within sbcl, this new feature was added
in version 1.5.2. The news for that version [2] reads "enhancement:
command-line option '--tls-limit' can be used". Therefore attempting
to use the argument "--tls-limit 16384" with sbcl versions prior to
1.5.2 causes it to fail. Thus a minimum version requirement is needed
for acl2-8.4.

[1] https://www.cs.utexas.edu/users/moore/acl2/v8-4/combined-manual/
index.html?topic=ACL2____NOTE-8-4
[2] http://www.sbcl.org/all-news.html

Closes: #1113 

### DIFF
--- a/sci-mathematics/acl2/acl2-8.3.ebuild
+++ b/sci-mathematics/acl2/acl2-8.3.ebuild
@@ -1,51 +1,86 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=7
+EAPI=8
 
-inherit eutils
+inherit elisp-common
 
 DESCRIPTION="Industrial strength theorem prover"
 HOMEPAGE="https://www.cs.utexas.edu/users/moore/acl2/"
-MY_PN=${PN}-devel
-SRC_URI="https://github.com/${MY_PN}/${MY_PN}/releases/download/${PV}/${P}.tar.gz"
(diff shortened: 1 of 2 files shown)


lucasmitrak:
sci-mathematics/acl2: Update 8.3 to the original intended ebuild 

* Update 8.3's ebuild to the one originally intended for 8.3

Currently, sci-mathematics/acl2-8.3 will not compile to the fixed bug
755368 [1]. This bug was fixed in PR 1109 on github [2]. However, due
to a version bump, acl2-8.3 remains unchanged. Therefore the ebuild for
8.3 should be updated to the original PR. acl2-8.3 was tested to
work with the latest unstable version of sbcl-2.1.8. With this commit,
acl2-8.3 will install successfully. Without this commit, it will not.

[1] https://bugs.gentoo.org/755368
[2] #1109

AndrewAmmerlaan:
Thanks, I revbumped the 8.3 ebuild since we are changing its `EAPI`.

PR #1115, opened 2021-10-01
sci-mathematics/acl2: Fix sed expression, clearer books install

lucasmitrak:
* Fix sed expression which does not execute at all
* Move books down to original install location

Currently, the sed expression
`/5/a export ACL2_SYSTEM_BOOKS=/usr/share/acl2/books/` does not modify
the acl2 executable script. The sed expression's intended behavior
seems to be adding the enviroment variable `ACL2_SYSTEM_BOOKS` to the
fifth line on the final script; however, no modification is made.
A new sed expression must replace it, namely
`5iexport ACL2_SYSTEM_BOOKS=/usr/share/acl2/books/` which accomplishes
inserting the enviroment variable into the script. This enviroment
variable is recommended when the books directory resides outside
where the executable was built [1].
In addition, the command `doins -r books` should be moved further down
for the sake of clarity. This way the src_install() function modifies
the files at the top and installs them at the bottom. Also, now
the command `doins -r books` is beside the other similar commands.

[1] https://raw.githubusercontent.com/acl2/acl2/master/doc.lisp

### DIFF
--- a/sci-mathematics/acl2/acl2-8.3-r2.ebuild
+++ b/sci-mathematics/acl2/acl2-8.3-r2.ebuild
@@ -1,0 +1,89 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit elisp-common
+
+DESCRIPTION="Industrial strength theorem prover"
+HOMEPAGE="https://www.cs.utexas.edu/users/moore/acl2/"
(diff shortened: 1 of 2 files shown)


lucasmitrak:
This should be my last PR for acl2.

lucasmitrak:
Another note is that the patch file for this package was submitted and accepted upstream [1] so the next version will hopefully not need it.
[1] https://github.com/acl2/acl2/pull/1318

PR #1128, opened 2021-12-16
net-misc/cernbox-client: Bump to 2.9.2.6206.

olifre:
Version bump to 2.9.2.6206, following `net.misc/owncloud-client`, also bumping to EAPI 8. 
Oldest version and corresponding patch is dropped.


AndrewAmmerlaan:
Thanks :+1:

PR #1123, opened 2021-11-11
sci-libs/pytorch: add PyTorch 1.10.1

Miezhiko:
https://pytorch.org/blog/pytorch-1.10-released/

PR #1130, opened 2021-12-25
sci-libs/pytorch: fix python build with >=setuptools-59.6

littlewu2508:
Backport an upstream PR: https://github.com/pytorch/pytorch/pull/69904

Bug: https://github.com/gentoo/sci/pull/1123#issuecomment-999597677
Package-Manager: Portage-3.0.22, Repoman-3.0.3

### DIFF
--- a/sci-libs/pytorch/files/pytorch-1.10.0-fix-distutils.patch
+++ b/sci-libs/pytorch/files/pytorch-1.10.0-fix-distutils.patch
@@ -1,0 +1,93 @@
+This fix distutils-r1_src_install failure with setuptools-59
+
+From 9af2edb158b3603c44eff6e12896f1d215e8b898 Mon Sep 17 00:00:00 2001
+From: Nikita Shulga <nshulga@fb.com>
+Date: Tue, 14 Dec 2021 07:15:34 -0800
+Subject: [PATCH] Properly import LooseVersion
+
+This fixes regression introduced by https://github.com/pytorch/pytorch/pull/57040
+
+Somehow importing `distutils` from `setuptool` caused import of
+`distutils.versions`, which is not a documented dependency and got
+change with the release of
+[setuptools-59.6.0](https://github.com/pypa/setuptools/tree/v59.6.0)
+We should not rely on that, as
+`import distutils` never re-imports `distutils.version`, which one can
+see by observing
+https://github.com/python/cpython/blob/3.9/Lib/distutils/__init__.py
+or by running:
+```
+% python3 -c "import distutils;print(distutils.__version__, dir(distutils))"
+3.7.5 ['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', '__version__', 'sys']
+% python3 -c "from setuptools import distutils;print(distutils.__version__, dir(distutils))"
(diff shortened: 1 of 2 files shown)


AndrewAmmerlaan:
Thanks :+1:

AndrewAmmerlaan:
There is still something strange happening here:
```
andrew@andrew-gentoo-pc ~ % python3.8
Python 3.8.12 (default, Dec 14 2021, 21:16:01)
[GCC 11.2.1 20211127] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch
File "<stdin>", line 1, in <module>
File "/usr/lib/python3.8/site-packages/torch/__init__.py", line 197, in <module>
from torch._C import *  # noqa: F403
ImportError: /usr/lib64/libtorch_python.so: undefined symbol: PyCMethod_New
>>>
```

```
andrew@andrew-gentoo-pc ~ % python3.9
Python 3.9.9 (main, Nov 22 2021, 16:39:27)
[GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch
zsh: segmentation fault (core dumped)  python3.9
```

```
andrew@andrew-gentoo-pc ~ % python3.10
Python 3.10.1 (main, Dec 18 2021, 12:34:29) [GCC 11.2.1 20211127] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch
>>>
```

It appears we aren't properly compiling pytorch for every enable python target, only the version for the highest python target is installed and works. We either need to fix this or set `DISTUTILS_SINGLE_IMPL=1` in the ebuild.

littlewu2508:
Exactly, I will try to come up with a solution. It's a bit tricky because src_compile is using cmake.eclass so the compilation runs only once. Complete recompile for every python target is also a bad idea because that means tons of redundant work.

AndrewAmmerlaan:
> Exactly, I will try to come up with a solution. It's a bit tricky because src_compile is using cmake.eclass so the compilation runs only once. Complete recompile for every python target is also a bad idea because that means tons of redundant work.

Replacing:
```
src_compile() {
	cmake_src_compile

	if use python; then
		USE_SYSTEM_LIBS=ON CMAKE_BUILD_DIR=${BUILD_DIR} distutils-r1_src_compile
	fi
}
```
with e.g.:
```
python_compile() {
	cmake_src_compile
       	if use python; then
	      USE_SYSTEM_LIBS=ON CMAKE_BUILD_DIR=${BUILD_DIR} distutils-r1_python_compile
         fi
}
```
(and equivalent for the other phases) should to the trick. (The default distutils-r1 `src_compile` will run `python_compile` for each python target if defined, we then also call the default distutils-r1  `python_compile` if USE=python). However, this will triple compile time (if all three targets are enabled), and to be honest I'm not sure that is worth it since the compile time is already rather long. Is there a use case for installing this package for multiple python versions, if not then I suggest to just keep pytorch single implementation only.

littlewu2508:
@AndrewAmmerlaan Thanks for changing pytorch to single python target. I've investigated the possibility of compiling multiple python target bindings without recompile python-independent parts like libtorch. It seems too complicated. The CMakeLists.txt in pytorch and caffe2 are not intended to configure multiple targets, so there would be too much hack to achieve multi-target support. Just single python implementation is enough since no one has wanted multi-target pytorch yet.

heroxbd:
littlewu2508 ***@***.***> writes:

> @AndrewAmmerlaan Thanks for changing pytorch to single python
> target. I've investigated the possibility of compiling multiple python
> target bindings without recompile python-independent parts like
> libtorch. It seems too complicated. The CMakeLists.txt in pytorch and
> caffe2 are not intended to configure multiple targets, so there would
> be too much hack to achieve multi-target support. Just single python
> implementation is enough since no one has wanted multi-target pytorch
> yet.

I am fine with a single python target for now.

PR #1131, opened 2021-12-27
dev-libs/cudnn: 8.3.1.22

Miezhiko:
TODO: 

- [ ]  cuda 10.2 support


AndrewAmmerlaan:
Cuda-10 is masked for removal from ::gentoo, so I wouldn't bother with adding support for that.

PR #1132, opened 2021-12-27
sci-libs/pytorch: fixes for cuda USE flag, nccl update

Miezhiko:


Miezhiko:
(still fails for me)

Miezhiko:
getting same error as here: https://zenn.dev/nbo/scraps/8f1505e365d961 
I think we should pull some patch / workaround for new gcc

AndrewAmmerlaan:
> (still fails for me)

The issue is that pytorch needs clhpp-1, but we only have clhpp-2 in ::gentoo. I added a patch earlier that replaces the includes for clhpp-1 with includes for clhpp-2, but we still need a patch that actually makes it build with clhpp-2.

Miezhiko:
@AndrewAmmerlaan maybe we can package or raise from dead clhpp 1 to avoid patches or those can't live together?

AndrewAmmerlaan:
> @AndrewAmmerlaan maybe we can package or raise from dead clhpp 1 to avoid patches or those can't live together?

clhpp-1 has a file conflict with clhpp-2 so it wouldn't be possible to install both (without patching things in clhpp, which would then also require patching here). I suspect there probably already is a patch for clhpp-2 support in pytorch somewhere, though I haven't been able to find it yet.

Miezhiko:
I see, I even wasn't able to find clhpp 1

I'd try that because I don't have anything else depending on it if but can't find it
```bash
 * These packages depend on clhpp:
sci-libs/pytorch-1.10.1 (opencl ? dev-libs/clhpp)
```

Miezhiko:
> arlier that replaces the includes for clhpp-1 with includes for clhpp-2, but we still need a patch that actually makes it build with clhpp-2.

do you have this in plans? if not maybe I'll try to dig into it when I will have time

AndrewAmmerlaan:
> > arlier that replaces the includes for clhpp-1 with includes for clhpp-2, but we still need a patch that actually makes it build with clhpp-2.
> 
> do you have this in plans? if not maybe I'll try to dig into it when I will have time @AndrewAmmerlaan

It's on the TODO list (but there are a lot of other things on there as well :D )

PR #1133, opened 2022-01-14
sci-libs/pytorch fixed thrust include path for rocm platform

anriha:
Before this change when trying to compile with rocm use flag thrust headers would not be found.

### DIFF
--- a/sci-libs/pytorch/pytorch-1.10.1.ebuild
+++ b/sci-libs/pytorch/pytorch-1.10.1.ebuild
@@ -230,6 +230,7 @@ src_prepare() {
 		export PYTORCH_ROCM_ARCH="${AMDGPU_TARGETS}"
 		sed -e "/set(roctracer_INCLUDE_DIRS/s,\${ROCTRACER_PATH}/include,${EPREFIX}/usr/include/roctracer," \
 			-e "/PYTORCH_HIP_HCC_LIBRARIES/s,\${HIP_PATH}/lib,${EPREFIX}/usr/lib/hip/lib," \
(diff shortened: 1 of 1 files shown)


AndrewAmmerlaan:
As per [Gentoo's Copyright Policy](https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin), could you add a `Signed-off-by` line to your commit message please?

sci-libs/pytorch/pytorch-1.10.1.ebuild
```diff
@@ -230,6 +230,7 @@ src_prepare() {
 		export PYTORCH_ROCM_ARCH="${AMDGPU_TARGETS}"
 		sed -e "/set(roctracer_INCLUDE_DIRS/s,\${ROCTRACER_PATH}/include,${EPREFIX}/usr/include/roctracer," \
 			-e "/PYTORCH_HIP_HCC_LIBRARIES/s,\${HIP_PATH}/lib,${EPREFIX}/usr/lib/hip/lib," \
+			-e "/set(roctracer_INCLUDE_DIRS/a\  set(thrust_INCLUDE_DIRS ${EPREFIX}/usr/include/rocthrust)" \
```
AndrewAmmerlaan, line 233: Does this line make line 231 obsolete? Or do we need both of them?

anriha: We will need both of them the `set(roctracer_INCLUDE_DIRS` is just to add `"/set(roctracer_INCLUDE_DIRS/a\  set(thrust_INCLUDE_DIRS ${EPREFIX}/usr/include/rocthrust)` to the right line. Not sure if patch wouldn't be more appropriate.

AndrewAmmerlaan:
Thanks :+1:

PR #1136, opened 2022-01-16
dev-python/llvmlite: bump version to 0.38.0

littlewu2508:
The llvm max slot is 11. >11 will cause compilation error

Package-Manager: Portage-3.0.30, Repoman-3.0.3


dev-python/llvmlite/llvmlite-0.38.0.ebuild
```diff
@@ -0,0 +1,60 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
AndrewAmmerlaan, line 4: Can we bump this to `EAPI=8` while we are touching the package anyway please?

AndrewAmmerlaan: Aah nevermind, I see `llvm.eclass` doesn't support EAPI 8 yet

AndrewAmmerlaan:
Thanks :+1:

PR #1139, opened 2022-01-18
sci-libs/pydicom: update to 2.2.2, py3.10 support

Miezhiko:


TheChymera:
@Miezhiko thank you!

@AndrewAmmerlaan a test failure will appear after this, but it's not cause by this commit. The repo is already "broken" following the sys-fabric/libibverbs hard-masking by @SoapGentoo . I' looking into whether this can be quickly fixed.

Miezhiko:
ah I wanted to add it to this overlay and maybe add it to package.unmask of overlay
https://github.com/gentoo/sci/pull/1140

PR #1140, opened 2022-01-18
sys-fabric/libibverbs: needed for pytorch (marked for remove in tree)

Miezhiko:


TheChymera:
@Miezhiko @AndrewAmmerlaan as per https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=a8d67eb0ac4ef3dfb726aad6cf527b5c0d4ed03f this was included/replaced with `sys-cluster/rdma-core`. Tested it with pytorch and it works, so the repo has been fixed with https://gitweb.gentoo.org/proj/sci.git/commit/?id=c1b59fbed7d118800f417014de364aed7b98913a and we don't need to bring this package back.

(Unless it's needed by something else, but it doesn't seem to be.)

PR #1143, opened 2022-02-10
fix mmseqs, add foldseek

joelb123:
CPU flags aren't respected by the mmseqs cmake, and attempts to set them manually broke all optimizations, so I backed that out.

Version numbers reported by the binaries are now consistent with upstream.

I have added foldseek, a related code, which got released yesterday.  I think it will be of broad interest.

I added live builds for mmseqs and foldseek.  The repos are moving fast between releases for both.


AndrewAmmerlaan:
The `-DHAVE_AVX2` etc options are ignored? Do we have an upstream issue about this?

sci-biology/mmseqs2/mmseqs2-13.ebuild
```diff
@@ -0,0 +1,48 @@
...
+MY_PV="${PV}-${SUB_PV}"
+
+SRC_URI="https://github.com/soedinglab/${MY_PN}/archive/refs/tags/${MY_PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64"
+
+# This package only builds correctly native and does not respect
+# CPU flags, despite a subset of them being defined as cmake flags.
```
AndrewAmmerlaan, line 21: If we can't force it to respect the cpu flags we set, can we force it to build generic (i.e. with all cpu flags disabled)?

Otherwise we might want to add `RESTRICT="bindist", since binary packages of this likely won't work.

There is a similar issue in[ `sci-biology/mrbayes`](https://github.com/gentoo/gentoo/blob/master/sci-biology/mrbayes/mrbayes-3.2.7.ebuild) that caused me a bit of a headache a couple of weeks back. Current workaround is to force clear `SIMD_FLAGS` and `CPUEXT_FLAGS`, something similar might work here as well.

AndrewAmmerlaan:
Sorry I'm still a bit confused about what is broken. We set `-DNATIVE_ARCH=OFF` which causes `-DHAVE_AVX2` etc to be ignored? but `-DNATIVE_ARCH=OFF` itself does behave as intended and causes a generic build?

joelb123:
I just said something incorrect; there are indeed AVX and SSE2 ifdefs in the code, so you can't set the CFLAGS.

What you said seems to be correct as per my (novice) understanding of the CMake output.  If `-DNATIVE_ARCH=OFF`, I don't see any messages about recognition of SSE2 or AVX.  Maybe that's okay and the configuration still works, but one would have to do some checks to find out. 

Perhaps the best solution for now would be to put a native use flag that is default?

AndrewAmmerlaan:
> What you said seems to be correct as per my (novice) understanding of the CMake output. If `-DNATIVE_ARCH=OFF`, I don't see any messages about recognition of SSE2 or AVX. Maybe that's okay and the configuration still works, but one would have to do some checks to find out.

I'll see if I can investigate this and figure out which configuration produces which build.  Might take a while until I can take a proper look at it though because I'm rather busy with real-life work stuff at the moment

> Perhaps the best solution for now would be to put a native use flag that is default?

Yes, if the `-DHAVE_AVX2` etc options don't work as expected then we can compromise and use `-DNATIVE_ARCH=OFF` to toggle between generic and native build.

joelb123:
After a bit more digging, it seems that the CPUFLAGS stuff you put in does work properly, and should go back in.  But I'll have to test it to be sure.

A generic-only build is pretty worthless as the boost from SSE2/AVX is big.

AndrewAmmerlaan:
> After a bit more digging, it seems that the CPUFLAGS stuff you put in does work properly, and should go back in. But I'll have to test it to be sure.

You were right, the flag for avx2 works, but the sse2/sse4 flags aren't properly respected. I'll see if I can find a solution.

AndrewAmmerlaan:
Since the issue with the CPU flags is fixed I think we can close this. Please re-open this if it is still relevant.

PR #1144, opened 2022-02-13
app-office/visidata: Bump to 2.8

zpuskas:
- app-office/visidata: Bump to 2.8
- app-office/visidata: Drop old 1.5.2
- app-office/visidata: Drop 2.1.1


AndrewAmmerlaan:
Thanks :+1: 

It looks like we also need `dev-python/odfpy` to successfully run the tests, so I added that dependency in a follow-up commit.

PR #1148, opened 2022-06-13
dev-python/rasterio: new rev for gdal>=3.5.0

t0b3:
bump rev for compatibility with gdal>=3.5.0
drop unnecessary gdal[python]


AndrewAmmerlaan:
Thanks :+1:

PR #1155, opened 2022-07-12
dev-python/python-igraph-0.9.0: support python 3.10; require >=dev-libs/igraph-0.9.0

anton-molyboha:
Testing done: run all tests in the `tests` subdir of the upstream.

### DIFF
--- a/dev-python/python-igraph/python-igraph-0.9.0.ebuild
+++ b/dev-python/python-igraph/python-igraph-0.9.0.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{7..9} )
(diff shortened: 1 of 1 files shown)


dev-python/python-igraph/python-igraph-0.9.0.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{7..9} )
+PYTHON_COMPAT=( python3_{7..10} )
```
AndrewAmmerlaan, line 6: Please drop `python3_7`, it is obsolete. Did you also test it with `python3_11`? Whenever possible it is nice to add both 10 and 11 at the same time to save time later.

PR #1157, opened 2022-07-20
sci-physics/formcalc: new package, add 9.9, 9.10

APN-Pucky:

Both versions 9.9 and 9.10 since 9.10 is latest and sometimes upstream overwrites the latest version breaking the checksum.

Patch removes striping and existing binaries are removed.

`form` and `tform` (threaded form) are both replaced by `form` which can have a thread use flag (is there a better way to get the binary than `which`?).

Installation into the `/usr/share/Mathematica/Applications` folder.


sci-physics/formcalc/formcalc-9.10.ebuild
```diff
@@ -0,0 +1,53 @@
...
+DEPEND="${RDEPEND}"
+BDEPEND="
+	sci-mathematics/mathematica
+	sci-mathematics/form
+	"
+PATCHES=( "${FILESDIR}"/${PN}-compile.patch )
+
+src_compile() {
+	# remove shipped binaries
+	rm bin/Linux-x86-64/*
```
AndrewAmmerlaan, line 31: `|| die`

sci-physics/formcalc/formcalc-9.10.ebuild
```diff
@@ -0,0 +1,53 @@
...
+	rm bin/Linux-x86-64-kernel2.6/*
```
AndrewAmmerlaan, line 32: `|| die`

sci-physics/formcalc/formcalc-9.10.ebuild
```diff
@@ -0,0 +1,53 @@
...
+	export DEST=Linux-x86-64
+	./compile ${LDFLAGS}
```
AndrewAmmerlaan, line 34: `|| die`

If we are calling external commands we need to ensure the ebuild dies if there is a failure

sci-physics/formcalc/formcalc-9.10.ebuild
```diff
@@ -0,0 +1,53 @@
...
+}
+
+src_install() {
+	MMADIR=/usr/share/Mathematica/Applications
+	dosym ${MY_P} $MMADIR/${MY_PN}
+	dodir $MMADIR/${MY_P}
+	insinto $MMADIR
+	doins -r "${S}"
+	cd "${S}"
```
AndrewAmmerlaan, line 43: `|| die`

Though this whole line should be unnecessary because you should already be in S

sci-physics/formcalc/formcalc-9.10.ebuild
```diff
@@ -0,0 +1,53 @@
...
+	# remove shipped binaries
+	rm bin/Linux-x86-64/*
+	rm bin/Linux-x86-64-kernel2.6/*
+	export DEST=Linux-x86-64
+	./compile ${LDFLAGS}
+}
+
+src_install() {
+	MMADIR=/usr/share/Mathematica/Applications
+	dosym ${MY_P} $MMADIR/${MY_PN}
```
AndrewAmmerlaan, line 39:
```suggestion
	dosym ${MY_P} ${MMADIR}/${MY_PN}
```

sci-physics/formcalc/formcalc-9.10.ebuild
```diff
@@ -0,0 +1,53 @@
...
+	dodir $MMADIR/${MY_P}
```
AndrewAmmerlaan, line 40:
```suggestion
	dodir ${MMADIR}/${MY_P}
```

sci-physics/formcalc/formcalc-9.10.ebuild
```diff
@@ -0,0 +1,53 @@
...
+	insinto $MMADIR
```
AndrewAmmerlaan, line 41:
```suggestion
	insinto ${MMADIR}
```

AndrewAmmerlaan:
> `form` and `tform` (threaded form) are both replaced by `form` which can have a thread use flag (is there a better way to get the binary than `which`?).

There is! `command -v` is preferred over `which` because `command` is built into bash whereas `which` is an external command provided by `sys-apps/which` (this is in the system set so we can safely assume that it is available, that being said it is still better to use `command -v`)

APN-Pucky:
> > `form` and `tform` (threaded form) are both replaced by `form` which can have a thread use flag (is there a better way to get the binary than `which`?).
> 
> There is! `command -v` is preferred over `which` because `command` is built into bash whereas `which` is an external command provided by `sys-apps/which` (this is in the system set so we can safely assume that it is available, that being said it is still better to use `command -v`)

Does this also apply to find in the permission copy for loop?

AndrewAmmerlaan:
> > There is! command -v is preferred over which because command is built into bash whereas which is an external command provided by sys-apps/which (this is in the system set so we can safely assume that it is available, that being said it is still better to use command -v)
> 
> Does this also apply to find in the permission copy for loop?

AFAIK there is no built-in bash equivalent for `find`.

[EDIT] Oops I misclicked and accidentally edited your post instead of quoted it.

AndrewAmmerlaan:
Great work, Thanks :+1:

PR #1158, opened 2022-07-20
sci-physics/*: bump fastjet and add fjcontrib

APN-Pucky:

Adding fastjet-contrib package with a new use flag for fastjet[contrib] requires a change in the fastjet package.
So I bumped it to 3.4.0 as well.
We also enable the python extension of fastjet.

The gfortran patch adds `-lgfortran` to one of the plugins in `fastjet`, since the tests in `fastjet-contrib` complained about missing gfortran links in the library `libfastjetplugins.so`.


APN-Pucky:
Still had old manifests (1.046 and 1.048), updated for single version 1.049.

sci-physics/fastjet-contrib/fastjet-contrib-1.049.ebuild
```diff
@@ -0,0 +1,43 @@
...
+src_configure() {
+	./configure --prefix=/usr --fastjet-config=/usr/bin/fastjet-config CFLAGS="${CFLAGS}" CXXFLAGS="${CXXFLAGS}" FFLAGS="${FFLAGS}" LDFLAGS="${LDFLAGS}"
+}
+src_compile() {
+	emake
+	emake fragile-shared
+}
+
+src_install() {
+	emake install PREFIX="${D}/usr"
```
AndrewAmmerlaan, line 38:
```suggestion
	emake install PREFIX="${ED}/usr"
```
ED also takes the EPREFIX into account

sci-physics/fastjet/fastjet-3.4.0.ebuild
```diff
@@ -0,0 +1,67 @@
...
+KEYWORDS="~amd64"
+IUSE="cgal doc examples python +plugins +contrib"
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
+
+DEPEND="
+sys-devel/gcc:=[fortran]
+contrib? ( sci-physics/fastjet-contrib )
+cgal? ( sci-mathematics/cgal:=[shared(+)] )
+plugins? ( sci-physics/siscone:= )
+python? ( ${PYTHON_DEPS} )
```
AndrewAmmerlaan, line 26:
```suggestion
	sys-devel/gcc:=[fortran]
	contrib? ( sci-physics/fastjet-contrib )
	cgal? ( sci-mathematics/cgal:=[shared(+)] )
	plugins? ( sci-physics/siscone:= )
	python? ( ${PYTHON_DEPS} )
```
I'm not 100% sure, but I think we can depend on `virtual/fortran` instead of on `sys-devel/gcc[fortran]`.
In any case it is probably a BDEPEND and not a DEPEND. BDEPEND is for things that we execute when emerging the package, DEPEND is for things that have to be present but are not executed.

sci-physics/fastjet/fastjet-3.4.0.ebuild
```diff
@@ -0,0 +1,67 @@
...
+	"${FILESDIR}"/${P}-gfortran.patch 
+	)
+
+src_prepare() {
+	default
+	eautoreconf
+}
+
+src_configure() {
+	use cgal && has_version 'sci-mathematics/cgal[gmp]' && append-libs -lgmp
```
AndrewAmmerlaan, line 42: is `has_version` necessary here? We DEPEND on it with USE=cgal so we can safely assume that it is present

sci-physics/fastjet/fastjet-3.4.0.ebuild
```diff
@@ -0,0 +1,67 @@
...
+	econf $(use_enable cgal) $(use_enable plugins allplugins) $(use_enable plugins allcxxplugins) --enable-shared --enable-static=no --disable-static --disable-auto-ptr  $(use_enable python pyext)
```
AndrewAmmerlaan, line 43:
```suggestion
	econf \
		$(use_enable cgal) \
		 $(use_enable plugins allplugins) \
		 $(use_enable plugins allcxxplugins) \
		 --enable-shared \
		 --enable-static=no \
		 --disable-static \
		 --disable-auto-ptr  \
		 $(use_enable python pyext)
```

sci-physics/fastjet/fastjet-3.4.0.ebuild
```diff
@@ -0,0 +1,67 @@
...
+
+}
+
+src_compile() {
+	default
+
+	if use doc; then
+		doxygen Doxyfile || die
+		HTML_DOCS=( html/. )
+	fi
```
AndrewAmmerlaan, line 53: `docs.eclass` can help you simplify things here.

sci-physics/fastjet/metadata.xml
```diff
@@ -0,0 +1,17 @@
...
+		<name>Gentoo Physics Project</name>
+	</maintainer>
+	<longdescription lang="en">
+	A software package for jet finding in pp and e+e- collisions. It includes fast native implementations of many sequential recombination clustering algorithms, plugins for access to a range of cone jet finders and tools for advanced jet manipulation. 
+	</longdescription>
+	<use>
+		<flag name="python"> Build python bindings</flag>
+		<flag name="contrib"> Build fastjet-contrib (fjcontrib)</flag>
+		<flag name="plugins"> Build all standard and C++ plugins </flag>
+		<flag name="cgal"> Use <pkg>sci-mathematics/cgal</pkg> instead of bundled algorithms</flag>
```
AndrewAmmerlaan, line 15:
```suggestion
		<flag name="contrib">Build fastjet-contrib (fjcontrib)</flag>
		<flag name="plugins">Build all standard and C++ plugins </flag>
		<flag name="cgal">Use <pkg>sci-mathematics/cgal</pkg> instead of bundled algorithms</flag>
```
python is a global flag, so you don't need to define it locally here

sci-physics/fastjet-contrib/fastjet-contrib-1.049.ebuild
```diff
@@ -0,0 +1,43 @@
...
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64"
+
+DEPEND=">=sci-physics/fastjet-3.4.0"
+RDEPEND="${DEPEND}"
+
+PATCHES=(
+"${FILESDIR}"/${PN}-soname.patch
```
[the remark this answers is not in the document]
APN-Pucky: If I have a different version with the same patch then do I use the patch with the same versioned name for each of them in the ebuild then?
Because if I copy the same patch for each version pkgcheck complains about them being the same file.

sci-physics/fastjet-contrib/fastjet-contrib-1.049.ebuild
```diff
@@ -0,0 +1,43 @@
...
+)
+
+src_prepare() {
+	default
+}
+
+src_configure() {
+	./configure --prefix=/usr --fastjet-config=/usr/bin/fastjet-config CFLAGS="${CFLAGS}" CXXFLAGS="${CXXFLAGS}" FFLAGS="${FFLAGS}" LDFLAGS="${LDFLAGS}"
```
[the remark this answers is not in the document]
APN-Pucky: Yes, I think it is just a handwritten bash file called configure. It only allows for parameter "help, listfastjet-config, prefix, only, exclude" so `econf` failed.

sci-physics/fastjet-contrib/fastjet-contrib-1.049.ebuild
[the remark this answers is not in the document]
AndrewAmmerlaan: > If I have a multiple different version with the same patch do I use the patch with the same versioned name for each of them in the ebuild?

Yes, then you remove the reference to `PV` and instead specify the version where the patch was first introduced.

sci-physics/fastjet/fastjet-3.4.0.ebuild
APN-Pucky: I thought so too, but I decided to follow the gentoo-tree version: https://gitweb.gentoo.org/repo/gentoo.git/tree/sci-physics/fastjet/fastjet-3.0.6-r3.ebuild

AndrewAmmerlaan: > I thought so too, but I decided to follow the gentoo-tree version: https://gitweb.gentoo.org/repo/gentoo.git/tree/sci-physics/fastjet/fastjet-3.0.6-r3.ebuild

That an ebuild is in ::gentoo does not necessarily mean that it is a perfect example, there are also some ebuilds in there that are not quiet up to spec. That being said, I think that in this case it can't hurt to leave that check in there, at worst it does nothing useful.

APN-Pucky: I think cgal can have gmp enabled and disabled. In one case we need to add -lgmp.

AndrewAmmerlaan: > I think cgal can have gmp enabled and disabled. In one case we need to add `-lgmp`.

Probably, but that opens up a new can of worms. If I have cgal[gmp] installed when I emerge this ebuild, but then at some later time re-install cgal without the gmp flag, does fastjet break?
If yes then we need to have a USE flag to toggle a dependency on cgal[gmp] to ensure that the above cannot happen because portage will complain. If no then this is fine.

AndrewAmmerlaan: Possibly related bug report: https://bugs.gentoo.org/851912

AndrewAmmerlaan: > Possibly related bug report: https://bugs.gentoo.org/851912

The problem here seems to be that in version 5.5 of cgal the gmp USE flag has been dropped and it instead depends on `dev-libs/gmp` unconditionally. This means that with version 5.5 the `has_version` check will return False when it should return True.

We can do two things, either we need specific cases of this check for versions <5.5 of cgal and for versions >=5.5. Or we just depend on `sci-mathematics/cgal[gmp(+)]` and drop the check entirely since then we can assume that the gmp flag is always enabled on cgal.

APN-Pucky: unstable cgal-5.5 removed the gmp flag, but still depends on the lib. So it depends on if the cgal defaults are with `-DWITH_GMP=ON`...

AndrewAmmerlaan: CC'ing the author of cgal-5.5 @thesamesam 

I'm confused about this, it looks weird to me. And then there is https://bugs.gentoo.org/851912

sci-physics/fastjet-contrib/fastjet-contrib-1.049.ebuild
```diff
@@ -0,0 +1,43 @@
+# Copyright 2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+MY_PN=fjcontrib
+MY_P=${MY_PN}-${PV}
+
+DESCRIPTION="3rd party extensions of FastJet."
+HOMEPAGE="http://fastjet.hepforge.org/contrib/"
```
AndrewAmmerlaan, line 10: We like to use `https` links whenever possible

sci-physics/fastjet/fastjet-3.4.0.ebuild
```diff
@@ -0,0 +1,67 @@
...
+
+EAPI=8
+
+FORTRAN_NEEDED=plugins
+PYTHON_COMPAT=( python3_{8..10} )
+
+inherit autotools flag-o-matic fortran-2 python-r1
+
+DESCRIPTION="A software package for jet finding in pp and e+e- collisions"
+HOMEPAGE="http://fastjet.fr/"
```
AndrewAmmerlaan, line 12: And here too, we can update this to `https`

sci-physics/fastjet/fastjet-3.4.0.ebuild
APN-Pucky: I decided to depend on `sci-mathematics/cgal[gmp(-)]`, which filters cgal-5.5 away until it is resolved.

sci-physics/fastjet/fastjet-3.4.0.ebuild
```diff
@@ -0,0 +1,74 @@
...
+	"${FILESDIR}"/${P}-gfortran.patch
+)
+
+src_prepare() {
+	default
+	eautoreconf
+}
+
+src_configure() {
+	append-libs -lgmp
```
AndrewAmmerlaan, line 46: This should still be behind `use cgal` I think. It is only the `has_version` check that seems weird to me.

sci-physics/fastjet/fastjet-3.4.0.ebuild
```diff
@@ -28,7 +28,7 @@ DEPEND="
 "
 RDEPEND="${DEPEND}"
 BDEPEND="
-	app-doc/doxygen[dot]
+	doc? ( app-doc/doxygen[dot] )
```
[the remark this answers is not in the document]
APN-Pucky: To me the use of `app-doc/doxygen[dot]` seems better in case the doxygen ebuild changes. Also <https://gitweb.gentoo.org/repo/gentoo.git/tree/app-doc/doxygen/doxygen-1.9.4-r1.ebuild> lists `media-libs/freetype`.

`DOCS_DEPEND="app-doc/doxygen[dot]"` did not work, probably because it already contains doxygen.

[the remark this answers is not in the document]
AndrewAmmerlaan: > `DOCS_DEPEND="app-doc/doxygen[dot]"` did not work, probably because it already contains doxygen.

That is weird, this should also work. It pulls in doxygen but doesn't pull in graphviz?

> To me the use of `app-doc/doxygen[dot]` seems better in case the doxygen ebuild changes. Also https://gitweb.gentoo.org/repo/gentoo.git/tree/app-doc/doxygen/doxygen-1.9.4-r1.ebuild lists `media-libs/freetype` for `dot`.

Yes but usually having graphviz present is enough to get things working.

Note that the dot flag on doxygen doesn't really do anything. All it does is pull in graphviz and freetype, the installed files are identical (unless USE="doc" is also enabled, and even then USE="dot" is still irrelevant at doxygen runtime). For this reason the doxygen package probably shouldn't have the dot flag in the first place, strictly speaking. Which is why I usually avoid depending on it.

AndrewAmmerlaan:
I finally had some time to properly dig into https://bugs.gentoo.org/851912
The root of the problem is that cgal versions newer then 5.2 do *not* install the shared libraries that fastjet looks for in the configure phase, even if USE="shared". The only version of cgal that fastjet[cgal] compiles with is version 5.2, so I added that as the dependency in the fastjet ebuild. The cgal ebuild needs to be fixed to properly respect USE="shared" though.

I merged this to ::gentoo to avoid double work in maintaining this package both here and in ::gentoo. I changed some small things before merging that I'll explain in some follow-up comments.

sci-physics/fastjet/fastjet-3.4.0.ebuild
```diff
@@ -0,0 +1,75 @@
+# Copyright 2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+FORTRAN_NEEDED=plugins
+PYTHON_COMPAT=( python3_{8..10} )
```
AndrewAmmerlaan, line 7: I added `python3_11` to the compat list because it seems to compile fine with it.

sci-physics/fastjet/fastjet-3.4.0.ebuild
```diff
@@ -0,0 +1,75 @@
...
+
+LICENSE="GPL-2+"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="cgal doc examples python +plugins +contrib"
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
+
+# For now, we depend on cgal[gmp(-)] and append -lgmp to libs, despite possible incompatibility with cgal-5.5
+DEPEND="
+	contrib? ( sci-physics/fastjet-contrib )
```
AndrewAmmerlaan, line 26: The `contrib` use flag doesn't really do anything apart from pulling in fastjet-contrib. We usually avoid introducing use flags for these kind of situations because toggling this flag will then result in a useless recompile that doesn't changed the installed files. Additionally the flag causes a circular dependency so I dropped it.

sci-physics/fastjet/fastjet-3.4.0.ebuild
```diff
@@ -0,0 +1,75 @@
...
+	default
+	eautoreconf
+}
+
+src_configure() {
+	use cgal && append-libs -lgmp
+	econf \
+		$(use_enable cgal) \
+		$(use_enable plugins allplugins) \
+		$(use_enable plugins allcxxplugins) \
```
AndrewAmmerlaan, line 51: There is a small problem here. Because it seems that (some of) the plugins are being built anyway even with USE="-plugins". With the plugins flag disabled I get the following QA notice:
```
* QA Notice: Files built without respecting CFLAGS have been detected
*  Please include the following list of files in your report:
* /usr/lib64/libfastjetplugins.so.0.0.0

strip: x86_64-pc-linux-gnu-strip --strip-unneeded -N __gentoo_check_ldflags__ -R .comment -R .GCC.command.line-R .note.gnu.gold-version
/usr/lib64/libfastjetplugins.so.0.0.0
/usr/lib64/libfastjet.so.0.0.0
/usr/lib64/libfastjettools.so.0.0.0
```
This problem is also present in the in-tree version 3.0.6-r3 so it does not prevent merging this. Still it would be nice to fix this at some point.

sci-physics/fastjet/fastjet-3.4.0.ebuild
```diff
@@ -0,0 +1,75 @@
...
+
+EAPI=8
+
+FORTRAN_NEEDED=plugins
+PYTHON_COMPAT=( python3_{8..10} )
+DOCS_BUILDER="doxygen"
+# we depend on these similar to putting doxygen[dot] in BDEPEND
+DOCS_DEPEND="media-gfx/graphviz media-libs/freetype"
+
+inherit autotools flag-o-matic fortran-2 python-r1 docs
```
AndrewAmmerlaan, line 12: When we inherit `python-r1` we imply that we want to build the python bindings for multiple enabled python versions (PYTHON_TARGETS). However, we do not do this in the ebuild below (we do not use the `python_foreach_impl` function), this would result in only building the python bindings for the highest enabled PYTHON_TARGET which is not what we want to do when we use `python-r1`. The easy solution (rather then using `python_foreach_impl` in prepare/configure/compile/install) is to just use `python-single-r1` instead, which is what I changed before merging. If we want we can always enable support for multiple simultaneous python versions at some later time, but I don't think it is necessary for this package.

AndrewAmmerlaan:
Great work here, Thanks :+1:

APN-Pucky:
Thanks for fixing the details. I have some more changes to `sci-physics`-packages that are in ::gentoo (eg. `sci-physics/hepmc`, `sci-physics/yoda`, ...). Should I file PRs first to ::science or directly to ::gentoo?

AndrewAmmerlaan:
> Thanks for fixing the details. I have some more changes to `sci-physics`-packages that are in ::gentoo (eg. `sci-physics/hepmc`, `sci-physics/yoda`, ...). Should I file PRs first to ::science or directly to ::gentoo?

Please open those directly to ::gentoo. If you want you're also welcome to [proxy-maintain](https://wiki.gentoo.org/wiki/Project:Proxy_Maintainers) some of those packages.

Our automated ebuild testing tool (Tinderbox) found some small QA issues with fastjet and fastjet-contrib. If you feel like it you can also dig into those, I've put you on the CC list for those bug reports.

For the next two weeks I'll be away and mostly unavailable so it might take a bit longer then usual to process sci related Pull Requests.

APN-Pucky:
I am considering the proxy-maintaining, however I am not sure how much time I will have spare in the future.

Let's hope these are the final changes.
https://github.com/gentoo/gentoo/pull/26520
https://github.com/gentoo/gentoo/pull/26521

PR #1161, opened 2022-07-22
sci-physics/spheno: Fix toolchain gfortran

APN-Pucky:
We should not call gfortran directly.
Instead use `FC` provided by toolchain.


AndrewAmmerlaan:
Thanks :+1:

PR #1168, opened 2022-08-25
sci-biology/open-ephys-gui: add 0.6.1

malfatti:
Hi!

This will bump `sci-biology/open-ephys-gui` to version `0.6.1`. Unlike the previous version, now `EAPI=8` works :)

Some comments:
1. `pkgcheck` complains about duplicated files, but that is because I used an `if` statement to maintain a single ebuild for all versions;
2. `pkgcheck` complains about duplicated `inherit 'git-r3'`, but I need it for the same reason above;
3. The `BUILD_DIR="${S}/Build"` override is still needed, the path is hardcoded in the source;
4. It is still necessary to patch the `CMakeLists.txt` to comment out the fatal error conditional statement due to "invalid build type";
5. I could not follow the contributing guidelines completely. When pushing to my fork I got errors:
```bash
$ git pull --rebase=merges upstream master
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.


$ pkgdev push malfatti 'sci-biology/open-ephys-gui'
sci-biology/open-ephys-gui
  DuplicateFiles: duplicate identical files in FILESDIR: 'files/open-ephys-gui-0.5.5.patch', 'files/open-ephys-gui-9999.patch', 'files/open-ephys-gui-99999999.patch'
  DuplicateEclassInherit: version 9999: duplicate eclass inherit 'git-r3', line 18
  DuplicateEclassInherit: version 99999999: duplicate eclass inherit 'git-r3', line 18
fatal: 'malfatti' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```
then instead I just
```bash
$ git push --set-upstream origin sci-biology/open-ephys-gui
[...]
```


sci-biology/open-ephys-gui/open-ephys-gui-0.6.1.ebuild
```diff
@@ -0,0 +1,83 @@
...
+}
+
+src_configure() {
+	local mycmakeargs=( -DCMAKE_SKIP_RPATH=ON )
+	cmake_src_configure
+}
+
+src_install() {
+	dodir opt/open-ephys-"${Suffix}"/ lib/udev/rules.d/
+	cp -R "${BUILD_DIR}"/RelWithDebInfo/* "${ED}"/opt/open-ephys-"${Suffix}"/
```
AndrewAmmerlaan, line 72: `|| die`

AndrewAmmerlaan:
> 1. `pkgcheck` complains about duplicated files, but that is because I used an `if` statement to maintain a single ebuild for all versions;

The complaint is about the duplicate patch file in FILESDIR. Usually we adjust the PATCHES variable to reference `${PN}-xx` instead of `${P}` to avoid duplicating the patch files.

sci-biology/open-ephys-gui/open-ephys-gui-0.6.1.ebuild
malfatti: Done :)

malfatti:
Thanks @AndrewAmmerlaan! I made the following changes:

1. Added `|| die` as suggested;
2. Added the Science Project as an additional maintainer (sorry I missed that before);
3. Removed the live patches and reference the last patch version in the live ebuilds;
4. Update the live ebuilds to match the current version.

sci-biology/open-ephys-gui/open-ephys-gui-9999.ebuild
```diff
@@ -65,8 +68,8 @@ src_configure() {
 }
 
 src_install() {
-	dodir opt/open-ephys-"${Suffix}"/ lib/udev/rules.d/
-	cp -R "${BUILD_DIR}"/Gentoo/* "${ED}"/opt/open-ephys-"${Suffix}"/
+	dodir opt/open-ephys-"${Suffix}"/ lib/udev/rules.d/ || die
```
AndrewAmmerlaan, line 71: `dodir` is a built-in function, so it doesn't need the `|| die`. We only need that for external commands. Not that it hurts to have it here.

AndrewAmmerlaan (changes requested):
With the patches deleted the 9999 and 999999 ebuilds will fail to find it because they still reference `${P}`. In e.g. the 9999 ebuild P will translate to `open-ephys-gui-9999` which now doesn't exist. We still need to change:
`PATCHES=( "${FILESDIR}"/${P}.patch )` to `PATCHES=( "${FILESDIR}"/${PN}-0.6.1.patch )`
so those ebuilds can also find the `open-ephys-gui-0.6.1.patch`

sci-biology/open-ephys-gui/open-ephys-gui-9999.ebuild
malfatti: Ah, got it! fixed.

malfatti:
D'oh!  I updated the 99999999 but forgot the 9999 version :) All fixed now, both live ebuilds pointing to the last patch, and `|| die` removed from `dodir` lines.

malfatti:
Now I noticed the branch name `master` was changed to `main` upstream, just fixed the ebuilds accordingly.

AndrewAmmerlaan:
Looks good, one final thing: Could you please add a sign-off line to your commit message, in accordance with [Gentoo's copyright policy](https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin). Sorry, I hadn't noticed before that it was missing.

malfatti:
Done :)

AndrewAmmerlaan:
Great work, Thanks :+1:

PR #1169, opened 2022-08-31
sci-physics/formcalc: update Manifest

APN-Pucky:
This update is due to formcalc's unfortunate versioning:
The latest version is continuously updated and there is no public vcs version (similar to eg. looptools).

I am not sure if it is possible to somehow use the 9.10 version as 9999 latest upstream version, since at some point it will be replaced by 9.11.


AndrewAmmerlaan:
Thanks :+1: 

Probably the best long-term fix is to host a tarball somewhere else and fetch it from there.

> I am not sure if it is possible to somehow use the 9.10 version as 9999 latest upstream version, since at some point it will be replaced by 9.11.

Unfortunately this won't work because everything in `SRC_URI` will be in the Manifest, even if it is a 9999 ebuild.

PR #1170, opened 2022-09-02
sci-physics/rivet: Use ESYSROOT path

APN-Pucky:
I think SYSROOT is better here for eg. gentoo prefix.
Or should it be ESYSROOT?

### DIFF
--- a/sci-physics/rivet/rivet-3.1.5.ebuild
+++ b/sci-physics/rivet/rivet-3.1.5.ebuild
@@ -62,8 +62,8 @@ src_configure() {
 	PREFIX_YODA=$(yoda-config --prefix) || die
 	PREFIX_FJ=$(fastjet-config --prefix) || die
 	econf \
-		$(usex hepmc2 "--with-hepmc=/usr" "") \
-		$(usex hepmc3 "--with-hepmc3=/usr" "") \
+		$(usex hepmc2 "--with-hepmc=${SYSROOT}/usr" "") \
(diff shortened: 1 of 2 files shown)


sci-physics/rivet/rivet-3.1.5.ebuild
```diff
@@ -62,8 +62,8 @@ src_configure() {
 	PREFIX_YODA=$(yoda-config --prefix) || die
 	PREFIX_FJ=$(fastjet-config --prefix) || die
 	econf \
-		$(usex hepmc2 "--with-hepmc=/usr" "") \
-		$(usex hepmc3 "--with-hepmc3=/usr" "") \
+		$(usex hepmc2 "--with-hepmc=${SYSROOT}/usr" "") \
```
thesamesam, line 65: Yeah, indeed, ESYSROOT.

sci-physics/rivet/rivet-3.1.5.ebuild
```diff
@@ -62,8 +62,8 @@ src_configure() {
...
+		$(usex hepmc3 "--with-hepmc3=${SYSROOT}/usr" "") \
 		--with-yoda=$PREFIX_YODA \
```
thesamesam, line 67: We use ${PREFIX_YODA} style in Gentoo: https://projects.gentoo.org/qa/policy-guide/ebuild-format.html#pg0101

AndrewAmmerlaan:
Thanks :+1:

PR #1171, opened 2022-09-26
net-misc/cernbox-client: add 2.11.1.8438, drop oldest version

olifre:
This also re-synchronizes with `net-misc/owncloud-client::gentoo` and updates to the latest theming package from CERN (also mirrored for ensured availability).


AndrewAmmerlaan:
Thanks :+1:

PR #1176, opened 2022-10-25
sci-physics/oneloop: new package, add 2020.07.31

APN-Pucky:


sci-physics/oneloop/oneloop-2020.07.31.ebuild
```diff
@@ -0,0 +1,49 @@
...
+	sed -i "/FC = /s/gfortran/${FC}/g" Config || die
+	sed -i "/FFLAGS = /s/ -O/${FFLAGS} -fPIC/g" Config || die
+	emake -f make_cuttools
+}
+
+src_compile() {
+	tc-export FC
+	./create.py || die
+	#./create.py dynamic || die
+	${FC} ${LDFLAGS} -Wl,-soname,libavh_olo.so -shared -o libavh_olo.so avh_olo.o
```
AndrewAmmerlaan, line 42: `|| die "Failed to compile"

sci-physics/oneloop/oneloop-2020.07.31.ebuild
```diff
@@ -0,0 +1,49 @@
...
+src_configure() {
+	tc-export FC
+	sed -i "/FC = /s/gfortran/${FC}/g" Config || die
+	sed -i "/FFLAGS = /s/ -O/${FFLAGS} -fPIC/g" Config || die
+	emake -f make_cuttools
+}
+
+src_compile() {
+	tc-export FC
+	./create.py || die
```
AndrewAmmerlaan, line 40: This should be `${EPYTHON} -m create.py` I think. Otherwise the script would use the system default python instead of the python version specified in PYTHON_TARGETS

sci-physics/oneloop/oneloop-2020.07.31.ebuild
```diff
@@ -0,0 +1,49 @@
...
+
+PATCHES=(
+	"${FILESDIR}"/${P}-config.patch
+)
+
+src_configure() {
+	tc-export FC
+	sed -i "/FC = /s/gfortran/${FC}/g" Config || die
+	sed -i "/FFLAGS = /s/ -O/${FFLAGS} -fPIC/g" Config || die
+	emake -f make_cuttools
```
AndrewAmmerlaan, line 35: It feels a bit weird to call `emake` in the configure phase, can we move it to the compile phase instead?

sci-physics/oneloop/oneloop-2020.07.31.ebuild
APN-Pucky: I used ${EPYTHON} because I think `-m` is only for modules.

AndrewAmmerlaan: > I used `${EPYTHON}` because I think `-m` is only for modules.

Yes you are right, my bad

sci-physics/oneloop/oneloop-2020.07.31.ebuild
AndrewAmmerlaan: We still need to add a `die` here since `$FC` is an external command.

APN-Pucky: Ah, I got confused it with the `create.py || die`.

AndrewAmmerlaan:
Nice work, Thanks :+1:

PR #1178, opened 2022-10-28
numba-0.56.3 dependency on <numpy-1.24

FoscoV:
According to https://numba.readthedocs.io/en/stable/user/installing.html#version-support-information the dependency changed since 0.56.2


AndrewAmmerlaan:
Thanks :+1:

PR #1179, opened 2022-10-28
xgboost-1.3.1 for python3_10

FoscoV:
this xgboost version already supports python3.10 no need to update just for pthon_target issues


AndrewAmmerlaan:
Thanks :+1:

PR #1180, opened 2022-10-31
sci-mathematics/netgen: add 6.2.2204

waebbl:
- OpenMP doesn't seem to be used, remove USE flag
- dev-tcltk/tix doesn't seem to be used, remove dependency
- dev-tcltk/togl is build internal and doesn't use a system togl installation
- Stub generation from pybind11 doesn't currently work in a sandbox, see https://github.com/NGSolve/netgen/issues/132
- only depend on sci-libs/metis for USE=mpi

Signed-off-by: Bernd Waibel <waebbl-gentoo@posteo.net>


waebbl:
How about moving this package to different category, like media-gfx? IMO it's not really a maths package. Also the desktop file we provide gives it a category of Graphics, instead of Education, Science, Maths.

waebbl:
Actually, the libtogl.a library is referenced in the netgen-targets.cmake file. So we have to leave it for now.

sci-mathematics/netgen/netgen-6.2.2204.ebuild
```diff
@@ -0,0 +1,137 @@
...
+	fi
+	cmake_src_configure
+}
+
+src_install() {
+	cmake_src_install
+	use python && python_optimize
+
+	local NETGENDIR="/usr/share/${PN}"
+	echo -e "NETGENDIR=${NETGENDIR}" > ./99netgen
```
AndrewAmmerlaan, line 125: `|| die`

waebbl: Didn't change this line ;) but yeah, you're right.

waebbl:
> Yeah I think we should move it.

Then I move it to media-gfx on my next push with your permission.

AndrewAmmerlaan:
Thanks :+1:

waebbl:
Thanks for merging @AndrewAmmerlaan 

I'd like to discuss the possibility of getting the package into ::gentoo. In my overlay, I tested the support in FreeCAD for this version and it builds fine. I'd be happy, if we could add this support to the FreeCAD versions in tree as well.

WDYT?

AndrewAmmerlaan:
> Thanks for merging @AndrewAmmerlaan
> 
> I'd like to discuss the possibility of getting the package into ::gentoo. In my overlay, I tested the support in FreeCAD for this version and it builds fine. I'd be happy, if we could add this support to the FreeCAD versions in tree as well.
> 
> WDYT?

Yeah since this is an optional dependency of freecad I think it is a good candidate for moving to ::gentoo. The ebuild looks fine, but there is a test failure with `USE="python"` that we should fix. It looks like the failure occurs because the package is not yet installed (it can't find its own modules), setting an appropriate `PYTHONPATH` should fix it.

waebbl:
> > Thanks for merging @AndrewAmmerlaan
> > I'd like to discuss the possibility of getting the package into ::gentoo. In my overlay, I tested the support in FreeCAD for this version and it builds fine. I'd be happy, if we could add this support to the FreeCAD versions in tree as well.
> > WDYT?
> 
> Yeah since this is an optional dependency of freecad I think it is a good candidate for moving to ::gentoo. The ebuild looks fine, but there is a test failure with `USE="python"` that we should fix. It looks like the failure occurs because the package is not yet installed (it can't find its own modules), setting an appropriate `PYTHONPATH` should fix it.

I look into the test failure. I also want to polish the ebuild some more and check adding support for `dev-libs/spdlog`

PR #1184, opened 2022-12-09
dev-python/geopandas: version bump and opitional dependencies listing

camaradadennis:


AndrewAmmerlaan:
Nice work, Thanks :+1:

PR #1186, opened 2022-12-15
media-gfx/netgen - add tests and enable logging, drop 6.2.2105

waebbl:
- drop 6.2.2105
- enable logging, using dev-libs/spdlog
- enable tests, disable some tests which do not work

Signed-off-by: Bernd Waibel <waebbl-gentoo@posteo.net>

As discussed in https://github.com/gentoo/sci/pull/1180 here's a patch to look into an issue with testing the package and enable logging.

I couldn't get the python tests working. The tests seem to expect an installed version. Adding PYTHONPATH to the environment before testing didn't solve the issues. If someone has an idea on how to get this working, please share it.


media-gfx/netgen/metadata.xml
```diff
@@ -5,8 +5,17 @@
 		<email>sci-mathematics@gentoo.org</email>
 		<name>Gentoo Mathematics Project</name>
 	</maintainer>
+	<maintainer type="person" proxied="yes">
+		<email>waebbl-gentoo@posteo.net</email>
+		<name>Bernd Waibel</name>
+	</maintainer>
+	<maintainer type="project">
+		<email>proxy-maint@gentoo.org</email>
```
AndrewAmmerlaan, line 13: Please drop proxy-maint here, they don't maintain packages in the science overlay.

waebbl: Adding proxy-maint and myself was in preparation for the move to ::gentoo.
Still remove it?

AndrewAmmerlaan: > Adding proxy-maint and myself was in preparation for the move to ::gentoo. Still remove it?

For adding it here we're going to have to remove it. When it is moved we can re-add it.

AndrewAmmerlaan:
Nice work, Thanks :+1:

PR #1187, opened 2022-12-28
dev-python/archspec: new package, add 0.1.4

APN-Pucky:
Signed-off-by: Alexander Puck Neuwirth <alexander@neuwirth-informatik.de>

This is a requirement for a the following EasyBuild PR.

PR #1188, opened 2022-12-28
Add Easybuild

APN-Pucky:
Requires (already included here):
- [x] https://github.com/gentoo/sci/pull/1187


sys-cluster/easybuild-easyblocks/easybuild-easyblocks-4.6.2.ebuild
```diff
@@ -0,0 +1,29 @@
...
+KEYWORDS="~amd64"
+
+RDEPEND="
+	>=sys-cluster/easybuild-framework-${SUB_P}[${PYTHON_USEDEP}]
+"
+BDEPEND="${RDEPEND}"
+
+src_prepare() {
+	# already there from easybuild_framework
+	rm easybuild/__init__.py
```
AndrewAmmerlaan, line 27: `|| die`

Though this looks a bit weird, if this is the case wouldn't there also be a file conflict if the package is installed with pip? Or does pip not care about that?

sys-cluster/easybuild/.build
```diff
@@ -0,0 +1,7 @@
+go
+a
+b
+c
```
AndrewAmmerlaan, line 4: This looks like a leftover file that we don't need in the repository.

sys-cluster/easybuild/easybuild-4.6.2.ebuild
```diff
@@ -0,0 +1,35 @@
...
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="+lmod modules"
+REQUIRED_USE=" ^^ ( lmod modules ) "
+
+RDEPEND="
+	dev-libs/openssl
+	dev-tcltk/tclx
+	lmod? ( >=sys-cluster/lmod-6.5.1 )
+	modules? ( >=sys-cluster/modules-3.2.10 )
```
AndrewAmmerlaan, line 24: If these are optional runtime dependencies only, perhaps we can use `optfeature` instead?

sys-cluster/easybuild/.build
APN-Pucky: Whoops, thanks. I have a second local .gitignore now masking my remnant .build files.

sys-cluster/easybuild/easybuild-4.6.2.ebuild
APN-Pucky: They are not optional, since easybuild needs one module tool -> `REQUIRED_USE` (iirc modules and lmod can not both be installed).
https://docs.easybuild.io/en/latest/Installation.html#requirements

sys-cluster/easybuild-easyblocks/easybuild-easyblocks-4.6.2.ebuild
APN-Pucky: Yes, pip does not care:
```
~/.local/lib/python3.9/site-packages $ ls
base  easyblocks  framework  __init__.py  main.py  __pycache__  toolchains  tools
```

sys-cluster/easybuild/easybuild-4.6.2.ebuild
AndrewAmmerlaan: Why can't we install both at the same time? If there is a problem with this then we shouldn't put this condition here, but add it to the `sys-cluster/modules` ebuild as a runtime blocker instead. 

I think here we want a `RDEPEND="|| ( sys-cluster/lmod sys-cluster/modules )"`

APN-Pucky: I agree, they use the same named binary `module` (I guess to make switching the module tool easier):

* https://lmod.readthedocs.io/en/latest/045_transition.html#purge-modules-using-old-module-command
*  https://modules.sourceforge.net/#Quick_examples

APN-Pucky: I created a revbump for modules with lmod strong blocked https://github.com/gentoo/sci/pull/1189

Then I'd adjust this PR'S modules version to include the revbump.

AndrewAmmerlaan: > I created a revbump for modules with lmod strong blocked #1189.
> 
> Then I'd adjust this PR's modules version to include the revbump.

Great, Thanks

I still think we don't need the USE flags to select either lmod or modules, an `|| ( )` runtime dependency should be sufficient since we can switch between these two without having to recompile the package, correct?

APN-Pucky: Yes, that makes more sense. I guess the first dependency in `|| ( )` is preferred?

AndrewAmmerlaan: > Yes, that makes more sense. I guess the first dependency in `|| ( )` is preferred?

Yes, it is.

APN-Pucky: Done.

AndrewAmmerlaan (commented):
Looks good, there just is a small problem with the installation paths:

```
* Messages for package sys-cluster/easybuild-framework-4.6.2:
* Log file: /var/log/portage/build/sys-cluster/easybuild-framework-4.6.2:20230106-104354.log

* QA Notice: The ebuild is installing to one or more unexpected paths:
*
*   /usr/contrib
*   /usr/easybuild
*   /usr/etc
*
* Please fix the ebuild to use correct FHS/Gentoo policy paths.
*
* QA Notice: This package installs one or more Python modules that are
* not byte-compiled.
* The following files are missing:
*
*   /usr/lib/python3.10/site-packages/easybuild/tools/py2vs3/__pycache__/py2.cpython-310.opt-1.pyc
*   /usr/lib/python3.10/site-packages/easybuild/tools/py2vs3/__pycache__/py2.cpython-310.opt-2.pyc
*   /usr/lib/python3.10/site-packages/easybuild/tools/py2vs3/__pycache__/py2.cpython-310.pyc
*   /usr/lib/python3.11/site-packages/easybuild/tools/py2vs3/__pycache__/py2.cpython-311.opt-1.pyc
*   /usr/lib/python3.11/site-packages/easybuild/tools/py2vs3/__pycache__/py2.cpython-311.opt-2.pyc
*   /usr/lib/python3.11/site-packages/easybuild/tools/py2vs3/__pycache__/py2.cpython-311.pyc
*
* For more information on bytecode files and related issues, please see:
*   https://projects.gentoo.org/python/guide/qawarn.html#compiled-bytecode-related-warnings
*
* QA Notice: This package seems to contain tests but they are not enabled.
* Please either run tests (via distutils_enable_tests or declaring
* python_test yourself), or add RESTRICT="test" along with an explanatory
* comment if tests cannot be run.
*

* Messages for package sys-cluster/easybuild-easyconfigs-4.6.2:
* Log file: /var/log/portage/build/sys-cluster/easybuild-easyconfigs-4.6.2:20230106-104424.log

* QA Notice: The ebuild is installing to one or more unexpected paths:
*
*   /usr/easybuild
*
* Please fix the ebuild to use correct FHS/Gentoo policy paths.
```

Would it be possible to move these files to `/etc` and `/usr/share` accordingly? The byte-compile warning you can fix by adding `python_optimize` to the install phase.

APN-Pucky:
These files are installed from `setup.py` `data_files`, apparently into the prefix `/usr`.

They are found either in `os.path.dirname(os.path.dirname(os.path.dirname(__file__)))` or `os.path.dirname(os.path.dirname(which('eb')))`, where `eb` is `/usr/local/bin`.
I could do quite some patching on the finding of these files, but I'd rather keep the sourcecode as it is.

AndrewAmmerlaan:
> These files are installed from `setup.py` `data_files`, apparently into the prefix `/usr`.
> 
> They are found either in `os.path.dirname(os.path.dirname(os.path.dirname(__file__)))` or `os.path.dirname(os.path.dirname(which('eb')))`, where `eb` is `/usr/bin/eb`. I could do quite some patching on the finding of these files, but I'd rather keep the sourcecode as it is.

Alright, lets keep it as it is now than. Thanks :+1:

APN-Pucky:
> Thanks. I am running my tests currently with the `python_optimize` option added.

From https://projects.gentoo.org/python/guide/qawarn.html#compiled-bytecode-related-warnings it should not be needed:
```
An acceptable workaround is to call python_optimize from the ebuild (note that in PEP517 mode, distutils-r1 does that unconditionally).
```

Adding it fails:
```
2023-01-06 11:58:34,609 gpep517 INFO The backend produced /var/tmp/portage/sys-cluster/easybuild-framework-4.6.2/work/easybuild-framework-4.6.2-python3_11/wheel/easybuild_framework-4.6.2-py3-none-any.whl
 *   Installing easybuild_framework-4.6.2-py3-none-any.whl to /var/tmp/portage/sys-cluster/easybuild-framework-4.6.2/work/easybuild-framework-4.6.2-python3_11/install
gpep517 install-wheel --destdir=/var/tmp/portage/sys-cluster/easybuild-framework-4.6.2/work/easybuild-framework-4.6.2-python3_11/install --interpreter=/usr/bin/python3.11 --prefix=/usr --optimize=all /var/tmp/portage/sys-cluster/easybuild-framework-4.6.2/work/easybuild-framework-4.6.2-python3_11/wheel/easybuild_framework-4.6.2-py3-none-any.whl
2023-01-06 11:58:34,724 gpep517 INFO Installing /var/tmp/portage/sys-cluster/easybuild-framework-4.6.2/work/easybuild-framework-4.6.2-python3_11/wheel/easybuild_framework-4.6.2-py3-none-any.whl into /var/tmp/portage/sys-cluster/easybuild-framework-4.6.2/work/easybuild-framework-4.6.2-python3_11/install
2023-01-06 11:58:35,651 gpep517 INFO Installation complete
*** Error compiling '/var/tmp/portage/sys-cluster/easybuild-framework-4.6.2/work/easybuild-framework-4.6.2-python3_11/install/usr/lib/python3.11/site-packages/easybuild/tools/py2vs3/py2.py'...
  File "/usr/lib/python3.11/site-packages/easybuild/tools/py2vs3/py2.py", line 82
    raise exception_class, message, traceback  # noqa: E999
                         ^
SyntaxError: invalid syntax

*** Error compiling '/var/tmp/portage/sys-cluster/easybuild-framework-4.6.2/work/easybuild-framework-4.6.2-python3_11/install/usr/lib/python3.11/site-packages/easybuild/tools/py2vs3/py2.py'...
  File "/usr/lib/python3.11/site-packages/easybuild/tools/py2vs3/py2.py", line 82
    raise exception_class, message, traceback  # noqa: E999
                         ^
SyntaxError: invalid syntax

*** Error compiling '/var/tmp/portage/sys-cluster/easybuild-framework-4.6.2/work/easybuild-framework-4.6.2-python3_11/install/usr/lib/python3.11/site-packages/easybuild/tools/py2vs3/py2.py'...
  File "/usr/lib/python3.11/site-packages/easybuild/tools/py2vs3/py2.py", line 82
    raise exception_class, message, traceback  # noqa: E999
                         ^
SyntaxError: invalid syntax

>>> Source compiled.
Unable to unshare: EPERM (for FEATURES="ipc-sandbox network-sandbox pid-sandbox")
>>> Test phase [not enabled]: sys-cluster/easybuild-framework-4.6.2
Unable to unshare: EPERM (for FEATURES="ipc-sandbox network-sandbox pid-sandbox")

>>> Install sys-cluster/easybuild-framework-4.6.2 into /var/tmp/portage/sys-cluster/easybuild-framework-4.6.2/image
 * ERROR: sys-cluster/easybuild-framework-4.6.2::hep-forge failed (install phase):
 *   No Python implementation set (EPYTHON is null).
 * Call stack:
 *     ebuild.sh, line  136:  Called src_install
 *   environment, line 3478:  Called python_optimize
 *   environment, line 3342:  Called die
 *       [[ -n ${EPYTHON} ]] || die 'No Python implementation set (EPYTHON is null).';
 * 
 * If you need support, post the output of `emerge --info '=sys-cluster/easybuild-framework-4.6.2::hep-forge'`,
 * the complete build log and the output of `emerge -pqv '=sys-cluster/easybuild-framework-4.6.2::hep-forge'`.
 * The complete build log is located at '/var/tmp/portage/sys-cluster/easybuild-framework-4.6.2/temp/build.log'.
 * The ebuild environment file is located at '/var/tmp/portage/sys-cluster/easybuild-framework-4.6.2/temp/environment'.
 * Working directory: '/var/tmp/portage/sys-cluster/easybuild-framework-4.6.2/work/easybuild-framework-4.6.2'
 * S: '/var/tmp/portage/sys-cluster/easybuild-framework-4.6.2/work/easybuild-framework-4.6.2'

>>> Failed to emerge sys-cluster/easybuild-framework-4.6.2, Log file:
```
They seem to ignore some flake8 qa failures there.

AndrewAmmerlaan:
> > Thanks. I am running my tests currently with the `python_optimize` option added.
> 
> From https://projects.gentoo.org/python/guide/qawarn.html#compiled-bytecode-related-warnings it should not be needed:
> 
> ```
> An acceptable workaround is to call python_optimize from the ebuild (note that in PEP517 mode, distutils-r1 does that unconditionally).
> ```
> 
> Adding it fails:

Looks like the problem is that there is a python2 file in the source code. This will fail to byte-compile with python3. Probably we can just remove the `/usr/lib/python3.X/site-packages/easybuild/tools/py2vs3/` submodule, I doubt it will work since it is python2 code. Or we can just ignore the warning, it is not super important anyway.

PR #1189, opened 2022-12-30
sys-cluster/modules: strong blocked lmod

APN-Pucky:

Both packages install binary `module`.


AndrewAmmerlaan:
Nice work, Thanks :+1:

PR #1190, opened 2023-02-06
dev-python/geopandas: add 0.12.2, drop 0.12.1

camaradadennis:
geopandas 0.12.2 released on Dec 10, 2022.


AndrewAmmerlaan:
Thanks :+1:

PR #1191, opened 2023-02-07
dev-python/fiona: add 1.9.0, drop 1.8.21

camaradadennis:
fiona-1.9.0 released on Jan 30, 2023. Dropped dev-python/six as a dependency.


AndrewAmmerlaan:
Thanks :+1:

PR #1192, opened 2023-03-20
Import apach-arrow from spark overlay

littlewu2508:
These commits pulls `dev-libs/apache-arrow`, `dev-python/pyarrow` and their depedency from [spark overlay](https://github.com/6-6-6/spark-overlay)

They are initially introduced into spark overlay because they are runtime dependencies of pyspark, responsible for parquet data format.

Nowadays we found the parquet data format is suitable for generic scientific computing, and pyspark is not the only use case, so I picked up these ebuilds,  bumped its version, did some QA enhancements, and created this PR at ::science.

There are many optional features apache-arrow provided, and I do not have time to turn all the options into use flags. So I just add some useful ones for me (mainly for parquet IO). Also, there are many programming languages supported, these two ebuilds provides support for c++ and python. So, further contributions are always welcome.


AndrewAmmerlaan:
Thanks :+1:

PR #1193, opened 2023-04-05
sci-biology/open-ephys-gui: version bump

malfatti:
Hi!

The newest release of this software required only a dependency version update. The patch to disable the `FATAL_ERROR` is still needed. `ExcessiveLineLength` warnings are due to the `sed` line, and `DuplicateEclassInherit` warnings are because the live ebuilds have the git-r3 eclass inherited conditionally.


AndrewAmmerlaan:
Looks good,

There is no need to copy the patch file if the old version still applies. In fact, we prefer not to because there is no point in having multiple identical files in the repository. We can just change `PATCHES=( "${FILESDIR}"/${P}.patch )` to `PATCHES=( "${FILESDIR}"/${PN}-0.6.1.patch )` in the 0.6.4 ebuild and re-use the old patch. This way we also don't have to touch the live ebuilds.

Also, could you please sign-off your commit in accordance to [Gentoo's copyright policy](https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin).

malfatti:
Hi,

Just signed the commit. The new patch is needed because the lines to be modified are different from the previous version. Also, the `webkit-gtk` dependency had to be updated, so the live ebuilds would have to be touched anyway :)

Cheers

AndrewAmmerlaan:
Thanks, Nice work :+1: 

I had missed that the patch was actually different, at first glance it looked identical. 

With regards to the `webkit-gtk:4` dependency, packages in ::gentoo are already being actively ported to 4.1. Usually this is as easy as replacing the `4` with a `4.1` in the relevant places in the source code. At some point in the not to far future `webkit-gtk:4` will be removed, so we might want to also port this to 4.1 soon.

PR #1194, opened 2023-04-17
Fix compiling error of dev-python/pyarrow

Universebenzene:
When merge `dev-python/pyarrow` directly from this repo, it will give an error like `arrow/json/option.h not found` , which is caused by the `-DARROW_JSON` not set for compiling `dev-libs/apache-arrow` . Here I add a flag `json` to let user enable/disable that.

What's more, the `snappy` flag of apache-arrow should be enabled for pyarrow, or it may cause problems like `pyarrow.lib.ArrowNotImplementedError: Support for codec 'snappy' not built` .

This pull request should fix those problems.


Universebenzene:
@AndrewAmmerlaan Thanks for reminding. Now I have split the commit and sign-off have been added. 😀

AndrewAmmerlaan:
Nice work, Thanks :+1:

PR #1196, opened 2023-06-26
net-misc/cernbox-client: add 3.2.0.10193

olifre:
3.2.0 is still officially considered beta at CERN, but the rest of the world has moved on, so let's synchronize with `net-misc/owncloud-client` in Gentoo's main tree and add some necessary patches.


AndrewAmmerlaan:
Thanks :+1: 

I fixed the complaint from pkgcheck in a follow-up commit, it was not caused by the changes here.

olifre:
Thanks a lot! :+1:

PR #1198, opened 2023-07-04
ImageJ: Add unstable git versions

IAHM-COL:
This pull request adds ebuilds for Github unstable versioning of the imageJ package.

with `remote-id: https://github.com/imagej/ImageJ`
 
### Pull request notes to consider:

- The [latest tagged release for the Github](https://github.com/imagej/ImageJ/tags), 1.54d, added, and marked unstable, `KEYWORDS=~amd64`. It fetches the tarball using SRC_URI 
- On the commit log of the upstream repository, it is possible to identify other _"Release version_" labelled commits ([1.54e](https://github.com/imagej/ImageJ/commit/14cc1a2c3), and [1.54f](https://github.com/imagej/ImageJ/commit/5addb754)), but these were not tagged yet. No ebuild was made for these. 
- The current head (bleeding edge) was added version 9999.  It fetches using the git-r3 class.

Regards


sci-biology/imagej/imagej-1.53t.ebuild
```diff
@@ -21,7 +21,7 @@ SRC_URI="https://imagej.nih.gov/ij/download/src/${MY_PN}${MY_PV}-src.zip
 LICENSE="public-domain" # http://imagej.net/disclaimer.html
 SLOT="0"
 
-KEYWORDS="~amd64"
+KEYWORDS="amd64"
```
AndrewAmmerlaan, line 24: Since this is an overlay, we should as a matter of policy stick to unstable keyword always.

sci-biology/imagej/imagej-9999.ebuild
```diff
@@ -0,0 +1,116 @@
...
+IUSE="doc debug"
+
+RDEPEND="
+	>=virtual/jre-1.7:*
+	dev-java/java-config
+"
+DEPEND="${RDEPEND}
+	>=virtual/jdk-1.7:*
+	dev-java/ant-core
+	app-arch/unzip
```
AndrewAmmerlaan, line 39: unzip is probably a BDEPEND, maybe ant-core too.

AndrewAmmerlaan (changes requested):

AndrewAmmerlaan:
Could you please commit using `pkgdev commit`? This will automatically format the commit title correctly. Also, can you add a `Signed-off-by:` line to your commit message, pkgdev can do this automatically for you too.

sci-biology/imagej/imagej-1.53t.ebuild
IAHM-COL: Question 1: I tried KEYWORDS="~amd64" on both 1.53t and 1.54d, but `pkgcheck scan` complains of this, saying something along the lines of _overshadowing packages_. 
Is this of importance? or can I make them both ~amd64 and accept pkgcheck complaining?

IAHM-COL: `
sci-biology/imagej
  RedundantVersion: version 1.53t: slot(0) keywords are overshadowed by version: 1.54d
`

sci-biology/imagej/imagej-9999.ebuild
IAHM-COL: This works

IAHM-COL:
> Could you please commit using pkgdev commit? This will automatically format the commit title correctly. Also, can you add a Signed-off-by: line to your commit message, pkgdev can do this automatically for you too.

@AndrewAmmerlaan 

```
$ pkgdev commit
error: gpg failed to sign the data
fatal: failed to write commit object
```

AndrewAmmerlaan:
> > Could you please commit using pkgdev commit? This will automatically format the commit title correctly. Also, can you add a Signed-off-by: line to your commit message, pkgdev can do this automatically for you too.
> 
> @AndrewAmmerlaan
> 
> ```
> $ pkgdev commit
> error: gpg failed to sign the data
> fatal: failed to write commit object
> ```

This means git is not configured to use your gpg key, or gpg fails to sign with your key for some other reason.

sci-biology/imagej/imagej-1.53t.ebuild
AndrewAmmerlaan: This is just warning you that the old version is a candidate for removal soon. You can ignore it for now.

IAHM-COL: ignored. New commit has both on ~amd64 now

IAHM-COL:
> > > Could you please commit using pkgdev commit? This will automatically format the commit title correctly. Also, can you add a Signed-off-by: line to your commit message, pkgdev can do this automatically for you too.
> > 
> > 
> > @AndrewAmmerlaan
> > ```
> > $ pkgdev commit
> > error: gpg failed to sign the data
> > fatal: failed to write commit object
> > ```
> 
> This means git is not configured to use your gpg key, or gpg fails to sign with your key for some other reason.

Indeed, it isnt. my gpg key is a different identity over the ssh used to github.

Seems I would be facing quite a workaround to pkgdev commit. 

If you decide,  you'll rebase rather than merge, so you could sign it. I'm Hoping.

AndrewAmmerlaan:
> > This means git is not configured to use your gpg key, or gpg fails to sign with your key for some other reason.
> 
> Indeed, it isnt. my gpg key is a different identity over the ssh used to github.
> 
> Seems I would be facing quite a workaround to pkgdev commit.

`git config --global user.signingkey <YOUR_KEY_ID_HERE>` is all that is required to make it work.

> If you decide, you'll rebase rather than merge, so you could sign it. I'm Hoping.

Signing does not matter, this is not required. But a sign-off is, the two concepts are confusingly similar named but are not the same. Signing is cryptographically proving the commit originates from you. Signing-off is simply adding a `Signed-off-by` line to your commit message as described [here](https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin), all this does is state that you are legally allowed to contribute the code in the commit.

It is very useful to use pkgdev for committing because it does all these things automatically for you. For example, currently the title of the commit is also not according to policy, pkgdev does this automatically for you too.

IAHM-COL:
@AndrewAmmerlaan 

Thanks for the explanation.
I followed your recommendation and also: 
https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key#telling-git-about-your-gpg-key

However,
```
$ git config --global user.signingkey
****************BF88
```

After trying to save the commit after` pkgdev commit`, I still get;

```
error: gpg failed to sign the data
fatal: failed to write commit object

```

IAHM-COL:
https://github.com/gentoo/sci/commit/6ab7ca8955f54b9c14a1e54ae5cceab50d19b412

Add the Signed by line in the commit message,

Regards

AndrewAmmerlaan:
> [6ab7ca8](https://github.com/gentoo/sci/commit/6ab7ca8955f54b9c14a1e54ae5cceab50d19b412)
> 
> Add the Signed by line in the commit message,
> 
> Regards

Thanks

> After trying to save the commit after pkgdev commit, I still get;

Does it work if you `gpg --sign` some random file?

IAHM-COL:
> > [6ab7ca8](https://github.com/gentoo/sci/commit/6ab7ca8955f54b9c14a1e54ae5cceab50d19b412)
> > Add the Signed by line in the commit message,
> > Regards
> 
> Thanks
> 
> > After trying to save the commit after pkgdev commit, I still get;
> 
> Does it work if you `gpg --sign` some random file?

Sure it does. It asks my passphrase, then creates a signed random.gpg file.

on this case, with pkgdev commit it opens an emacs commit log editor, but fails to ever ask my passphrase. It might be since I am working sshed to the gentoo terminal, and over a screen session? I am not sure, but once the emacs closes ready to commit it dies with the message above

I could try with magit, if you need it, or wait till I am physically on the gentoo box.

AndrewAmmerlaan:
> on this case, with pkgdev commit it opens an emacs commit log editor, but fails to ever ask my passphrase. It might be since I am working sshed to the gentoo terminal, and over a screen session? I am not sure, but once the emacs closes ready to commit it dies with the message above

Aaah yes, that explains the problem. To use a local gpg-agent (and thus local keys) on a remote client you'll need to forward the gpg socket over ssh. This is actually quiet simple, the procedure is described here: https://wiki.gnupg.org/AgentForwarding

In any case there is no rush to get it working, I still have to actually test this PR.

IAHM-COL:
> Aaah yes, that explains the problem. To use a local gpg-agent (and thus local keys) on a remote client you'll need to forward the gpg socket over ssh. This is actually quiet simple, the procedure is described here: https://wiki.gnupg.org/AgentForwarding


Interesting content.  I never knew of forwarding the GPG to use it remotely!

Unlikely to be the issue, here today, as the gpg key I am attempting to use is on the remote, gentoo box, not on the client I used to connect to it.

Thanks for your patient guidance, tho.

IAHM-COL:
@AndrewAmmerlaan 

I had pushed a new commit to this pull request, [999bd01](https://github.com/gentoo/sci/pull/1198/commits/999bd01069a7423b4f3b91bcab3c26d643ef7c4d)

### Notes to commit:

Having `KEYWORD=~amd64` both in 1.53t and 1.54d  makes logical sense based on your comment that this is an overlay. However, when I tried to `emerge -auDU @world`, of course, emerge immediately bumped to 1.54d, unless accept keyword was set to **, accepting the bleeding edge. 

The issue at hand was the plugins, which are only available in the zip for 1.53

I reasoned, that the best course of action was then to allow the most updated plugins to be also fetched and installed on the 1.54d and in the 9999 version ebuilds.  The installation with those changes in the commit then install the plugins available. No apparent version error occurred and the software appeared to handle the plugins graciously, which made sense. 

the current version of the ebuilds therefore allow using the USE=plugins for all the versions.

### On another note; 

The pkgdev commit works if I am locally on the gentoo, but not if I am sshing remotely. Most importantly,

1. I created a new gpg key for the IAHMCOL identity
2. I had to pass the command as `pgkdev commit --gpg-sign`
3. It did not create the Signed-off-by: line in the commit, so I added it manually, while editing the commit log

I hope that sufficed

Regards
@IAHM-COL

AndrewAmmerlaan:
> It did not create the Signed-off-by: line in the commit, so I added it manually, while editing the commit log

For the sign-off line you'll need the `--signoff` argument or `commit.signoff = true` in `.config/pkgdev/pkgdev.conf`.

There's something weird going on with your patch and I can't merge it for testing:
```
andrew@andrew-gentoo-laptop imagej % pram -r gentoo/sci 1198 --no-signoff
--2023-07-07 11:55:02--  https://github.com/gentoo/sci/pull/1198.patch
Resolving github.com (github.com)... 140.82.121.3
Connecting to github.com (github.com)|140.82.121.3|:443... connected.
HTTP request sent, awaiting response... 302 Found
Location: https://patch-diff.githubusercontent.com/raw/gentoo/sci/pull/1198.patch [following]
--2023-07-07 11:55:02--  https://patch-diff.githubusercontent.com/raw/gentoo/sci/pull/1198.patch
Resolving patch-diff.githubusercontent.com (patch-diff.githubusercontent.com)... 140.82.121.3
Connecting to patch-diff.githubusercontent.com (patch-diff.githubusercontent.com)|140.82.121.3|:443... connected.
HTTP request sent, awaiting response... 200 OK
Cookie coming from patch-diff.githubusercontent.com attempted to set domain to github.com
Length: unspecified [text/plain]
Saving to: ‘/tmp/tmp.WOseb3u5rL/all.patch’

/tmp/tmp.WOseb3u5rL/all.patch                    [ <=>                                                                                        ]  15.27K  --.-KB/s    in 0.01s

2023-07-07 11:55:03 (1.32 MB/s) - ‘/tmp/tmp.WOseb3u5rL/all.patch’ saved [15633]

QSocketNotifier: Can only be used with threads started with QThread
QSocketNotifier: Can only be used with threads started with QThread
Do you want to merge this? (Y/n/q) y
Applying: ImageJ: add unstable git versions
error: sha1 information is lacking or useless (sci-biology/imagej/Manifest).
error: could not build fake ancestor
Patch failed at 0001 ImageJ: add unstable git versions
hint: Use 'git am --show-current-patch=diff' to see the failed patch
When you have resolved this problem, run "git am --continue".
If you prefer to skip this patch, run "git am --skip" instead.
To restore the original branch and stop patching, run "git am --abort".
git am failed
```

Could you maybe squash all the commits together into a single patch? Not sure if that will fix it, but we like to use 1 commit per package change anyway.

IAHM-COL:
> 
> Could you maybe squash all the commits together into a single patch? Not sure if that will fix it, but we like to use 1 commit per package change anyway.

Ok, that's strange, but we can try again, with [efac488](https://github.com/gentoo/sci/pull/1198/commits/efac4887bfeca1868845be5816d7a9a44396f5c8)

Regards

sci-biology/imagej/imagej-1.54d.ebuild
```diff
@@ -0,0 +1,128 @@
...
+	dev-java/ant-core
+	app-arch/unzip
+"
+
+IJ_S=${WORKDIR}/ImageJ
+
+src_prepare() {
+	cp "${DISTDIR}"/ImageJ.png "${WORKDIR}/${PN}.png" || die
+
+	if ! use debug ; then
```
AndrewAmmerlaan, line 52: Don't we need the `if use plugins` stuff from the other ebuild here as well?

sci-biology/imagej/imagej-9999.ebuild
```diff
@@ -0,0 +1,131 @@
...
+
+EGIT_REPO_URI="https://github.com/imagej/ImageJ"
+
+SRC_URI="
+	https://rsb.info.nih.gov/ij/images/ImageJ.png
+	plugins? ( https://wsr.imagej.net/distros/cross-platform/${MY_PN}${IJ_PV}.zip )"
+# plugins are under a different licenses and can be installed into user's $IJ_HOME/plugins
+#	plugins? ( http://rsb.info.nih.gov/ij/download/zips/${MY_PN}${IJ_PV}.zip )"
+
+KEYWORD="**"
```
AndrewAmmerlaan, line 26: It's `KEYWORDS` not `KEYWORD`. But in any case live 9999 ebuilds should just have no KEYWORDS variable of have it set to an empty string.
```suggestion
```

AndrewAmmerlaan (changes requested):
Great, now merging works. Tested it and looks good, just some minor comments left.

sci-biology/imagej/imagej-1.54d.ebuild
```diff
@@ -0,0 +1,128 @@
...
+	https://imagej.nih.gov/ij/
+	https://github.com/imagej
+"
+
+SRC_URI="
+	https://github.com/imagej/ImageJ/archive/refs/tags/v${PV}.tar.gz -> ${P}.gh.tar.gz
+	https://rsb.info.nih.gov/ij/images/ImageJ.png
+	plugins? ( https://wsr.imagej.net/distros/cross-platform/${MY_PN}${IJ_PV}.zip )"
+# plugins are under a different licenses and can be installed into user's $IJ_HOME/plugins
+#	plugins? ( http://rsb.info.nih.gov/ij/download/zips/${MY_PN}${IJ_PV}.zip )"
```
AndrewAmmerlaan, line 23: I guess we don't need this line anymore
```suggestion
```

sci-biology/imagej/imagej-1.54d.ebuild
```diff
@@ -0,0 +1,128 @@
...
+"
+DEPEND="${RDEPEND}
+	>=virtual/jdk-1.7:*
+"
+BDEPEND="
+	dev-java/ant-core
+	app-arch/unzip
+"
+
+IJ_S=${WORKDIR}/ImageJ
```
AndrewAmmerlaan, line 47:
```suggestion
IJ_S="${WORKDIR}/ImageJ"
```

sci-biology/imagej/imagej-9999.ebuild
```diff
@@ -0,0 +1,131 @@
...
+	app-arch/unzip
+"
+
+IJ_S=${WORKDIR}/ImageJ
+
+src_prepare() {
+	cp "${DISTDIR}"/ImageJ.png "${WORKDIR}/${PN}.png" || die
+
+	if use plugins ; then
+		unzip "${DISTDIR}/${MY_PN}${IJ_PV}.zip" -d "${WORKDIR}"
```
AndrewAmmerlaan, line 52: Instead of unzip (which is an external command and needs `|| die`. Use `unpack` which is the ebuild specific function for this.

sci-biology/imagej/imagej-1.54d.ebuild
IAHM-COL: Hi @AndrewAmmerlaan 
I am not sure it is needed, but in the imagej-1.54d ebuild, the SRC_URI is actually both fetching and unpacking the ImageJ zip sources and unpacking them correctly. 

It only appeared necessary for the 9999 ebuild, which after inheriting the git-r3 class was no longer unpacking every source listed on the SRC_URI

In other others words; No. It did not seem necessary, except for the 9999 to explicitly unpack.

sci-biology/imagej/imagej-9999.ebuild
IAHM-COL: @AndrewAmmerlaan 

I can't seem to be able to get it to work with unpack.

Code in sci-biology/imagej/imagej-9999.ebuild

```
src_unpack(){
	if use plugins ; then
		unpack "${MY_PN}${IJ_PV}.zip"
	fi
}

src_prepare() {
	cp "${DISTDIR}"/ImageJ.png "${WORKDIR}/${PN}.png" || die

	if ! use debug ; then
		sed -i 's: debug="on">: debug="off">:' "${S}"/build.xml || die
	fi
	default
}
```

Fails as follow:

```
>>> Emerging (1 of 1) sci-biology/imagej-9999::localrepo
 * ImageJ.png BLAKE2B SHA512 size ;-) ...                                                                                                                                                                                  [ ok ]
 * ij153.zip BLAKE2B SHA512 size ;-) ...                                                                                                                                                                                   [ ok ]
 * Using: openjdk-bin-17
>>> Unpacking source...
>>> Source unpacked in /var/tmp/portage/sci-biology/imagej-9999/work
 * ERROR: sci-biology/imagej-9999::localrepo failed (prepare phase):
 *   The source directory '/var/tmp/portage/sci-biology/imagej-9999/work/imagej-9999' doesn't exist
 * 
 * Call stack:
 *            ebuild.sh, line  780:  Called __ebuild_main 'prepare'
 *   phase-functions.sh, line 1072:  Called __dyn_prepare
 *   phase-functions.sh, line  400:  Called die
 *              die "The source directory '${S}' doesn't exist"
 * 
 * If you need support, post the output of `emerge --info '=sci-biology/imagej-9999::localrepo'`,
 * the complete build log and the output of `emerge -pqv '=sci-biology/imagej-9999::localrepo'`.
!!! When you file a bug report, please include the following information:
GENTOO_VM=openjdk-bin-17  CLASSPATH="" JAVA_HOME="/opt/openjdk-bin-17.0.6_p10"
JAVACFLAGS="-source 1.7 -target 1.7" COMPILER=""
and of course, the output of emerge --info =imagej-9999
 * The complete build log is located at '/var/tmp/portage/sci-biology/imagej-9999/temp/build.log'.
 * The ebuild environment file is located at '/var/tmp/portage/sci-biology/imagej-9999/temp/environment'.
 * Working directory: '/var/tmp/portage/sci-biology/imagej-9999/empty'
 * S: '/var/tmp/portage/sci-biology/imagej-9999/work/imagej-9999'

>>> Failed to emerge sci-biology/imagej-9999, Log file:

```

To be noted, the src_unpack is causing the git url NOT to be fetched at all, and therefore the " The source directory '/var/tmp/portage/sci-biology/imagej-9999/work/imagej-9999' doesn't exist". (True, I verified and it didnt')

The unpacking worked ok.

Not sure if I am missing something. 
Or if you want me to push the changes as is, not working for you to look at it,
Or if in this case the unzip use is acceptable and we add || die

Regards

sci-biology/imagej/imagej-1.54d.ebuild
AndrewAmmerlaan: > It only appeared necessary for the 9999 ebuild, which after inheriting the git-r3 class was no longer unpacking every source listed on the SRC_URI

This happens because the `git-r3.eclass` comes with its own `git-r3_src_unpack()` overriding the default `src_unpack()`. If you want to use both, you'll need to do something like this:
```
src_unpack() {
    default
    git-r3_src_unpack
}
```

I would suggest structuring both ebuilds like this:
```
inherit ....

DESCRIPTION=...
HOMEPAGE=...
SRC_URI=.... # The png and the plugins

if [[ ${PV} == 9999 ]]; then
    inherit git-r3
    EGIT_REPO_URI=....
    S=.....
else
    SRC_URI+=.... # The main sources
    S=.... 
    KEYWORDS="~amd64"
fi

src_unpack() {
    default
    if [[ ${PV} == 9999 ]]; then
         git-r3_src_unpack
    fi

.....

```
This way both ebuild files can be identical, this simplifies maintenance later on, for the next release you can just update the 9999, copy and rename it.

IAHM-COL: Interesting ebuild structure. I surely don't hate it, but also I surely don't thing it necessarily simplifies maintenance. 

As the ebuilds currently look, the unpack is only being called for the situation that appear to require it. That is, when the default unpack worked graciously (in 1.53 and 1.54), then nothing was added. 
Say, KISS.

On the maintenance issue, it seems the current situation is a pretty straightfoward case:

1. for 1.53: Rename ebuild accordingly when the NIH releases 1.54 zip sources
2. for new tag releases in github: Rename ebuild accordingly when releases are Tagged in the repository
3. for the plugins in the 1.54 and 9999 versions: Update the variables that control the version of plugins on these ebuilds. 

But I would admit, here, we enter a bit on stylistic issues; but also a lot more on an experience situation. Therefore, my initial reaction being: I don't concur. But I can be persuaded to the contrary.

AndrewAmmerlaan: > On the maintenance issue, it seems the current situation is a pretty straightfoward case:
> 
>     1. for 1.53: Rename ebuild accordingly when the NIH releases 1.54 zip sources
> 
>     2. for new tag releases in github: Rename ebuild accordingly when releases are Tagged in the repository
> 
>     3. for the plugins in the 1.54 and 9999 versions: Update the variables that control the version of plugins on these ebuilds.

Yes this is one way to do it, we go from old, copy to new and then make changes. The argument to do it the other way around is that changes always land in 9999 first. Therefore if 9999 had to be changed in between releases you don't have to make this same change again in the next release if you copy from 9999 instead of from the previous release. For this to work 9999 has to be written in a way that the git stuff is behind if statements.

The second argument to use the if statements is that the base structure of the ebuilds is then basically the same, this way you can easily spot changes in configure/install phases between different releases and live.

This is the way we usually do it, if you take a look at ::gentoo you'll see that most/all of the live ebuilds use this if statement structure. That being said, if you're maintaining it you decide so feel free to use a different workflow.

IAHM-COL: Ok. given the method you outline is the Gentoo way for most/all live ebuilds, I shall go ahead and try if I can make the ebuilds work with the structure proposed. 
I'll let you know if I can (or not) make it work.

Regards

IAHM-COL: https://github.com/gentoo/sci/pull/1198/commits/365a963f3d5b436e0961cdd70b1fd7e3bc5770e9 

Restructures the new ebuilds as suggested, versions 1.54d and 9999, which fetch from github, are the same file copied over. Conditional statements on the ebuild handle the differences between versions.

Regards

AndrewAmmerlaan:
Nice work, Thanks :+1:

PR #1199, opened 2023-07-11
media-libs/imglib2: Add new package

IAHM-COL:
**Pull requests notes:**

- It adds [imglib2](https://imagej.net/libs/imglib2/)
- I used java-ebuilder for the initial ebuild parsing, and then manually edit
- Per the pom file, jmh-core-1.36 is required, however, gentoo has [jmh-core-1.35 ](https://packages.gentoo.org/packages/dev-java/jmh-core)at the moment. 
- I changed the ebuild to accept 1.35, but did not compile with 'test' flag, in order to see any adverse consequence. When possible, this configuration can be updated.

Regards
@IAHM-COL


media-libs/imglib2/imglib2-6.2.0.ebuild
```diff
@@ -0,0 +1,53 @@
...
+"
+
+RDEPEND="
+	>=virtual/jre-1.8:*
+"
+
+JAVA_SRC_DIR="src/main/java"
+JAVA_MAIN_CLASS=""
+
+JAVA_TEST_GENTOO_CLASSPATH="jama-1.0.3,junit-4.13.2,jmh-core-1.35"
```
AndrewAmmerlaan, line 52: And I guess we would have the same problem here.

IAHM-COL: @AndrewAmmerlaan 

I pushed [8d3f964](https://github.com/gentoo/sci/pull/1199/commits/8d3f964577491d21811cc37c4fac6412ca7db9f1)


I am not sure if this fixed it, but it is worth to try

Regards

AndrewAmmerlaan: Sorry for the delay, I've been pre-occupied with `secureboot.eclass`

There is still some problem here:
```
>>> Compiling source in /tmp/portage/media-libs/imglib2-6.2.0/work/imglib2-imglib2-6.2.0 ...
!!! ERROR: Package jama was not found!
```

It looks like `jama` is not installing any java stuff at all:
```
andrew@andrew-gentoo-laptop ~ % equery f jama
* Searching for jama ...
* Contents of sci-libs/jama-1.2.5:
/usr
/usr/include
/usr/include/jama_cholesky.h
/usr/include/jama_eig.h
/usr/include/jama_lu.h
/usr/include/jama_qr.h
/usr/include/jama_svd.h
```
Are you sure this is the correct dependency?

IAHM-COL: Hi @AndrewAmmerlaan :

1. Welcome back. No worries. I have been troubleshooting the ebuilds for bioformats and scijava-common; facing issues on that end before I can send pull request here. So no much in a hurry
2. I hope the secureboot.class is going well. 
3. It sounds like sci-libs/jama is right, 

- per example on the [pom.xml](https://github.com/imglib/imglib2/blob/master/pom.xml) file of imglib2,  line 250
```
<dependency>
			<groupId>gov.nist.math</groupId>
			<artifactId>jama</artifactId>
			<scope>test</scope>
</dependency>
```
It indicates that with the scope of test, jama by gov.nist.math is dependency, 

-Looking on the gentoo packages, for jama results in 
```
Java-like matrix C++ templates
https://math.nist.gov/tnt/

```
Which clearly points to the same gropuId.

However, when I check the package
`emerge -pav1 jama`
I get
```

!!! All ebuilds that could satisfy "jama" have been masked.
!!! One of the following masked packages is required to complete your request:
- sci-libs/jama-1.2.5::gentoo (masked by: ~amd64 keyword)

```
Nothing that it has to be unmasked, for ~amd64 so it installs.


So, bottomline, _I think yes,_ that's the dependency that's being called for

However, also by math.nist.gov, I find this
https://math.nist.gov/javanumerics/jama/
which appears to be java classes (rather than C classes)

Same upstream, same name, different code? `I presume, not`

IAHM-COL: I don't know the answer to this.  
on https://math.nist.gov/javanumerics/jama/ it reads

```
JAMA was originally created as a proof-of-concept *** As such, it is no longer actively developed to keep track of evolving usage patterns in the Java language, nor to further improve the API. We will, however, fix outright errors in the code.
```

I don't know if this would mean it's unlikely that making an ebuild for this is a good forward plan. I wonder if we remove it as a test BDEPEND if a test will fail to compile+

AndrewAmmerlaan: If you take a look at the sources listed at `https://math.nist.gov/javanumerics/jama/` then you'll see the version numbers don't match at all with `sci-libs/jama`. So they must be different libraries with the same name.

IAHM-COL: ok. I will try making the ebuild for this dependency (Even thou development on that is arrested).

I'll place it on dev-java/jama, which is going to create the case for ambiguity, unfortunately

Regards

IAHM-COL: I added dev-java/jama to Sci in a new pull request and with [af69a26](https://github.com/gentoo/sci/pull/1199/commits/af69a26725b6487358ff33b0705639e629bc43a6) redirected the dependency

AndrewAmmerlaan:
Merged

IAHM-COL:
thanks @AndrewAmmerlaan 
I am glad the jama dependency was correct this time
Regards @IAHM-COL

PR #1200, opened 2023-07-12
Add parsington

IAHM-COL:
### Notes to pull request

- It adds scijava's parsington
- ebuild initially created with java-ebuilder
- ebuild manually adjusted,
- It includes last tagged and bleeding edge


dev-java/parsington/parsington-9999.ebuild
```diff
@@ -0,0 +1,46 @@
...
+"
+
+RDEPEND="
+	>=virtual/jre-1.8:*
+"
+
+JAVA_SRC_DIR="src/main/java"
+JAVA_MAIN_CLASS="org.scijava.parsington.Main"
+
+JAVA_TEST_GENTOO_CLASSPATH="junit-5.9.1"
```
AndrewAmmerlaan, line 45: I'm getting this error: `!!! ERROR: Package junit-5.9.1 was not found!`

The version of junit I have installed is 5.9.2. I'm not super familiar with java packaging, but is there any way we can write this so it doesn't break every time junit updates. Does it work if you simply don't specify the version number?

IAHM-COL: Interesting. 
I am not :100:  either :worried: 
But I just pushed it as "junit-5", as many ebuilds in the dev-java subdir of the gentoo repo appear to have a line like this:

`
JAVA_TEST_GENTOO_CLASSPATH="junit-4"
`
I hope it works. 
I didn't build with the test flag.

Also, I am not sure how your are testing, but I suggest 

`emerge -av1`

Since these are really being added as a dependency tree, and dependencies are better installed with the `1`, for depclean to operate graciously

Regards

dev-java/parsington/parsington-3.1.0.ebuild
```diff
@@ -0,0 +1,46 @@
...
+	KEYWORDS="~amd64"
+fi
+
+LICENSE="BSD-2"
+SLOT="0"
+
+BDEPEND="
+	>=virtual/jdk-1.8:*
+	test? (
+	      >=dev-java/junit-5.9.1
```
AndrewAmmerlaan, line 34: Now it compiles fine, but there is still a warning:
```
* java-pkg_ensure-dep: The ebuild is attempting to use junit-5, which is not declared with a SLOT in DEPEND.
* java-pkg_ensure-dep: The ebuild is attempting to use junit-5, which is not declared with a SLOT in DEPEND.
```

We need slot dependency `:5` on the `junit` dependency. The warning seems to imply this should be in `DEPEND` instead of `BDEPEND` so maybe I was wrong earlier, or maybe the warning message will go away as well if the slot operator is added to `BDEPEND`.

IAHM-COL: I just pushed https://github.com/gentoo/sci/pull/1200/commits/ad997359ec931fca721596d50c53df04ffba0f4b
I think that might suppress the warning by indicating any slot

regards

IAHM-COL: `The warning seems to imply this should be in `DEPEND` instead of `BDEPEND` so maybe I was wrong earlier, or maybe the warning message will go away as well if the slot operator is added to `BDEPEND`.
`
I think BDEPEND is right, since is a test? dependency

IAHM-COL: 1. Is junit:5 (slot5 ) actually junit-5?
2. will the ebuild break against future junit-6?
Regards

AndrewAmmerlaan: > 1. Is junit:5 (slot5 ) actually junit-5?

Yes, in this case it is

> will the ebuild break against future junit-6?

Yes, because we have `JAVA_TEST_GENTOO_CLASSPATH="junit-5"` in the ebuild.

IAHM-COL: ok, I will send a fixed (and squashed) pull request then

IAHM-COL: Hi @AndrewAmmerlaan 

Changed to

```
BDEPEND="
	>=virtual/jdk-1.8:*
	test? (
	      >=dev-java/junit-5.9.1:5
```

squashed the commit and updated the pull request,

Regards
@IAHM-COL

IAHM-COL: @AndrewAmmerlaan 

Commit https://github.com/gentoo/sci/pull/1200/commits/b99c9c2e7aca7e09682e4da68f981284813fc934 might suppress the SLOT in DEPEND warning now


Regards
@IAHM-COL

PR #1202, opened 2023-07-21
dev-java/scijava-common: new package, add 2.94.2, 9999

IAHM-COL:
Pull Requests Notes:

It adds Scijava-common:
https://imagej.net/libs/scijava#scijava-common

* It complains of missing SLOT in DEPEND, with dependencies, even thou these are added as BDEPEND or RDEPEND and a slot is given


AndrewAmmerlaan:
> * It complains of missing SLOT in DEPEND, with dependencies, even thou these are added as BDEPEND or RDEPEND and a slot is given

Then they should be moved from BDEPEND to DEPEND. Languages like java and python are a bit of a grey area when it comes to DEPEND versus BDEPEND, since the dependencies are usually executed but often there is no actual architecture dependent compiling involved. With python we always use BDEPEND in this case, but it appears the java eclasses have made a different decision.

IAHM-COL:
> 

Interesting. That might be the case with parsington too. Let me know, as in my gentoo it exists quiet since I build with no test flags

AndrewAmmerlaan:
> Interesting. That might be the case with parsington too. Let me know, as in my gentoo it exists quiet since I build with no test flags

Yes, with `USE="test" FEATURES="test"` this also still gives the same warning

IAHM-COL:
> > * It complains of missing SLOT in DEPEND, with dependencies, even thou these are added as BDEPEND or RDEPEND and a slot is given
> 
> Then they should be moved from BDEPEND to DEPEND. Languages like java and python are a bit of a grey area when it comes to DEPEND versus BDEPEND, since the dependencies are usually executed but often there is no actual architecture dependent compiling involved. With python we always use BDEPEND in this case, but it appears the java eclasses have made a different decision.

With https://github.com/gentoo/sci/pull/1202/commits/75449a7410bfafb815a200f622fc5921b001a225 the SLOT warning is now suppressed in this package

I used the Commond Dependencies "CDEPEND" strategy outlined [here](https://wiki.gentoo.org/wiki/Java_Developer_Guide#Slots):


and includes a DEPEND line

Regards
IAHMCOL

dev-java/scijava-common/scijava-common-2.94.2.ebuild
```diff
@@ -0,0 +1,68 @@
...
+	dev-java/jaxws-api:0
+	dev-java/jaxb-api:2
+"
+
+BDEPEND="
+	>=virtual/jdk-1.8:*
+	${CDEPEND}
+	test? (
+		>=dev-java/junit-4.13.2
+		>=dev-java/mockito-2.19.0
```
AndrewAmmerlaan, line 46:
```suggestion
		>=dev-java/mockito-2.19.0:2
```
Since we explicitly use `mockito-2` in CLASSPATH below

PR #1207, opened 2023-07-23
dev-java/scijava-table: new package, add 1.0.2, 9999

IAHM-COL:
Add scijava-table

It depends on scijava-common and scijava-optional, added in further pull requests sent for consideration


AndrewAmmerlaan:
Thanks :+1:

PR #1209, opened 2023-07-26
sci-physics/particle: new package, add 0.23.0, 9999

APN-Pucky:
This is a very useful python package `scikit-hep/particle` for getting data from the PDG.

While these packages could also be sorted to `dev-python` and are purely available from pypi, they are clearly only of physics purpose so I placed them there.

I have many more new hep-physics packages (and bumps), but I'll only PR them slowly sorted by relevance, to avoid unused packages in `::sci` and keep them and experimental/recent stuff in `::hep-forge`.


APN-Pucky:
`dev-python/pytest-benchmark` exists already in `::guru` and I don't want to copy that ebuild here. Can the `::guru` be moved to `::gentoo`?

AndrewAmmerlaan:
> Otherwise I could also just patch the benchmarking away, but that feels a bit ugly.

This is what we usually do, the benchmarking tests usually aren't very useful downstream. (Same for coverage tests etc)

AndrewAmmerlaan (approved):
Haven't tested because I'm away for the week, but LGTM

APN-Pucky:
Thanks, I'll merge it then. (For completeness I successfully ran `pkg-testing-tool` here https://gitlab.com/APN-Pucky/gentoo-hep-forge/-/commit/99e93b1bc74c5a8b1c2388231f7af3dcdbbb7d28/pipelines?ref=master)

PR #1210, opened 2023-08-04
net-misc/cernbox-client: add 4.1.0.11250, drop oldest version

olifre:
This syncs again with `net-misc/owncloud-client` from `gentoo`s main tree, and pulls in latest themeing sources from CERNs current "future" packages. 
It also drops a leftover, unused patch file and drops the oldest (now legacy) version.


AndrewAmmerlaan:
Thanks :+1:

PR #1211, opened 2023-08-07
dev-python/cadquery: updating dependencies

Equidamoid:
The actual cadquery changes are coming later in a separate PR, pywrap needs some patching


Equidamoid:
Not sure how to deal with the pkgcheck errors:
```
NonsolvableDepsInStable: version 0.12.3: nonsolvable depset(depend) keyword(~amd64) stable profile (default/linux/amd64/17.1) (30 total): solutions: [ >=dev-libs/spdlog-1.8.5[abi_x86_32(-),abi_x86_64(-)] ]
(...)
```
Any hints?

dev-libs/lief/lief-0.12.3.ebuild
```diff
@@ -27,6 +27,8 @@ KEYWORDS="~amd64 ~x86"
 IUSE="c examples +python static-libs"
 REQUIRED_USE="python? ( ${PYTHON_REQUIRED_USE} !abi_x86_32 !abi_x86_x32 )"
 
+PATCHES=( "${FILESDIR}" )
```
AndrewAmmerlaan, line 30: Why is this here?

AndrewAmmerlaan (changes requested):
When adding 3.11 it is not required to drop 3.10. 3.10 is still supported by us so please keep it around if possible. Have you also tested 3.12? If we can also add 3.12 now it will save work later.

AndrewAmmerlaan:
Now we have the additional problem of cadquery depending on `dev-python/nptyping` which was removed because it was broken and lacked upstream activity.

Equidamoid:
Sorry, didn't have time to look into this. Will try to come back to it soon.

AndrewAmmerlaan:
This has been treecleaned for now, but we can still re-add it later. We need to:
- version bump
- update toml dependency
- update python_compats

PR #1212, opened 2023-08-13
net-fs/eos-client: new package

Berrysoft:
- [x] Add eos-client.
- [ ] Patch `/etc` directory.


Berrysoft:
The code of `eos` hardcoded the `/etc/eos` directory. It's messive to change it to `${EPREFIX}/etc/eos`. Is that OK?

heroxbd:
The `common/fmt` directory is empty.  It is a submodule in the upstream. Shall we populate its contents?

heroxbd:
`dev-libs/libfmt` is there.

There are plenty of submodules of git that needs to be downloaded and extracted before heading to `src_prepare`, much like the https://github.com/gentoo/sci/blob/203194ee62f8d4ffce6fb8278078b693af3ec87f/sci-libs/pytorch/pytorch-1.10.2.ebuild

Berrysoft:
Most of them could be replaced by the system dependencies without changes. The cmake script of eos seems clever enough.

net-fs/eos-client/eos-client-5.1.27.ebuild
```diff
@@ -0,0 +1,79 @@
...
+	net-libs/cppzmq
+	sys-apps/help2man
+	sys-fs/xfsprogs
+	dev-cpp/gtest
+"
+
+src_prepare() {
+	default
+
+	cmake_src_prepare
```
AndrewAmmerlaan, line 63: Both `default` and `cmake_src_prepare` are going to try and apply any user patches. The second try will cause the build to fails since the patches are already applied. You can probably drop the entire explicit `src_prepare` phase.

net-fs/eos-client/eos-client-5.1.27.ebuild
```diff
@@ -0,0 +1,79 @@
...
+	cmake_src_configure
+}
+
+src_compile() {
+	cmake_src_compile
+}
+
+src_install() {
+	cmake_src_install
+}
```
AndrewAmmerlaan, line 79: You don't have to set `src_compile` and `src_install` if you're only using the standard function.

net-fs/eos-client/eos-client-5.1.27.ebuild
```diff
@@ -0,0 +1,79 @@
...
+
+inherit cmake
+
+DESCRIPTION="EOS Storage"
+HOMEPAGE="https://eos-web.web.cern.ch/eos-web/"
+SRC_URI="https://github.com/cern-eos/eos/archive/refs/tags/${PV}.tar.gz"
+LICENSE="GPL-3"
+
+SLOT="0"
+KEYWORDS="~amd64"
```
AndrewAmmerlaan, line 14: Try to keep the layout as close to the ebuild skeleton as possible. (i.e. move `LICENSE` one line down)

net-fs/eos-client/eos-client-5.1.27.ebuild
```diff
@@ -0,0 +1,79 @@
...
+		app-arch/zstd
+		app-arch/lz4
+	)
+	httpd? ( net-libs/libmicrohttpd )
+"
+
+DEPEND="${RDEPEND}"
+
+BDEPEND="
+	dev-lang/python
```
AndrewAmmerlaan, line 51: If this requires python during build time, it should be inheriting `python-any-r1.eclass` and setting `PYTHON_COMPAT` to the supported python versions.

AndrewAmmerlaan (changes requested):

AndrewAmmerlaan:
> The code of `eos` hardcoded the `/etc/eos` directory. It's messive to change it to `${EPREFIX}/etc/eos`. Is that OK?

I think for `::sci` this is fine.


> Most of them could be replaced by the system dependencies without changes. The cmake script of eos seems clever enough.

Lets us system versions whenever possible.

PR #1214, opened 2023-08-26
net-misc/cernbox-client: Version bump to 4.2.0.11670, drop oldest version.

olifre:
This adds `=net-misc/cernbox-client: add 4.2.0.11670` following along `net-misc/owncloud-client` and drops the oldest version.


AndrewAmmerlaan:
Thanks :+1:

PR #1216, opened 2023-09-01
sci-physics/pyhf: new package, add 0.7.3

APN-Pucky:
This serves as a demo/discussion on including references/citations in metadata.xml.
Integration with `epkginfo`/`equery meta` works nicely out of the box, even with repeated `lang` entries (i.e. here 'doi')

```bash
$ equery meta pyhf
 * sci-physics/pyhf [hep-forge]
Maintainer:  sci@gentoo.org (Gentoo Science Project)
Maintainer:  alexander@neuwirth-informatik.de (Alexander Puck Neuwirth)
Upstream:    Docs:       https://doi.org/10.5281/zenodo.1169739 (doi)
             Docs:       https://zenodo.org/record/8256635 (zenodo)
             Docs:       https://doi.org/10.21105/joss.02823 (doi)
             Docs:       https://inspirehep.net/literature/2598491 (inspirehep)
             Docs:       https://arxiv.org/abs/2211.15838 (arxiv)
             Docs:       https://doi.org/10.22323/1.414.0245 (doi)
             Remote-ID:   particle ID: pypi
             Remote-ID:   scikit-hep/particle ID: github
Homepage:    https://github.com/scikit-hep/pyhf
Location:    /var/db/repos/hep-forge/sci-physics/pyhf
Keywords:    0.7.0:0: 
Keywords:    0.7.0-r1:0: 
Keywords:    0.7.1:0: 
Keywords:    0.7.3:0: ~amd64
License:     BSD
```
However, repeated `lang` entries are prohibited by pkgcheck, so I reduced them for this PR.
Also 'inspirehep' was not tolerated as a `lang` entry (probably due to its length).

The longer route instead of abusing the `lang`-attribute, would be another attribute maybe `provider` or instead of a `doc` element a `reference` element.

I however don't think it fits into an other `upstream`-remote id since it is not a source code repo.

Thanks and I am looking forward to your thoughts on this.


APN-Pucky:
Based on @ulm remarks I moved the references to `HOMEPAGE`. And once there are enough references (in `::gentoo`/`::sci`) one can think about dropping the uniqueness of `<doc>` in metadata.xml.

AndrewAmmerlaan (approved):
LGTM :+1:

PR #1218, opened 2023-09-19
sci-misc/openfst: add 1.8.2, drop 1.8.1

n1ghtbyte:
Update openfst package to version 1.8.2.

I would like to work on the possible use flags for this package if possible


AndrewAmmerlaan (approved):

n1ghtbyte:
I will work on the ebuild and send a PR when possible, thanks for merging.

PR #1219, opened 2023-10-23
sci-physics/lhapdf-sets: new package, add 0

APN-Pucky:
Following the discussion in https://github.com/gentoo/gentoo/pull/32056#issuecomment-1730859253 this is very similar to https://github.com/gentoo/guru/blob/master/sci-geosciences/GeographicLib-data/GeographicLib-data-0-r1.ebuild


AndrewAmmerlaan (approved):
Calling `unpack` in `src_install` is a bit unconventional, but I guess there is no real reason why it shouldn't be fine. LGTM :+1:

APN-Pucky:
Thanks, I'll merge it soon. The `src_install`/`unpack` I copied https://gitweb.gentoo.org/repo/gentoo.git/tree/sci-physics/geant-data/geant-data-4.11.1.0.ebuild

PR #1225, opened 2023-12-19
sci-biology/fastp: add ebuild

mschubert:
A widely used nucleotide sequence trimmer


sci-biology/fastp/fastp-0.23.4.ebuild
```diff
@@ -0,0 +1,21 @@
...
+EAPI=8
+
+DESCRIPTION="An ultra-fast all-in-one FASTQ preprocessor"
+HOMEPAGE="https://github.com/OpenGene/fastp"
+SRC_URI="https://github.com/OpenGene/fastp/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz"
+KEYWORDS="~amd64"
+
+LICENSE="MIT"
+SLOT="0"
+IUSE=""
```
AndrewAmmerlaan, line 13: Please drop empty vars

sci-biology/fastp/fastp-0.23.4.ebuild
```diff
@@ -0,0 +1,21 @@
...
+
+DEPEND="app-arch/libdeflate
+	dev-libs/isa-l"
+
+src_install() {
+	dodir /usr/bin
```
AndrewAmmerlaan, line 19: `dodir` should not be necessary here.

mschubert: Without this `make -j8 install` fails:
> install: cannot create regular file '/dev/shm/portage/sci-biology/fastp-0.23.4/image/home/m.schubert/gentoo/usr/bin/fastp': No such file or directory

mschubert:
Removed empty var and added signoff

AndrewAmmerlaan:
Thanks :+1:

PR #1226, opened 2023-12-19
app-shells/radian: add ebuild

mschubert:
A modern R console


app-shells/radian/radian-0.6.9.ebuild
```diff
@@ -0,0 +1,25 @@
...
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
+
+DEPEND="dev-python/setuptools[${PYTHON_USEDEP}]
+	dev-python/prompt-toolkit[$PYTHON_USEDEP]
+	dev-python/rchitect[$PYTHON_USEDEP]
+	dev-python/six[$PYTHON_USEDEP]
+	dev-lang/R"
```
AndrewAmmerlaan, line 23: These are probably `RDEPEND`s.

DEPEND is for things we need present during build
BDEPEND is for things we need to execute in order to build
RDEPEND is for things we need during runtime of what we have built

mschubert: They can indeed be `REPEND`s

mschubert:
This should be good now

AndrewAmmerlaan:
Sorry, I just noticed, but we have what looks to be the same package already in `::guru` but as `sci-mathematics/radian`. 

I'd like to avoid duplication (this only serves to double work), but the version in `::guru` is older, so maybe you can push this new version to `::guru`?

PR #1229, opened 2023-12-28
Bump scijava common

IAHM-COL:


IAHM-COL:
@AndrewAmmerlaan 
on the tests, pkgcheck build on github fails, I can't figure out why.
It compiles and passes pkgcheck on my box
There is no llvm dependency as far as I can see on the ebuild

Regards

AndrewAmmerlaan:
Thanks :+1: 

Pkgcheck complaint is caused by llvmlite which still depends on llvm:14 which was recently removed.

PR #1232, opened 2024-02-01
sys-cluster/easybuild*: add 4.9.0, 9999 

APN-Pucky:


sys-cluster/easybuild-easyblocks/easybuild-easyblocks-4.9.0.ebuild
```diff
@@ -0,0 +1,37 @@
+EAPI=8
+
+PYTHON_COMPAT=( python3_{9..11} )
```
AndrewAmmerlaan, line 3:
```suggestion
PYTHON_COMPAT=( python3_{10..11} )
```

Old python versions should be dropped otherwise we'll get annoying complaints from portage.

sys-cluster/easybuild-easyblocks/easybuild-easyblocks-4.9.0.ebuild
```diff
@@ -0,0 +1,37 @@
...
+	https://easybuild.io/
+	https://github.com/easybuilders/easybuild-easyblocks
+"
+
+LICENSE="GPL-2"
+SLOT="0"
+if [[ ${PV} == 9999 ]]; then
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/easybuilders/easybuild-easyblocks"
+	SUB_P=9999
```
AndrewAmmerlaan, line 22:
```suggestion
```

This I think is unnecessary `ver_cut 1` should already return 9999 when that is the PV.

sys-cluster/easybuild-easyblocks/easybuild-easyblocks-4.9.0.ebuild
```diff
@@ -0,0 +1,37 @@
...
+fi
+
+RDEPEND="
+	>=sys-cluster/easybuild-framework-${SUB_P}[${PYTHON_USEDEP}]
+"
+BDEPEND="${RDEPEND}"
+
+src_prepare() {
+	# already there from easybuild_framework
+	rm easybuild/__init__.py
```
AndrewAmmerlaan, line 35:
```suggestion
	rm easybuild/__init__.py || die
```

sys-cluster/easybuild-easyblocks/easybuild-easyblocks-4.9.0.ebuild
AndrewAmmerlaan: (these changes should be made to the other files too, sorry, I wasn't clear before)

APN-Pucky: Nope, lazy me on the train just only hit the commit suggestion button.

PR #1234, opened 2024-02-01
dev-python/vector: new package, add 1.1.1

APN-Pucky:


dev-python/vector/vector-1.1.1.ebuild
```diff
@@ -0,0 +1,21 @@
...
+SLOT="0"
+KEYWORDS="~amd64"
+
+RDEPEND="
+	>=dev-python/numpy-1.13.3[${PYTHON_USEDEP}]
+	dev-python/hatch-vcs[${PYTHON_USEDEP}]
+"
+BDEPEND="${RDEPEND}"
+
+export SETUPTOOLS_SCM_PRETEND_VERSION=${PV}
```
AndrewAmmerlaan, line 21: We probably don't have to `export` this.

And if we are using setuptools_scm, then we should also BDEPEND on `dev-python/setuptools-scm`.

dev-python/vector/vector-1.1.1.ebuild
```diff
@@ -0,0 +1,21 @@
+EAPI=8
+
+PYTHON_COMPAT=( python3_{10..12} )
+DISTUTILS_USE_PEP517=hatchling
+inherit distutils-r1 pypi
+
+DESCRIPTION="Vector classes and utilities"
+HOMEPAGE="https://github.com/scikit-hep/vector"
+#SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
```
AndrewAmmerlaan, line 9:
```suggestion
```

dev-python/vector/vector-1.1.1.ebuild
APN-Pucky: It is through hatch-vcs cf. https://projects.gentoo.org/python/guide/distutils.html#source-archives

APN-Pucky: Hatch-vcs uses that variable too. Thats why pkgdev also does not show the warning that setuptools-vcs is missing

AndrewAmmerlaan: aah right, makes sense

PR #1235, opened 2024-02-01
dev-python/boost-histogram: new package, add 1.4.0

APN-Pucky:


dev-python/boost-histogram/boost-histogram-1.4.0.ebuild
```diff
@@ -0,0 +1,42 @@
...
+
+DESCRIPTION="Python bindings for the C++14 Boost::Histogram library"
+HOMEPAGE="https://github.com/scikit-hep/boost-histogram"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64"
+
+RDEPEND="
+	dev-libs/boost[python,${PYTHON_USEDEP}]
```
AndrewAmmerlaan, line 15:
```suggestion
	dev-libs/boost:=[python,${PYTHON_USEDEP}]
```
Probably this will need to be rebuilt when updating boost, have not checked though.

dev-python/boost-histogram/boost-histogram-1.4.0.ebuild
```diff
@@ -0,0 +1,42 @@
...
+	dev-python/numpy[${PYTHON_USEDEP}]
+	test? (
```
AndrewAmmerlaan, line 17: test dependencies should be BDEPEND not RDEPEND

dev-python/boost-histogram/boost-histogram-1.4.0.ebuild
APN-Pucky: I don't think so since i thought it uses the boost python interface, but maybe i am wrong. I'll check that

dev-python/boost-histogram/boost-histogram-1.4.0.ebuild
APN-Pucky: I guess i also do not need pytest explicitly if i declare pytest via `enable_tests`?

AndrewAmmerlaan: > I guess i also do not need pytest explicitly if i declare pytest via `enable_tests`?

Yes, it is indeed added automatically.

PR #1236, opened 2024-02-01
dev-python/jacobi: new package, add 0.8.1

APN-Pucky:


dev-python/jacobi/jacobi-0.8.1.ebuild
```diff
@@ -0,0 +1,27 @@
...
+BDEPEND="${RDEPEND}"
+
+distutils_enable_tests pytest
+
+python_test() {
+	local EPYTEST_IGNORE=(
+		tests/bench.py
+	)
+	epytest
+}
```
AndrewAmmerlaan, line 26:
```suggestion
EPYTEST_IGNORE=(
	tests/bench.py
)
```

We don't have to redefine python_test if we do it this way.

PR #1237, opened 2024-02-01
sci-physics/pyhepmc: new package, add 2.11.0

APN-Pucky:


sci-physics/pyhepmc/pyhepmc-2.11.0.ebuild
```diff
@@ -0,0 +1,17 @@
+EAPI=8
+
+PYTHON_COMPAT=( python3_{10..12} )
+DISTUTILS_USE_PEP517=setuptools
+inherit distutils-r1 pypi
+
+DESCRIPTION="A Pythonic wrapper for the HepMC3 C++ library."
```
AndrewAmmerlaan, line 7: Should we RDEPEND on hepmc?

APN-Pucky: No it compiles it by itself at a fixed version through some python build system. But i should add a BDEPEND on cmake.

AndrewAmmerlaan: It bundles hepmc? Is it feasible to make it use the system version instead?

AndrewAmmerlaan: That makes sense

PR #1238, opened 2024-02-16
sci-mathematics/ltl2ba: add 1.2, 1.3

APN-Pucky:
I also added a second `SRC_URI` since a few weeks ago the first source was down.


sci-mathematics/ltl2ba/ltl2ba-1.2.ebuild
```diff
@@ -0,0 +1,34 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
AndrewAmmerlaan, line 4:
```suggestion
EAPI=8
```

sci-mathematics/ltl2ba/ltl2ba-1.2.ebuild
```diff
@@ -0,0 +1,34 @@
...
+
+inherit flag-o-matic
+
+DESCRIPTION="Fast LTL to Buechi Automata Translation"
+HOMEPAGE="http://www.lsv.ens-cachan.fr/~gastin/ltl2ba"
+SRC_URI="
+	http://www.lsv.ens-cachan.fr/~gastin/${PN}/${P}.tar.gz
+	https://www.lrde.epita.fr/dload/spot/deps/${P}.tar.gz
+"
```
AndrewAmmerlaan, line 13: > I also added a second SRC_URI since a few weeks ago the first source was down.

I don't think we can use `SRC_URI` this way, if the tarballs are not exactly the same then user may get a hash mismatch complaint when trying to emerge the package. It's probably better to leave only the new source and maybe update the `HOMEPAGE` as well.

PR #1240, opened 2024-02-16
sci-mathematics/Oid: Fix java, update URLs

APN-Pucky:
This concludes my quick check of the sci-mathematics packages. Only `sci-mathematics/frama` still fails and I know nothing about ocaml.

### DIFF
--- a/sci-mathematics/Oid/Oid-4.0-r1.ebuild
+++ b/sci-mathematics/Oid/Oid-4.0-r1.ebuild
@@ -1,0 +1,62 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit java-pkg-2
+
+MY_PV="${PV//./-}"
+MY_P="${PN}${MY_PV}"
+
+DESCRIPTION="An interactive, extensible software system for experimenting with matroids"
+HOMEPAGE="http://userhome.brooklyn.cuny.edu/skingan/matroids/software.html"
+SRC_URI="
+	http://userhome.brooklyn.cuny.edu/skingan/matroids/${MY_P}.tar.gz -> ${P}.tar.gz
+	http://userhome.brooklyn.cuny.edu/skingan/matroids/${PN}UserManual${MY_PV}.pdf
+"
+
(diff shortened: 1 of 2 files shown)


sci-mathematics/Oid/Oid-4.0-r1.ebuild
```diff
@@ -0,0 +1,62 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
AndrewAmmerlaan, line 4:
```suggestion
EAPI=8
```

sci-mathematics/Oid/Oid-4.0-r1.ebuild
```diff
@@ -0,0 +1,62 @@
...
+	default
+}
+
+src_compile () {
+	ejavac -d classes @Oid/filelist.unix
+	ejavac -d classes @filelist
+}
+
+src_install () {
+	jar cef Oid Oid.jar -C classes . || die "Failed to create jar"
```
AndrewAmmerlaan, line 54: I am not super familiar with java packaging, but I believe there is some function in the eclasses to do this.

AndrewAmmerlaan: > I am not super familiar with java packaging, but I believe there is some function in the eclasses to do this.

`java-pkg-simple.eclass` is the eclass I was thinking of before. I think it should make most of the explicit phase declaration above unnecessary.

APN-Pucky: https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/java-pkg-simple.eclass#n372 here the eclass scans for any java files. But we already have lists than must(?) be used in `src_compile`.

AndrewAmmerlaan: Probably the issue here is that java-pkg-2-simple considers every single java file, whereas the current ebuild code uses only some subset of java files. 

If the eclass is not cooperating for this ebuild, then let's just keep the code as it is now. I don't know enough about java and java packaging to properly debug unusual packages.

PR #1241, opened 2024-02-16
sci-biology/libBigWig: add 0.4.7

APN-Pucky:
Please note that 0.4.4 builds fine, but fails test probed by `pkg-testing-tools`.


sci-biology/libBigWig/libBigWig-0.4.7.ebuild
```diff
@@ -0,0 +1,31 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
AndrewAmmerlaan, line 4:
```suggestion
EAPI=8
```

PR #1242, opened 2024-02-18
sci-physics/hoppet: new package, add 1.2.0

APN-Pucky:


sci-physics/hoppet/hoppet-1.2.0.ebuild
```diff
@@ -0,0 +1,41 @@
+# Copyright 2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
```
AndrewAmmerlaan, line 4: We should probably `inherit fortran-2` here.

sci-physics/hoppet/hoppet-1.2.0.ebuild
```diff
@@ -0,0 +1,35 @@
+# Copyright 2022 Gentoo Authors
```
AndrewAmmerlaan, line 1:
```suggestion
# Copyright 2022-2024 Gentoo Authors
```

PR #1243, opened 2024-02-27
sci-physics/cernlib: new package, add 2023.10.31.0

APN-Pucky:
The cernlib historically is a bit of a mess..., since there are still programs needing the 2006 version around.


sci-physics/cernlib/cernlib-2023.10.31.0.ebuild
```diff
@@ -0,0 +1,34 @@
...
+DEPEND="
+	${RDEPEND}
+	x11-misc/imake
+	x11-misc/makedepend
+	virtual/pkgconfig
+"
+
+src_install() {
+	cmake_src_install
+	# do not install cfortran.h, it conflicts with ROOT
```
AndrewAmmerlaan, line 32: Does it make sense to move this to a subdir instead? Or are these headers completely unused?

sci-physics/cernlib/cernlib-2023.10.31.0.ebuild
```diff
@@ -0,0 +1,34 @@
...
+	virtual/lapack
+	dev-lang/cfortran
+	x11-libs/xbae
+	net-libs/libnsl
+"
+DEPEND="
+	${RDEPEND}
+	x11-misc/imake
+	x11-misc/makedepend
+	virtual/pkgconfig
```
AndrewAmmerlaan, line 27: `pkgconfig` is a BDEPEND, those other two packages also suspiciously sound like the might be used during building.

sci-physics/cernlib/cernlib-2023.10.31.0.ebuild
APN-Pucky: I followed the https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=cernlib but had to remove the second one too.

sci-physics/cernlib/cernlib-2023.10.31.0.ebuild
APN-Pucky: I should check which of them are still required. They originally are from [cernlib-2006-r9.ebuild](https://gitlab.com/APN-Pucky/gentoo-hep-forge/-/blob/920955d8197ab8e6e1fd34ababd509a7fef9dd35/sci-physics/cernlib/cernlib-2006-r9.ebuild) that I found somewhere on the web, but since they (seemingly) switched from `imake` to `cmake` I should check what is still really required.

sci-physics/cernlib/cernlib-2023.10.31.0.ebuild
APN-Pucky: The headers are actually the same (up to version differences), so I removed it in `src_prepare` now already, since it exists already from `dev-lang/cfortran`.

AndrewAmmerlaan (approved):

sci-physics/cernlib/cernlib-2023.10.31.0.ebuild
AndrewAmmerlaan: Great, Thanks :+1: 

I think we should add a little comment here though, because this is the kind of thing that in x years time is going to be confusing again.

PR #1245, opened 2024-02-27
dev-python/uhi: new package, add 0.4.0

APN-Pucky:


dev-python/uhi/uhi-0.4.0.ebuild
```diff
@@ -0,0 +1,27 @@
+EAPI=8
+
+PYTHON_COMPAT=( python3_{10..11} )
+DISTUTILS_USE_PEP517=hatchling
+export SETUPTOOLS_SCM_PRETEND_VERSION=${PV}
```
AndrewAmmerlaan, line 5: `export` should not be necessary here

AndrewAmmerlaan: hmm, maybe I am wrong then.

APN-Pucky: It worked without the export, maybe it is just copied from the docs https://projects.gentoo.org/python/guide/distutils.html or other ebuilds with the export.

PR #1246, opened 2024-02-27
dev-python/awkward-cpp: new package, add 28

APN-Pucky:


dev-python/awkward-cpp/awkward-cpp-28.ebuild
```diff
@@ -0,0 +1,25 @@
+EAPI=8
```
AndrewAmmerlaan, line 1:
```suggestion
EAPI=8

```

Another nit, but unwritten convention states that there is an empty line after `EAPI`

PR #1251, opened 2024-02-27
sci-physics/thepeg: new package, add 2.3.0

APN-Pucky:


APN-Pucky:
This package does already exist in `::gentoo` but the new version depends on `sci-physics/rivet` that is only available in `::sci`. I'd prefer to move rivet to `::gentoo`, but I thought I'd ask here first what is better?

AndrewAmmerlaan:
> This package does already exist in `::gentoo` but the new version depends on `sci-physics/rivet` that is only available in `::sci`. I'd prefer to move `rivet` to `::gentoo`, but I thought I'd ask here first what is better?

The `rivet` ebuild looks fine (haven't tested it though), so I think we can move this to `::gentoo`. We should check if we can add `python3_12` though, and it looks like it should inherit `fortran-2.eclass` (we can then drop the explicit dependency on `virtual/fortran`.

sci-physics/thepeg/thepeg-2.3.0.ebuild
```diff
@@ -0,0 +1,108 @@
...
+CDEPEND="
+	sci-libs/gsl:0=
+	emacs? ( >=app-editors/emacs-23.1:* )
+	fastjet? ( sci-physics/fastjet:0= )
+	rivet? ( sci-physics/rivet:3=[hepmc3] )
+	hepmc3? ( sci-physics/hepmc:3= )
+	lhapdf? ( >=sci-physics/lhapdf-6.0:0= )
+	zlib? ( sys-libs/zlib:0= )"
+DEPEND="${CDEPEND}
+	sci-libs/gsl
```
AndrewAmmerlaan, line 38:
```suggestion
	sci-libs/gsl:=
```

sci-physics/thepeg/thepeg-2.3.0.ebuild
```diff
@@ -0,0 +1,108 @@
...
+src_test() {
+	emake LHAPATH="${DISTDIR}" check
+}
+
+src_install() {
+	default
+	use emacs && elisp-install ${PN} lib/ThePEG.el{,c}
+	use java && java-pkg_newjar java/ThePEG.jar
+
+	cat <<-EOF > "${T}"/50${PN}
```
AndrewAmmerlaan, line 94:
```suggestion
	cat <<-EOF > "${T}"/50${PN} || die
```

sci-physics/thepeg/thepeg-2.3.0.ebuild
```diff
@@ -0,0 +1,108 @@
...
+	zlib? ( sys-libs/zlib:0= )"
+DEPEND="${CDEPEND}
+	sci-libs/gsl
+	java? ( virtual/jdk:1.8 )
+	test? (
+		sys-process/time
+		dev-libs/boost
+	)"
+RDEPEND="${CDEPEND}
+	java? ( virtual/jre:1.8 )"
```
AndrewAmmerlaan, line 45:
```suggestion
	java? ( virtual/jre:1.8 )
"
```

Can it be used with newer java? 8 is quite old.

APN-Pucky: I haven't used it with it with java, this is still from the ebuilds in `::gentoo`.

APN-Pucky:
> > This package does already exist in `::gentoo` but the new version depends on `sci-physics/rivet` that is only available in `::sci`. I'd prefer to move `rivet` to `::gentoo`, but I thought I'd ask here first what is better?
> 
> The `rivet` ebuild looks fine (haven't tested it though), so I think we can move this to `::gentoo`. We should check if we can add `python3_12` though, and it looks like it should inherit `fortran-2.eclass` (we can then drop the explicit dependency on `virtual/fortran`.

`rivet` even works well in prefixed gentoo. I added a docker image with startprefix entrypoint to also test prefix compatibility recently.

APN-Pucky:
Moved to https://github.com/gentoo/gentoo/pull/36341

PR #1252, opened 2024-02-27
dev-python/nptyping: update maintainers

APN-Pucky:
Repology updates broke by this. See https://github.com/repology/repology-updater/commit/688fc91d1092e3c7b7b5c79bd03935bb54e872b1

I am surprissed that `pkgcheck scan` reports `PkgBadlyFormedXml` only as a warning. Should we promote it to an error in the CI?

### DIFF
--- a/dev-python/nptyping/metadata.xml
+++ b/dev-python/nptyping/metadata.xml
@@ -9,6 +9,7 @@
 		<email>gentoo@chymera.eu</email>
 		<name>Horea Christian</name>
 	</maintainer>
+	<upstream>
 		<remote-id type="pypi">nptyping</remote-id>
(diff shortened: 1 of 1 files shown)


AndrewAmmerlaan:
See Bug https://bugs.gentoo.org/922385

@TheChymera I did CC you on that bug... And please do use `pkgcheck` and `pkgdev push`, you'd have noticed this problem if you'd used these tools.

> I am surprissed that pkgcheck scan reports PkgBadlyFormedXml only as a warning. Should we promote it to an error in the CI?

Yes definitely, it causes fatal failures both in our automated checks and in Repology. Could you open an issue for this over at https://github.com/pkgcore/pkgcheck?

PR #1162, opened 2022-07-24
sci-physics/spheno: tc-export AR

APN-Pucky:
Signed-off-by: Alexander Puck Neuwirth <alexander@neuwirth-informatik.de>

PR #1254, opened 2024-03-12
sci-physics/spheno: add 4.0.5

APN-Pucky:


sci-physics/spheno/spheno-3.3.8-r1.ebuild
```diff
@@ -3,29 +3,26 @@
...
-
 PATCHES=( "${FILESDIR}"/${P}-gfortran.patch )
 
 src_compile() {
 	# single thread force needed since fortan mods depend on each other
-	export MAKEOPTS=-j1
+	export MAKEOPTS="$MAKEOPTS -j1"
 	tc-export FC AR
 	emake AR="${AR}" F90="${FC}"
 }
```
AndrewAmmerlaan, line 28:
```
src_compile() {
	# single thread force needed since fortan mods depend on each other
	tc-export FC AR
	emake -j1 AR="${AR}" F90="${FC}"
}
```

AndrewAmmerlaan: It also seems a bit double to both `tc-export` and set the variables explicitly in the `emake` command.

APN-Pucky: Normally `fortran-2` would set it but we need to overwrite `F90` by `FC`.

APN-Pucky: But in this case it was a single PR adding AR https://github.com/gentoo/sci/pull/1162
Maybe it should be `${AR}` instead of `$(AR)` in the patch?

APN-Pucky: Either way I am testing it now.

PR #1256, opened 2024-03-16
sci-libs/gvar: new package, add 13.0.2

APN-Pucky:


sci-libs/gvar/gvar-13.0.2.ebuild
```diff
@@ -0,0 +1,20 @@
...
+
+DISTUTILS_EXT=1
+PYTHON_COMPAT=( python3_{10..12} )
+DISTUTILS_USE_PEP517=setuptools
+inherit distutils-r1 pypi
+
+DESCRIPTION="Gaussian random variables."
+HOMEPAGE="https://github.com/gplepage/gvar"
+
+LICENSE="MIT"
```
AndrewAmmerlaan, line 11: `LICENSE.txt` indicates `GPL-3`

sci-libs/gvar/gvar-13.0.2.ebuild
```diff
@@ -0,0 +1,20 @@
...
+SLOT="0"
+KEYWORDS="~amd64 ~arm64"
+
```
AndrewAmmerlaan, line 14: The tarball includes some unittests, can we add support for that in the ebuild?

```
distutils_enable_tests unittest
```

sci-libs/gvar/gvar-13.0.2.ebuild
```diff
@@ -0,0 +1,20 @@
...
+RDEPEND="
+	>=dev-python/cython-0.17[${PYTHON_USEDEP}]
```
AndrewAmmerlaan, line 16: I think (but I'm not 100% sure), that a dependency on cython is added automatically when `DISTUTILS_EXT` is set.

APN-Pucky: `::gentoo` has > 20 packages with both https://github.com/search?q=repo%3Agentoo%2Fgentoo+DISTUTILS_EXT+cython&type=code so I'd expect it is needed.

AndrewAmmerlaan: The eclass has this though:
```
		if [[ ${DISTUTILS_EXT} ]]; then
			packages+=(
				dev-python/cython
			)
		fi
```

APN-Pucky: Ok, stuff like numpy then just has it to filter versions to `>cython-3.0.0`, but by now that is also satisfied by all cythons.

AndrewAmmerlaan: Aah nevermind, this seems to be something to do with the logging, not with the dependencies: https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=677b69f23cb3c0e021a21b8a775b1da3143fd564

APN-Pucky: Wait, that is just a debug print what you showed, right?

AndrewAmmerlaan: Yeah it prints the version of cython, but never adds a dependency on it. So I was wrong and we do need to add the cython dependency here.

I wonder if this is intentional though.

sci-libs/gvar/gvar-13.0.2.ebuild
APN-Pucky: tests passed.

APN-Pucky:
I added `vegas` to this PR which needs `gvar`.

sci-libs/vegas/vegas-6.0.1.ebuild
```diff
@@ -0,0 +1,22 @@
...
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~arm64"
+
+RDEPEND="
+	>=dev-python/cython-0.17[${PYTHON_USEDEP}]
+	>=dev-python/numpy-1.16[${PYTHON_USEDEP}]
+	>=sci-libs/gvar-13.0.2[${PYTHON_USEDEP}]
+"
+BDEPEND="${RDEPEND}"
```
AndrewAmmerlaan, line 20: Ooh right, this will add `cython` as `BDEPEND`. But are `numpy` and `gvar` also `BDEPEND? They should only be BDEPEND if they are executed by the `setup.py`.

AndrewAmmerlaan (approved):

PR #1257, opened 2024-03-16
dev-python/awkward: new package, add 2.5.2

APN-Pucky:


dev-python/awkward/awkward-2.5.2.ebuild
```diff
@@ -0,0 +1,27 @@
...
+SLOT="0"
+KEYWORDS="~amd64"
+
+RDEPEND="
+	>=dev-python/numpy-1.18.0[${PYTHON_USEDEP}]
+	~dev-python/awkward-cpp-28[${PYTHON_USEDEP}]
+	dev-python/packaging[${PYTHON_USEDEP}]
+"
+BDEPEND="
+	${RDEPEND}
```
AndrewAmmerlaan, line 21: Probably not all the RDEPENDs are BDEPENDs, packaging sounds like it might be BDEPEND.

dev-python/awkward/awkward-2.5.2.ebuild
```diff
@@ -0,0 +1,27 @@
...
+	~dev-python/awkward-cpp-28[${PYTHON_USEDEP}]
+"
+DEPEND="${RDEPEND}"
+BDEPEND="
+	dev-python/packaging[${PYTHON_USEDEP}]
+	dev-python/hatch-vcs[${PYTHON_USEDEP}]
+	dev-python/hatch-fancy-pypi-readme[${PYTHON_USEDEP}]
+"
+
+# disabled since it needs jax and numba
```
AndrewAmmerlaan, line 26: Does `EPYTEST_IGNORE` and/or `EPYTEST_DESELECT` help to run a subset of tests that we do have the dependencies for?

AndrewAmmerlaan (approved):

dev-python/awkward/awkward-2.5.2.ebuild
APN-Pucky: I have/had an ebuild for jax that I realixed was very broken, but awkward was happy with it since it only uses it's typing of arrays. I'll see if the tests are grouped in awkward.

APN-Pucky: Ok they have ~3000 test files, I'll check which fail.

PR #1259, opened 2024-03-17
dev-python/{hist,histoprint}: new package

APN-Pucky:


dev-python/histoprint/histoprint-2.4.0.ebuild
```diff
@@ -0,0 +1,21 @@
...
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64"
+
+RDEPEND="
+	dev-python/numpy[${PYTHON_USEDEP}]
+	>=dev-python/click-7.0.0[${PYTHON_USEDEP}]
+	>=dev-python/uhi-0.2.1[${PYTHON_USEDEP}]
+"
+BDEPEND="${RDEPEND}"
```
AndrewAmmerlaan, line 19: Usually we don't need the rdeps at build time for python packages, unless some building script imports them. Is that the case here?

AndrewAmmerlaan (approved):

PR #1264, opened 2024-04-02
sci-physics/formcalc: fix chmod cannot access

APN-Pucky:

Before I got 
```
chmod: cannot access '/usr/share/Mathematica/Applications/FormCalc-9.10/COPYING': No such file or directory
```

### DIFF
--- a/sci-physics/formcalc/formcalc-9.9-r1.ebuild
+++ b/sci-physics/formcalc/formcalc-9.9-r1.ebuild
@@ -38,10 +38,9 @@ src_compile() {
 
 src_install() {
 	MMADIR=/usr/share/Mathematica/Applications
-	dodir ${MMADIR}/${MY_P}
-	mv "${WORKDIR}/${MY_P}" "${ED}${MMADIR}" || die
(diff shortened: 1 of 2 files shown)


AndrewAmmerlaan (commented):
This is a bit strange because according to the install functions reference fperms already accounts for ED

sci-physics/formcalc/formcalc-9.9-r1.ebuild
```diff
@@ -44,7 +44,7 @@ src_install() {
 	doins -r "${S}"
 	# Copy executable, etc. permissions
 	for f in $(find * ! -type l); do
-		fperms --reference="${S}/$f" ${MMADIR}/${MY_P}/$f
+		fperms --reference="${S}/$f" "${D}"/${MMADIR}/${MY_P}/$f
 	done
```
AndrewAmmerlaan, line 48: How about all of this `insinto`/`doins` and then restoring permissions, we just use `mv` or `cp`?

AndrewAmmerlaan (approved):

PR #1265, opened 2024-04-08
sci-physics/cernlib: split free and non-free

APN-Pucky:

Open question: is it possible to provide a LICENSE that depends on the use flags? `S` can't have use flags, so I add a custom unpack.


AndrewAmmerlaan:
> Open question: is it possible to provide a LICENSE that depends on the use flags? S can't have use flags, so I add a custom unpack.

Yes! You can have USE conditionals in LICENSE.

We can't have them in S, but what you can do is:
```
src_unpack(){
     if use flag; then
           S=....
     else
           S=
     default
}
```

APN-Pucky:
> Yes! You can have USE conditionals in LICENSE.

That's cool. Unfortunately cernlib is a historic mess (from their readme, locked behind a cern-login):

> Licensing of these sources might be subject to restrictions. This needs further clarification. Please do NOT distribute.
Some information about the licensing situation can be found in the contrib folder.

So I'd rather keep it `all-rights-reserved` for both, or I could try to list all of the licenses individually. On the other hand one of the developers list GPL+ and LGPLv2+ license https://github.com/andriish/HEPrpms/blob/bb918c3706857a8ee03a3131dd38a3ec5dcd08e7/cernlib/2023.10.31.0/cernlib.spec#L47

AndrewAmmerlaan:
> Please do NOT distribute.

We should also have `RESTRICT="mirror"` in the ebuild then. Not that it really matters since nothing in `::sci` is mirrored anyway, but it would still be good to put in there for correctness.

AndrewAmmerlaan (approved):

PR #1266, opened 2024-04-10
Use fortran eclass

APN-Pucky:
Also some minor changes to make `pkgcheck` happier.

- [x] I should test the packages again before merging.

### DIFF
--- a/sci-physics/collier/metadata.xml
+++ b/sci-physics/collier/metadata.xml
@@ -16,7 +16,4 @@
   to the scalar one-loop functions of FF and in addition provides 
   the 2-, 3-, and 4-point tensor coefficient functions.
   </longdescription>
-  <use>
-    <flag name="static">Build static library</flag>
-  </use>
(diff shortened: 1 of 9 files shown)


sci-physics/collier/metadata.xml
```diff
@@ -17,6 +17,6 @@
   the 2-, 3-, and 4-point tensor coefficient functions.
   </longdescription>
   <use>
-    <flag name="static">Build static library</flag>
+    <flag name="static-libs">Build static library</flag>
```
AndrewAmmerlaan, line 20: This flag is already described globally (in `::gentoo`), so there is no need to describe it locally here.

AndrewAmmerlaan (approved):

sci-physics/collier/metadata.xml
APN-Pucky: That explains why `pkgcheck` did not complain before.

PR #1267, opened 2024-04-17
sci-physics/{oneloop,cuttools,openloops}: add {3.6, 1.8.2, 2.1.3}

APN-Pucky:
At least two imminent bigger problems here (before I clean it up further):

- kind_types.f90 is from openloops2, which I'd link to PR sooner or later, since they modified oneloop I'd include these changes here. Alternatively, openloops2 could just use their own version of oneloop, but that seems a bit wrong (unless they perform many more changes).
- I dropped the 2020.07.31, since I found a the latest official version 3.6 and therefore the 2020.07.31 version is just from the bitbucket and should be called 3.6_p2020.07.31, IMO. 
- Maybe I should just throw the openloops PR inside of here too? 

How does Gentoo typically handle these (rare science mess) cases?


AndrewAmmerlaan:
> How does Gentoo typically handle these (rare science mess) cases?

Do I understand correctly that this bundles a modified version of a dependency? 

In this case I would say it depends on which modifications they made and how many. If the changes are generally useful I think we can patch the dependency in its own package. Otherwise I think the bundled dependency should be in SRC_URI.

sci-physics/oneloop/oneloop-3.6.ebuild
```diff
@@ -0,0 +1,97 @@
...
+SRC_URI="http://helac-phegas.web.cern.ch/helac-phegas/tar-files/${MY_P}.tgz"
+S="${WORKDIR}/${MY_P}"
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="+dpkind +qpkind qpkind16 dpkind16 qdcpp ddcpp mpfun90 arprec tlevel cppintf kind_types"
+REQUIRED_USE="
+	?? ( dpkind dpkind16 ddcpp )
+	?? ( qpkind qpkind16 qdcpp )
```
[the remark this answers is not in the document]
APN-Pucky: I don't think so one is quad and other double precision.

APN-Pucky:
> > How does Gentoo typically handle these (rare science mess) cases?
> 
> Do I understand correctly that this bundles a modified version of a dependency?
> 
> In this case I would say it depends on which modifications they made and how many. If the changes are generally useful I think we can patch the dependency in its own package. Otherwise I think the bundled dependency should be in SRC_URI.

I guess I'll add the other stuff here too, so one can explicitly check that it works with the final package.

sci-physics/oneloop/oneloop-3.6.ebuild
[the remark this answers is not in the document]
AndrewAmmerlaan: :O 

I must have been very tired when I was looking at this, sorry.

sci-physics/cuttools/cuttools-1.8.2.ebuild
```diff
@@ -0,0 +1,50 @@
+# Copyright 2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit toolchain-funcs
+
+DESCRIPTION="Computing 1-loop amplitudes at the integrand level"
+HOMEPAGE="https://www.ugr.es/~pittau/CutTools/"
+SRC_URI="http://www.ugr.es/~pittau/CutTools/${PN}_v${PV}.tar.gz"
```
AndrewAmmerlaan, line 10:
```suggestion
SRC_URI="https://www.ugr.es/~pittau/CutTools/${PN}_v${PV}.tar.gz"
```

sci-physics/cuttools/cuttools-1.8.2.ebuild
```diff
@@ -0,0 +1,50 @@
...
+"
+RDEPEND="${DEPEND}"
+BDEPEND="
+	virtual/fortran
+"
+src_prepare() {
+	default
+	sed -i 's/^ALL =.*$/ALL = $(CTS)/' src/makefile ||  die
+	use dummy && cp "${FILESDIR}"/mpnumdummy.f90 src/cts/mpnumdummy.f90
+	use dummy && sed -i 's/CTS =/CTS = mpnumdummy.o/' src/makefile
```
AndrewAmmerlaan, line 30: These should `|| die`, so we need a proper if statement here.

sci-physics/cuttools/cuttools-1.8.2.ebuild
```diff
@@ -0,0 +1,50 @@
...
+}
+
+src_compile() {
+	emake -j1 FFLAGS="${FFLAGS} -I${ESYSROOT}/usr/include -fPIC -std=legacy"
+	tc-export AR CXX
+	cd includects
```
AndrewAmmerlaan, line 36:
```suggestion
	cd includects || die
```

sci-physics/cuttools/cuttools-1.8.2.ebuild
```diff
@@ -0,0 +1,50 @@
...
+	${AR} -x libcts.a
+	${CXX} -shared *.o -o lib${PN}.so
```
AndrewAmmerlaan, line 38:
```suggestion
	${AR} -x libcts.a || die
	${CXX} ${CXXFLAGS} -shared *.o -o lib${PN}.so || die
```

sci-physics/cuttools/cuttools-1.8.2.ebuild
```diff
@@ -0,0 +1,50 @@
...
+
+src_install() {
+	cd includects
+	dolib.so lib${PN}.so
+	cd ..
+	mv includects ${PN}
+	rm ${PN}/*.a
+	rm ${PN}/*.so
+	rm ${PN}/*.o
+	doheader -r ${PN}
```
AndrewAmmerlaan, line 49:
```suggestion
	cd includects || die
	dolib.so lib${PN}.so
	cd .. || die
	mv includects ${PN} || die
	rm ${PN}/*.a || die
	rm ${PN}/*.so || die
	rm ${PN}/*.o || die
	doheader -r ${PN}
```

sci-physics/cuttools/cuttools-1.8.2.ebuild
```diff
@@ -0,0 +1,50 @@
...
+
+IUSE="+dummy"
+DEPEND="
+	sci-physics/qcdloop
+	sci-physics/oneloop[dpkind,qpkind16,-qpkind,-tlevel]
+"
+RDEPEND="${DEPEND}"
+BDEPEND="
+	virtual/fortran
+"
```
AndrewAmmerlaan, line 25: `fortran-2.eclass` :P

sci-physics/openloops/openloops-2.1.2.ebuild
```diff
@@ -0,0 +1,106 @@
...
+DEPEND="${RDEPEND}"
+
+PATCHES=(
+	"${FILESDIR}/${P}-python3.12.patch"
+	"${FILESDIR}/${P}-ldflags.patch"
+)
+
+src_prepare() {
+	default
+	mv openloops.cfg.tmpl openloops.cfg
```
AndrewAmmerlaan, line 39:
```suggestion
	mv openloops.cfg.tmpl openloops.cfg || die
```

sci-physics/openloops/openloops-2.1.2.ebuild
```diff
@@ -0,0 +1,106 @@
...
+link_flags = ${LDFLAGS} -I${ESYSROOT}/usr/include/ -I${ESYSROOT}/usr/include/cuttools -lcollier
+cc = $(tc-getCC)
+cxx = $(tc-getCXX)
+fortran_compiler = $(tc-getFC)
+process_src_dir = ${EPREFIX}/opt/${MY_P}/process_src/
+process_obj_dir = ${EPREFIX}/opt/${MY_P}/process_obj/
+process_lib_dir = ${EPREFIX}/opt/${MY_P}/proclib/
+release = $PV
+import_env = @all
+EOF
```
AndrewAmmerlaan, line 60: Here you can also use `-EOF` instead so you can use indentation

sci-physics/cuttools/cuttools-1.8.2.ebuild
APN-Pucky: I am going to patch that now into my pkgcheck...

AndrewAmmerlaan (approved):

PR #1268, opened 2024-04-23
sci-physics/cernlib: Fix test failures

APN-Pucky:
Login required issue link: https://gitlab.cern.ch/DPHEP/cernlib/cernlib/-/issues/43

> APN-Pucky: Both 2024.04.15.0-beta and 2023.10.31.0 pass without failures for me now, but only with -DBUILD_SHARED_LIBS=OFF and -O0. Going to -O2 fails some testzebfzs or going to shared fails the testkernnumtest.

### DIFF
--- a/sci-physics/cernlib/cernlib-2023.10.31.0-r2.ebuild
+++ b/sci-physics/cernlib/cernlib-2023.10.31.0-r2.ebuild
@@ -1,7 +1,7 @@
 EAPI=8
 
 CMAKE_MAKEFILE_GENERATOR="emake"
-inherit cmake fortran-2
+inherit cmake fortran-2 flag-o-matic
 
 DESCRIPTION="CERN program library for High Energy Physics"
 HOMEPAGE="https://cernlib.web.cern.ch/cernlib/"
@@ -16,7 +16,8 @@ LICENSE="
 "
 SLOT="0"
 KEYWORDS="~amd64"
-IUSE="+free"
+# static-libs as default since otherwise test fail...
+IUSE="+free +static-libs"
 RESTRICT="mirror"
 
 RDEPEND="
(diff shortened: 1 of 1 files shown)


sci-physics/cernlib/cernlib-2023.10.31.0-r2.ebuild
```diff
@@ -59,6 +60,11 @@ src_configure() {
 	# docs follow rpm like spliting into packages cernlib, cernlib-devel, etc.
 	# we move them into a folder that agrees with gentoo doc structure.
 	sed -i "s#/doc/#/doc/${PF}/#g" CMakeLists.txt || die
+	# with -O2 some tests fail
+	append-flags -O0
```
AndrewAmmerlaan, line 64: This should be `replace-flags` as per https://devmanual.gentoo.org/ebuild-writing/functions/src_compile/build-environment/index.html

What about `-O1` does this also cause failure?

AndrewAmmerlaan: Hmm, then maybe `filter-flags` is what we want here. AFAIK `-O0` is the default, so if we filter out any other `-Ox` flags then it should work.

sci-physics/cernlib/cernlib-2023.10.31.0-r2.ebuild
```diff
@@ -61,7 +61,8 @@ src_configure() {
 	# we move them into a folder that agrees with gentoo doc structure.
 	sed -i "s#/doc/#/doc/${PF}/#g" CMakeLists.txt || die
 	# with -O2 some tests fail
-	filter-flags -O2 -O3 -Os
+	# let upstream decide on optimization (-O0) since code is fragile
+	filter-flags -O1 -O2 -O3 -Os
```
AndrewAmmerlaan, line 65: And `-Oz -Og -Ofast` too I guess

AndrewAmmerlaan (approved):

PR #1273, opened 2024-06-11
sci-mathematics/pulp: Added pulp version 2.8.0 + python3.12

lima-limon-inc:
Updated to Pulp verison 2.8.0 and added python3.12.

In theory, according to their Github Actions, python 3.12 is supported

Source: https://github.com/coin-or/pulp/blob/master/.github/workflows/pythonpackage.yml#L11


AndrewAmmerlaan:
Thanks :+1: 

> In theory, according to their Github Actions, python 3.12 is supported

The way to know for sure is `FEATURES=test USE=test emerge -1 sci-mathematics/pulp`

PR #1275, opened 2024-06-16
sci-biology/imagej:  ant-core deprecated

IAHM-COL:
Signed-off-by: IAHMCOL <IAHMCOL@thejabberwocky.net>

This commit removes dev-java/ant-core from dependencies This has been deprecated, and it gets automatically used with ant2

Latest updates had depclean  dev-java/ant-core, and the packages still builds after dependency removal

see:
https://packages.gentoo.org/packages/dev-java/ant-core

### DIFF
--- a/sci-biology/imagej/imagej-1.54i-r1.ebuild
+++ b/sci-biology/imagej/imagej-1.54i-r1.ebuild
@@ -50,7 +50,6 @@ DEPEND="
 "
 
 BDEPEND="
-	dev-java/ant-core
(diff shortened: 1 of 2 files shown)


IAHM-COL:
pkgcheck on github failed in ways I do not understand.
I ran 'pkgcheck scan` before pushing, and it comes clean.

AndrewAmmerlaan:
Makes sense, Thanks :+1: 

pkgcheck complaint is about another package.

PR #1279, opened 2024-06-27
sci-physics/sherpa: new package, add 3.0.0_beta1, 9999

APN-Pucky:


AndrewAmmerlaan (approved):
Looks good to me.

Maybe it makes sense to test these against python3.13 now so it doesn't have to be done later.

APN-Pucky:
> Looks good to me.
> 
> Maybe it makes sense to test these against python3.13 now so it doesn't have to be done later.

I don't have p3.13 yet (compiling...). Does it make sense to test it on the beta py3.13 releases? I'd think waiting for the first non-beta release is better for testing the (these uncommon) packages.

AndrewAmmerlaan:
> > Looks good to me.
> > 
> > Maybe it makes sense to test these against python3.13 now so it doesn't have to be done later.
> 
> I don't have py3.13 yet (compiling...). Does it make sense to test it on the beta py3.13 releases? I'd think waiting for the first non-beta release is better for testing the (these uncommon) packages.

The beta should not change functionality wise any more, but I don't mind postponing till the proper release if you prefer.

PR #1288, opened 2024-07-21
sci-libs/rdkit: version bump 2024.03.4

mschubert:


sci-libs/rdkit/rdkit-2024.03.4.ebuild
```diff
@@ -54,5 +53,6 @@ src_configure() {
 }
 
 src_test() {
-	RDBASE="${WORKDIR}/${PN}-Release_2021_09_4_build" cmake_src_test
+	rsync -aur "${S}_build/rdkit" "${S}" || die
```
mschubert, line 56: I'm not sure about the rsync here, but it's the only way I could get the python import test to work

AndrewAmmerlaan: Instead of copying the build stuff into the `PYTHONPATH`, simply set `PYTHONPATH` to where the built stuff is.

sci-libs/rdkit/rdkit-2024.03.4.ebuild
```diff
@@ -33,12 +33,11 @@ RDEPEND="dev-libs/boost
 DEPEND="${RDEPEND}"
 
 PATCHES=(
-	"${FILESDIR}/${PN}-2021.09.4-find-rapidjson.patch"
+	"${FILESDIR}/find-rapidjson.patch"
```
AndrewAmmerlaan, line 36: Patches should contain the `PN` and usually also the first version where it was introduced.

sci-libs/rdkit/rdkit-2024.03.4.ebuild
mschubert: The problem here seems to be that the build directory contains all `.so` files but no `.py` files. So setting the `PYTHONPATH` to the build directory will produce an error that it's not a package

AndrewAmmerlaan: Does `CMAKE_IN_SOURCE_BUILD=1` help?

mschubert: `CMAKE_IN_SOURCE_BUILD=1` did indeed solve it

Nowa-Ammerlaan:
Thanks :+1:

PR #1292, opened 2024-10-19
Bump and add scikit-hep packages and enable py3.13

APN-Pucky:
New packages are

- mplhep
- mplhep-data
- decaylanguage
- resample
- hepstats

and pulled in from updated asdf

- asdf-standard
- asdf-transform-schemas

All ebuilds here are for at least `python3_{11..13}` (except `mplhep` where I wait for https://bugs.gentoo.org/941848). Nothing requires a review IMO, but I will wait for possible comments before merging all of it. Some commits shall still be merged to make it all a bit nicer (drop py3.10/metadata).

One question:

There is a scikit-hep package that groups all their packages: https://github.com/scikit-hep/scikit-hep/blob/main/requirements.txt but versioned by YYYY-MM-DD with ~fixed versions. I do not want to keep track with all their monthly versioning, nor do I like to fix the versions. What would be the best Gentoo way to handle this? Maybe a loose `virtual/scikit-hep-0` that pulls them all without caring about the version. If there is no nice solution I do not mind not offering that collection-package.


dev-python/asdf-transform-schemas/asdf-transform-schemas-0.5.0.ebuild
```diff
@@ -0,0 +1,26 @@
...
+DESCRIPTION="ASDF schemas for validating transform tags."
+HOMEPAGE="
+	https://github.com/asdf-format/asdf-transform-schemas/
+	https://pypi.org/project/asdf-transform-schemas/
+"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64"
+restrict="test" # needs asdf-astropy
```
Nowa-Ammerlaan, line 20:
```suggestion
RESTRICT="test" # needs asdf-astropy
```

sci-physics/decaylanguage/decaylanguage-0.18.5.ebuild
```diff
@@ -0,0 +1,29 @@
...
+	>=dev-python/numpy-1.12[${PYTHON_USEDEP}]
+	>=dev-python/pandas-0.22[${PYTHON_USEDEP}]
+	>=sci-physics/particle-0.21.0[${PYTHON_USEDEP}]
+	>=sci-physics/hepunits-2.0.0[${PYTHON_USEDEP}]
+	>=dev-python/plumbum-1.7.0[${PYTHON_USEDEP}]
+"
+BDEPEND="
+	${RDEPEND}
+	dev-python/hatch-vcs[${PYTHON_USEDEP}]
+"
```
Nowa-Ammerlaan, line 27: Are the RDEPENDs really BDEPENDs here? Probably this should be DEPEND=RDEPEND instead.

sci-physics/hepstats/hepstats-0.8.1.ebuild
```diff
@@ -0,0 +1,28 @@
...
+	dev-python/asdf[${PYTHON_USEDEP}]
+	dev-python/numpy[${PYTHON_USEDEP}]
+	dev-python/pandas[${PYTHON_USEDEP}]
+	dev-python/scipy[${PYTHON_USEDEP}]
+	dev-python/tqdm[${PYTHON_USEDEP}]
+	dev-python/uhi[${PYTHON_USEDEP}]
+"
+BDEPEND="
+	${RDEPEND}
+"
```
Nowa-Ammerlaan, line 28: And here too

APN-Pucky: Thanks, I'll review the `*DEPEND`s.

APN-Pucky: Since they are (mostly) 100% python packages (No C code or dependencies), I followed https://projects.gentoo.org/python/guide/basic.html#dependencies-in-python-packages and just put them as `RDEPEND`. I tested that they still build, by uninstalling the `RDEPEND`s and then run `ebuild ... install`.

Nowa-Ammerlaan (approved):

PR #1293, opened 2024-10-30
sci-libs/torchvision: add 0.19.1, drop olds

littlewu2508:
Enable tests.

torchvision should be rebuilt if caffe2 turns on/off cuda or rocm.
So adding cuda/rocm to USE flag and dependencies to handle this.

Reference: https://bugs.gentoo.org/680496


Nowa-Ammerlaan:
Thanks :+1:

PR #1294, opened 2024-11-07
sci-*/*: remove dead java-ant-2 eclass

APN-Pucky:
- [x] Test if the packages still compile 
- [x] Use eant instead of ant 
- [x] Patch java source/target version


APN-Pucky:
Still Failing:
sci-biology/fastqc: source/target options 6 is no longer supported Use 7 or later
sci-biology/trimmomatic: source/target options 5 is no longer supported Use 7 or later
sci-biology/imagej: source/target options 6 is no longer supported Use 7 or later

Nowa-Ammerlaan (commented):
I don't know much about java packaging, but note that some of these packages are already masked because the dependencies no longer resolve: https://github.com/gentoo/sci/blob/master/profiles/package.mask

Nowa-Ammerlaan:
> even with explicit version args, so probably the ant file needs to be patched...

If you want to you could check if there is a newer version available that supports newer java.

Otherwise I'd just drop these packages, java 5/6 is very old. In theory you maybe could still compile them by eselecting dev-java/openjdk:8, but if upstream is gone then it is probably not worth keeping.

APN-Pucky:
imagej is not only biology and generally usefull, also that version is of September this year, so that can be kept. I think they all work more recent java versions (usually backwards compatible), but they need patches.

Nowa-Ammerlaan:
> Also I am not sure where ant-core went, but it is still available through some other/renamed package (#1275), so the masking could probably be undone after the changes.

According to the `package.deprecated` entry ant-core was merged into ant itself: https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=90d27074d2d374159463e6adebedf96658cbdb50
 
> I just do not know how to "really" test the packages (and I frankly do not care too much about sci-biology :))

Personally, I also don't care about the individual packages. What we really need is some biologist who does care about sci-biology to help maintain all of these packages. I do care about the overall usability of this repository though, and global scope errors and unresolvable dependencies tend to interfere not just with the offending package but with all the other packages in the repository as well. For that reason I usually just remove them.

APN-Pucky:
I bumped them now. All three compile and for imageJ only I checked that the GUI works.

Nowa-Ammerlaan (approved):
Looks good, Thanks :+1: 

You can also drop the now obsolete entries from `package.mask`

sci-biology/imagej/imagej-1.54i-r1.ebuild
```diff
@@ -65,6 +65,9 @@ src_prepare() {
 	if ! use debug ; then
 		sed -i 's: debug="on">: debug="off">:' "${S}"/build.xml || die
 	fi
+	sed -i 's/source="1\.6"/source="1.7"/g' build.xml || die
+	sed -i 's/target="1\.6"/target="1.7"/g' build.xml || die
```
vaukai, line 69: Better delete `-source` / `-target` attributes completely from alll `javac` elements of all `build.xml` files and use the `ant.build.javac.source` `ant.build.javac.target` options as done in [dev-java/c3p0](https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-java/c3p0/c3p0-0.9.5.5-r2.ebuild#n52) or in [dev-java/lucene](https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-java/lucene/lucene-2.4.1-r5.ebuild#n50).

APN-Pucky: Thanks. Done.

Nowa-Ammerlaan (approved):

sci-biology/imagej/imagej-1.54i-r1.ebuild
vaukai: > Thanks. Done.

Sorry, forgot to mention, pls avoid using `jre-1.7:*` / `jdk-1.7:*`. It is no longer supported by Java 21.
https://wiki.gentoo.org/wiki/Java_Developer_Guide#JDK.2FJRE
Also, the support for `EAPI 7` [will be dropped soon](https://github.com/gentoo/gentoo/pull/39233) on [java-pkg-simple.eclass](https://qa-reports.gentoo.org/output/eapi-per-eclass/java-pkg-simple.eclass/7.txt).

PR #1296, opened 2024-11-11
dev-python/snakemake and deps: version bumps

mschubert:
Removal of `BDEPENDS` in `snakemake-8.25.3.ebuild` is tested and those deps are not required

Ignored/deselected tests have been reported upstream: https://github.com/snakemake/snakemake/issues/2961 or are missing additional plugin ebuilds


dev-python/snakemake/snakemake-8.25.3.ebuild
```diff
@@ -71,4 +55,17 @@ RDEPEND="
...
 # 	dev-python/recommonmark \
 # 	dev-python/myst-parser
+
+EPYTEST_IGNORE=(
+	# __file__ attribute does not match test name ('python_3_7')
+	tests/test_conda_python_3_7_script/test_script.py
+	# ModuleNotFoundError: No module named 'snakemake_executor_plugin_cluster_generic'
+	tests/test_executor_test_suite.py
+	# ImportError: cannot import name 'snakemake' from 'snakemake.script'
+	tests/test_script_py/scripts/test_explicit_import.py
```
Nowa-Ammerlaan, line 65: Does setting `PYTHONPATH` (or similar) help for this one?

mschubert: If I understand [this](https://github.com/snakemake/snakemake/issues/2961#issuecomment-2258991074) comment correctly, it is not a matter of simply adding to `PYTHONPATH`:

> running `from snakemake.script import snakemake` outside of a Snakemake "context", since `snakemake.script.snakemake` (lower case) is simply a type hint that can be used by editor. At execution time, the `snakemake` object needs to be created by snakemake, which is done by the boilerplate/preamble

Nowa-Ammerlaan: That comment seems to suggest that passing some path to `epytest` would resolve the problem, have you tried this?

mschubert: The import works neither in `${BUILD_DIR}/build/lib` nor `${BUILD_DIR}/install/${EPREFIX}/<python-site-dir>`, nor with these added to `$PYTHONPATH` in the original directory.

Given that this PR is already an improvement over restricting all tests, I'd suggest to accept excluding this test until I can dig more into it.

Nowa-Ammerlaan:
Thanks :+1:

PR #1297, opened 2024-11-11
sci-biology/bx-python: version bump 0.13.0

mschubert:
and enable optional test with `python-lzo`


sci-biology/bx-python/bx-python-0.13.0.ebuild
```diff
@@ -17,6 +17,7 @@ LICENSE="MIT"
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
 
+BDEPEND="test? ( dev-python/python-lzo )"
```
Nowa-Ammerlaan, line 20:
```suggestion
BDEPEND="test? ( dev-python/python-lzo[${PYTHON_USEDEP}] )"
```

Nowa-Ammerlaan (changes requested):

sci-biology/bx-python/bx-python-0.13.0.ebuild
mschubert: Uh, sorry - shouldn't have missed that! Thanks

Nowa-Ammerlaan:
Thanks :+1:

PR #1303, opened 2024-11-26
profiles: mask sci-biology/{samri,ants} depends on non-existent vtk version

APN-Pucky:
For some reason just use masking `sci-biology/ants[vtk]` was not enough. Now I just mask the package and it revdeps.

### DIFF
--- a/profiles/package.mask
+++ b/profiles/package.mask
@@ -30,6 +30,14 @@
 
 #--- END OF EXAMPLES ---
 
+# Alexander Puck Neuwirth <alexander@neuwirth-informatik.de> (2024-11-26)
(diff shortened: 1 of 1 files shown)


Nowa-Ammerlaan (approved):
> For some reason just use masking `sci-biology/ants[vtk]` was not enough. Now I just mask the package and it revdeps.

That is because USE flag masks are part of a profile. It is unfortunately not possible to "append" to a `::gentoo` profile in another repository.  You can define new profiles,  and have that new profile inherit the `::gentoo` profile, but this does not solve the dependency problem.

It is a very unfortunate limitation of portage.

PR #1306, opened 2024-12-04
sci-libs/starparse: switch to guile eclass

APN-Pucky:
- old upstream with no updates
- no revdeps

### DIFF
--- a/sci-libs/starparse/starparse-0.9-r1.ebuild
+++ b/sci-libs/starparse/starparse-0.9-r1.ebuild
@@ -1,7 +1,10 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=7
+EAPI=8
+
+GUILE_COMPAT=( 1-8 )
+inherit guile-single
 
 DESCRIPTION="Library for parsing NMR star files (peak-list format) and CIF files"
 HOMEPAGE="http://burrow-owl.sourceforge.net/"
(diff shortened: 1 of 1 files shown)


profiles/package.mask
```diff
@@ -30,6 +30,10 @@
 
 #--- END OF EXAMPLES ---
 
+# Alexander Puck Neuwirth <alexander@neuwirth-informatik.de> (2024-12-04)
+# Depends on removed version of dev-scheme/guile:12
```
thesamesam, line 34: Note that guile is just slotted now, so removed is accurate but not quite so helpful -- it's more that the package needs porting to the slotted guile eclass, no?

APN-Pucky: It depends on `dev-scheme/guile:12` and `eix` shows me 1.8 2.2 and 3.0 as slots.

thesamesam: guile 1.8.8-r5 was dropped in 7df4aaeb4d02b0c7e225c8194a5962c5d46507ee in gentoo.git where it had its slot as 12. I have no idea why it had that...

APN-Pucky: I see they (new slots) all were slot 12 previously.

APN-Pucky: Ok, I will give porting it a quick try.

APN-Pucky: Only `1-8` passes tests, which is not surprising, given the code is from 2008 or so.

thesamesam: Thank you!

PR #1313, opened 2025-01-26
dev-python/rich-click: version bump 1.8.5 and enable tests

mschubert:


Nowa-Ammerlaan:
Thanks :+1: 

There was a little error in the SRC_URI that I fixed while merging.

PR #1314, opened 2025-01-26
dev-python/yte: version bump 1.5.6 and add tests

mschubert:


dev-python/yte/yte-1.5.6.ebuild
```diff
@@ -20,3 +21,9 @@ RDEPEND="
 	dev-python/plac[${PYTHON_USEDEP}]
 	dev-python/pyyaml[${PYTHON_USEDEP}]
 "
+
+distutils_enable_tests pytest
+
+python_test() {
+	epytest tests.py || die "Tests failed with ${EPYTHON}"
+}
```
Nowa-Ammerlaan, line 29: `|| die` is not necessary here, `epytest` already does that.

mschubert: Fixed, and also removed `/refs/tags/` from github url (the link in *Releases* includes it, but it works without)

Nowa-Ammerlaan:
Thanks :+1:

PR #1316, opened 2025-01-26
dev-python/pybedtools: version bump 0.11.0 and enable docs

mschubert:


dev-python/pybedtools/pybedtools-0.11.0.ebuild
```diff
@@ -24,11 +24,9 @@ RDEPEND="
 	dev-python/pandas[${PYTHON_USEDEP}]
 	dev-python/matplotlib[${PYTHON_USEDEP}]
 "
+BDEPEND="dev-python/cython[${PYTHON_USEDEP}]"
```
Nowa-Ammerlaan, line 27: dependency on cyhthon is not required when `DISTUTILS_EXT=1` is set

mschubert: fixed

Nowa-Ammerlaan (commented):
```to1: internal compiler error: unexpected file name /tmp/portage/dev-python/pybedtools-0.11.0/work/pybedtools-0.11.0-python3_12/build/temp.linux-x86_64-cpython-312/pybedtools/include/fileType.o in linker resolution file.Expected /tmp/portage/dev-python/pybedtools-0.11.0/work/pybedtools-0.11.0-python3_12/build/temp.linux-x86_64-cpython-312/pybedtools/include/gzstream.o
0x55c30ed92fbf internal_error(char const*, ...)
???:0
0x55c30d335cf8 read_cgraph_and_symbols(unsigned int, char const**)
???:0
0x55c30d2f9807 lto_main()
???:0
Please submit a full bug report, with preprocessed source (by using -freport-bug).
Please include the complete backtrace with any bug report.
See <https://bugs.gentoo.org/> for instructions.
lto-wrapper: fatal error: /usr/bin/x86_64-pc-linux-gnu-g++ returned 1 exit status
compilation terminated.
```

It fails to build on my end

mschubert:
Was this with `USE=doc`?

I realized I misunderstood how I still have to set this flag for sphinx, so that is likely broken (looking into it)

Nowa-Ammerlaan:
> Was this with `USE=doc`?
> 
> I realized I misunderstood how I still have to set this flag for sphinx, so that is likely broken (looking into it)

It was, but not sure if the error originates from sphinx.

dev-python/pybedtools/pybedtools-0.11.0.ebuild
mschubert: I've removed this and am now getting a QA Notice:

> * QA Notice: Package seems to be missing a BDEPEND on dev-python/cython.

Nowa-Ammerlaan: Then please check if you are using the most recent version of the eclass.

mschubert: I have the main tree synced at `2025-01-27 11:18:44 UTC`, so this should include the latest `distutils-r1` eclass. I still get the QA notice

Nowa-Ammerlaan:
It's still failing, looks like an lto issue.

mschubert:
I can confirm this is broken with `-flto`, but this already happens with the previous version `0.10.0`.

I've got no idea how to debug this.

Nowa-Ammerlaan:
> I can confirm this is broken with `-flto`, but this already happens with the previous version `0.10.0`.
> 
> I've got no idea how to debug this.

Don't bother, just add `filter-lto` to the ebuild to ensure the lto flags are removed if they are present.

mschubert:
Added in the current commit

Nowa-Ammerlaan:
Thanks :+1:

PR #1318, opened 2025-01-26
dev-python/pyBigWig: version bump 0.3.24, add tests, remove outdated ebuilds

mschubert:
required for deeptools PR


dev-python/pyBigWig/pyBigWig-0.3.24.ebuild
```diff
@@ -0,0 +1,32 @@
...
+
+REPEND="
+	sys-libs/zlib:0=
+	sci-biology/libBigWig
+"
+DEPEND="${RDEPEND}"
+
+distutils_enable_tests pytest
+
+python_test() {
```
mschubert, line 27: I'm not sure if there is a better way than linking

dev-python/pyBigWig/pyBigWig-0.3.24.ebuild
```diff
@@ -0,0 +1,32 @@
...
+	local pysite="${BUILD_DIR}"/install/${EPREFIX}/usr/lib/${EPYTHON}/site-packages
+	ln -s "${S}"/pyBigWigTest/test.bigBed "${pysite}"/pyBigWigTest || die
+	ln -s "${S}"/pyBigWigTest/test.bw "${pysite}"/pyBigWigTest || die
+	epytest pyBigWigTest/test.py
```
mschubert, line 31: I'm not sure if there is a better way than linking

Nowa-Ammerlaan: What about setting `PYTHONPATH`?

mschubert: The problem here is that:

1. The test accesses the file using `os.path.dirname(pyBigWig.__file__) + "/pyBigWigTest/test.bigBed"`
2. `pyBigWig` only exists as `.so` file in the build directory (without those files)
3. Setting `PYTHONPATH` doesn't change the file path construction of the former

Nowa-Ammerlaan: `DISTUTILS_IN_SOURCE_BUILD=1` might help here.

Nowa-Ammerlaan: Right, let's just keep it as it is then.

Nowa-Ammerlaan:
Thanks :+1:

PR #1319, opened 2025-01-31
dev-python/dotenv: add ebuild

mschubert:
will be required for `sci-biology/multiqc` update


kiwifb:
Already in the main tree https://packages.gentoo.org/packages/dev-python/python-dotenv

mschubert:
Oops, sorry! I searched for `dotenv` only and emerge didn't find it (but I should have checked better)

PR #1324, opened 2025-02-17
dev-python/py2bit: version bump 0.3.3; remove live ebuild

mschubert:
gcc14 patch is no longer required

live ebuild is very close to release, no need to duplicate


Nowa-Ammerlaan (commented):
Why is the `9999` being removed? Seems to me we might as well keep it.

mschubert:
`9999` no longer worked with the recent version and the package has very low dev activity (so no added value)

Nowa-Ammerlaan:
> `9999` no longer worked with the recent version and the package has very low dev activity (so no added value)

The added value is that you can upgrade before the ebuild is bumped, which I think is valuable considering ebuilds in ::sci are often outdated.

mschubert:
I agree with you about tools that are more frequently updated. But `py2bit` has [had a small bugfix in 2024, one in 2022, and the last feature added in 2018](https://github.com/deeptools/py2bit/releases) (the 2025 change was only to build infrastructure).

It's very unlikely that any additional upgrades will be required _for this package_.

But I can re-add the ebuild if you prefer.

Nowa-Ammerlaan (commented):
```
* Python extension modules (*.so) found installed. Please set:
*   DISTUTILS_EXT=1
* in the ebuild.
```

mschubert:
> ```
> * Python extension modules (*.so) found installed. Please set:
> *   DISTUTILS_EXT=1
> * in the ebuild.
> ```

fixed, and also added py 3.13 compat

Nowa-Ammerlaan:
Thanks :+1:

PR #1328, opened 2025-02-24
dev-python/yte: version bump 1.7.0, py3.13 compat

mschubert:
incl. py3.13 compat deps plac and dpath (all tests pass)


Nowa-Ammerlaan:
Thanks :+1:

PR #1330, opened 2025-03-07
dev-python/llvmlite: fix double EPREFIX

mschubert:
as described here: https://github.com/gentoo/sci/pull/1286/files#r1960450387

### DIFF
--- a/dev-python/llvmlite/llvmlite-0.44.0.ebuild
+++ b/dev-python/llvmlite/llvmlite-0.44.0.ebuild
@@ -36,6 +36,7 @@ python_test() {
 
 python_install() {
 	distutils-r1_python_install
-	dosym "$(get_llvm_prefix)/$(get_libdir)/libLLVM-${LLVM_SLOT}.so" \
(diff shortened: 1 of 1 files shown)


Nowa-Ammerlaan (commented):
Don't you have the same problem here with `get_llvm_prefix`?

I'd propose to instead replace `dosym` with `ln -s || die`

mschubert:
The destination of the link is correct, I think because `dosym` only prefixes `${D}` where the link is created. As far as I understand, this shouldn't affect `$(get_llvm_prefix)` where the link points to because the path doesn't need to be prefixed with `${D}`.

I've now changed `dosym` with `ln -s`, let me know if this looks good to you!

Nowa-Ammerlaan:
> The destination of the link is correct, I think because `dosym` only prefixes `${D}` where the link is created. As far as I understand, this shouldn't affect `$(get_llvm_prefix)` where the link points to because the path doesn't need to be prefixed with `${D}`.

Well that is why we usually use relative paths for symlinks. Not sure if it works properly now, but this does solve the immediate problem.

PR #1332, opened 2025-03-08
dev-python/numcodecs: version bump 0.15.1

mschubert:
I don't know why, but `filter-flags -pipe` is strictly required on my system and leads to missing symbols otherwise

Requirement of `local -x DISABLE_NUMCODECS_{AVX2,SSE2}=1` is a known problem upstream


Nowa-Ammerlaan:
> I don't know why, but `filter-flags -pipe` is strictly required on my system and leads to missing symbols otherwise

I really think this should not be in the ebuild, I cannot reproduce your problem and it does not make sense. There is something weird happening on your end, but filtering this flag here is a weird symptomatic solution.

mschubert:
I totally agree that this should not be in the ebuild if it only fixes the build on my system but not others.

Does it work for you without `filter-flags -pipe` _and without_ `-march=native` from the LTO settings?

related: https://github.com/gentoo/sci/pull/1289#issuecomment-2258741445

Nowa-Ammerlaan:
> Does it work for you without `filter-flags -pipe` _and without_ `-march=native` from the LTO settings?

Somehow it now also breaks for me without filtering `-pipe`, I am confused.

PR #1337, opened 2025-03-18
dev-python/pybedtools: version bump 0.12.0

mschubert:
py3.13 compat depends on pysam=0.23, not yet in gentoo main tree


Nowa-Ammerlaan:
> py3.13 compat depends on pysam=0.23, not yet in gentoo main tree

Done!

PR #1340, opened 2025-03-26
sci-physics/samurai: new package, add 2.1.1

APN-Pucky:
- [x] test it again


sci-physics/samurai/samurai-2.1.1.ebuild
```diff
@@ -0,0 +1,51 @@
...
+src_configure() {
+	default
+	# Fix that qcdloop and oneloop are already installed
+	sed -i 's/lib_LTLIBRARIES.*/lib_LTLIBRARIES = libsamurai.la/g' Makefile.am || die
+	econf FCFLAGS="${FCFLAGS} -std=legacy -fPIC -I${ESYSROOT}/usr/include" \
+		--with-avh_olo="${ESYSROOT}"/usr/$(get_libdir)/libavh_olo.so \
+		--with-avh_olo_precision=double \
+		--with-precision=double \
+		$(use_with qcdloop qcdloop "${ESYSROOT}"/usr) \
+		$(use_with looptools looptools "${ESYSROOT}"/usr)
```
Nowa-Ammerlaan, line 38: This would look a bit nicer if we'd use a `myeconfargs` array here instead.

APN-Pucky: I think also the default before needs to be removed, since that calls emake already... (Also needs bash...).

Nowa-Ammerlaan (approved):

sci-physics/samurai/samurai-2.1.1.ebuild
```diff
@@ -0,0 +1,54 @@
...
+
+src_configure() {
+	# Fix that qcdloop and oneloop are already installed
+	sed -i 's/lib_LTLIBRARIES.*/lib_LTLIBRARIES = libsamurai.la/g' Makefile.am || die
+	local myeconfargs=(
+		FCFLAGS="${FCFLAGS} -std=legacy -fPIC -I${ESYSROOT}/usr/include"
+		--with-avh_olo="${ESYSROOT}"/usr/$(get_libdir)/libavh_olo.so
+		--with-avh_olo_precision=double
+		--with-precision=double
+		$(use_with qcdloop qcdloop "${ESYSROOT}"/usr)
```
APN-Pucky, line 37: Does this need quotes again in case `ESYSROOT` has a space? I.e. https://github.com/gentoo/gentoo/blob/f84b09f5dd30d2e41d33afd6c284cc99cd9e22fa/app-admin/sudo/sudo-1.9.16_p2-r1.ebuild#L184 quotes twice.

Nowa-Ammerlaan: Yes it needs (double) quotes, the inside quotes ensure that the `ESYSROOT/usr` is one argument for `use_with`, the outside quotes ensures that the `ESYSROOT/usr` is one single element of the array.

PR #1344, opened 2025-04-09
dev-python/numcodecs: version bump

mschubert:
draft because zarr v2 might need numcodecs<16, still need to test: https://github.com/zarr-developers/zarr-python/releases


mschubert:
deps are a bit strange, but this is tested on my machine:

* numcodecs-15 test fails if zarr<3 installed, but succeeds if no zarr installed or zarr>3 installed
* zarr-2 itself needs <numcodecs-16

Nowa-Ammerlaan:
> * numcodecs-15 test fails if zarr<3 installed, but succeeds if no zarr installed or zarr>3 installed

It probably automatically skips those failing tests if zarr is not installed. It is pretty common for pytest to run more tests is certain optional dependencies are installed. The correct solution is then to BDEPEND=test? on zarr>=3. This maximises the amount of tests run.

mschubert:
Fixed now

dev-python/numcodecs/numcodecs-0.16.0.ebuild
```diff
@@ -0,0 +1,53 @@
+# Copyright 1999-2025 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+PYTHON_COMPAT=( python3_{10..13} )
```
Nowa-Ammerlaan, line 6: You dropped py3.11 in the other ebuild, so lets do the same here.

PR #1349, opened 2025-04-14
dev-python/donfig: version bump

mschubert:


dev-python/donfig/donfig-0.8.1_p1.ebuild
```diff
@@ -10,7 +10,8 @@ inherit distutils-r1
 
 DESCRIPTION="Configure a package including defaults, env variable loading, and yaml loading"
 HOMEPAGE=https://donfig.readthedocs.io/en/latest/
-SRC_URI="https://github.com/pytroll/donfig/archive/v${PV}.tar.gz -> ${P}.gh.tar.gz"
+SRC_URI="https://github.com/pytroll/donfig/archive/v0.8.1.post1.tar.gz -> ${P}.gh.tar.gz"
+S="${WORKDIR}/${PN}-0.8.1.post1"
```
Nowa-Ammerlaan, line 15: We should not hardcore the version number. Instead you can use bash string manipulation here to translate the PV to the primary and post part.

mschubert: I changed this to `${PV/_p/.post}`

PR #1356, opened 2025-05-05
fix old pythons

APN-Pucky:
Most of them are `python3_10` so I assume they are not in use and mask them. If someone has objections I think some of them would work with just a python version increase, but I don't want to test all of them now.

... Apparently masking them does not make `pkgcheck` happy.


Nowa-Ammerlaan:
Masking won't help in this case, the eclass is calling `die` when it is `inherit`ed. Hence the ebuild cannot be sourced.

We'll have to remove these packages directly, or update them.

APN-Pucky:
dev-libs/lief-0.11.5 fails to build

Nowa-Ammerlaan (approved):

PR #1368, opened 2025-07-13
dev-python/numcodecs: fix gcc15 build and version bump 0.16.1

mschubert:
borrowing a patch from `sci-libs/c-blosc::gentoo`

ideally we would unbundle c-blosc entirely, but there are ongoing discussions whether to support that upstream


dev-python/numcodecs/numcodecs-0.16.1.ebuild
```diff
@@ -48,6 +51,6 @@ python_compile() {
 python_test() {
 	local PY_BUILD_DIR=$(${EPYTHON} -c "import sysconfig; print('lib.' + sysconfig.get_platform() +
 		'-cpython-' + sysconfig.get_python_version().replace('.', ''))") || die
-	cd "${BUILD_DIR}/build/${PY_BUILD_DIR}" || die
+	cd "${BUILD_DIR}/build${#DISTUTILS_WHEELS}/${PY_BUILD_DIR}" || die
```
Nowa-Ammerlaan, line 54:
```suggestion
	cd "${BUILD_DIR}/build$((${#DISTUTILS_WHEELS[@]}-1))/${PY_BUILD_DIR}" || die
```

PR #1377, opened 2025-09-15
sci-physics/openloops: add OPENLOOPS_PROCESSES USE_EXPAND

APN-Pucky:
This allows setting e.g. in `/etc/portage/package.use/openloops`:
```
sci-physics/openloops openloops_processes_*
```
, which is very convenient if one just wants all processes. (similar to other USE_EXPAND for lhapdf_sets)


APN-Pucky:
merged

PR #1389, opened 2026-01-30
net-misc/cernbox-client: treeclean

olifre:
CERN does not provide full sourcecode anymore (only binary AppImage) and from support contact, no plans to change that. OwnCloud remains open, but CERN-specific branding (quota path etc.) are now effectively closed-source. Treeclean, users should use net-misc/owncloud-client which works fine apart from branding and showing available quota.


Nowa-Ammerlaan:
The license of the owncloud-client is GPL (according to the ebuild). That would mean CERN is required to somehow distribute the source code if they are distributing binaries.

olifre:
Indeed. I know (and pointed that out in my query to support) — the AppImage distributed right now is built for CERN by OwnCloud, and it seems they are unwilling to ship full sources for that. 
I was not willing to fight further, even though this is for sure a fight which could legally be won.

Nowa-Ammerlaan:
Could we do what Arch/AUR does, unpack and install the fedora/rhel rpm? https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=cernbox

We have some examples of ebuilds that are essentially just wrapping around an rpm in ::gentoo.

olifre:
> Could we do what Arch/AUR does, unpack and install the fedora/rhel rpm? https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=cernbox

Technically, yes. But the RPM contains binaries of most of Qt6 libraries, binaries of CERNBox itself, OpenSSL etc., it's essentially a wrap of all dependencies / a container. 
In other words: The RPM is just the unpacked AppImage they ship repacked into RPM format, for extracting into `/opt` with a few wrappers. 

The AUR package leaves these all these in in the versions shipped by upstream, which I would personally say is a no-go for a distro package. 
Replacing all dep libraries with symlinks (de-vendoring) or modifying the wrappers seems not well-maintainable, given their testing is reduced to the exact versions they chose. It might then even be easier to ship the AppImage via an `ebuild` in that case... 

For a package which requires to enter CERN credentials for using it, I would personally never use a closed-source binary, so it would be something I would realistically not maintain. 

As the upstream service of CERN is compatible with `net-misc/owncloud-client`, this seems like the better approach (security-wise), the only lost feature is the displaying of available quota.

Nowa-Ammerlaan:
Alrighty, that does make sense. Merged!

PR #1391, opened 2026-03-15
sci-biology/bx-python: version bump 0.14.0

mschubert:


Nowa-Ammerlaan:
This was merged but it looks like the mirroring to GitHub is not working anymore or already has been disabled.

In the future please submit pull requests to our new Codeberg mirror.

PR #1401, opened 2026-05-29
*/*: fix python errors from py3.11 removal and enable py3.14

APN-Pucky:
I did not fix everything in this, but let's merge this fraction already and more PRs will follow.


Nowa-Ammerlaan (commented):
We should also drop 10 and 11 from existing Python compat when adding the new targets, otherwise pkgcheck/qawarn will still complain. 

I usually mass-sed the dead Python implementations out, and then bump or remove whatever needs manual attention.

APN-Pucky:
> I usually mass-sed the dead Python implementations out, and then bump or remove whatever needs manual attention.

I see, I am testing each to see if the tests pass with new python versions too. Taking a bit more time, but since there is no ::sci ci I think it makes sense to at least check every now and then. (i.e. some masks here are because they don't compile anymore)

Nowa-Ammerlaan:
> > I usually mass-sed the dead Python implementations out, and then bump or remove whatever needs manual attention.
> 
> I see, I am testing each to see if the tests pass with new python versions too. Taking a bit more time, but since there is no ::sci ci I think it makes sense to at least check every now and then. (i.e. some masks here are because they don't compile anymore)

Oh yes, that is for sure the way to go when adding targets. What I meant is that the removal of dead targets can be done in mass to clear the QA warnings.

APN-Pucky:
Finally, the CI passes :). I had to wipe a few ebuilds because of failures of `sys-devel/DPC++` and `dev-python/python-igraph` and whatever used them. I masked some packages, but that does not help with the python compat eclass failures so some of the newly masked packages got treecleaned or enabled py3.14 afterwards.

Nowa-Ammerlaan:
> Finally, the CI passes :). I had to wipe a few ebuilds because of failures of `sys-devel/DPC++` and `dev-python/python-igraph` and whatever used them. I masked some packages, but that does not help with the python compat eclass failures so some of the newly masked packages got treecleaned or enabled py3.14 afterwards.

Awesome, thanks!!

Nowa-Ammerlaan (approved):

